PR #4, opened 2026-02-23
Initial implementation of tracking task wakes

rcoh:
Track what caused a wake and scheduling delay

### DIFF
--- a/dial9-tokio-telemetry/examples/analyze_trace.rs
+++ b/dial9-tokio-telemetry/examples/analyze_trace.rs
@@ -1,5 +1,5 @@
 use dial9_tokio_telemetry::telemetry::{
-    TraceReader, TelemetryEvent, analyze_trace, compute_wake_to_poll_delays, detect_idle_workers,
+    TelemetryEvent, TraceReader, analyze_trace, compute_wake_to_poll_delays, detect_idle_workers,
     print_analysis,
 };
 use std::collections::HashMap;
@@ -51,13 +51,22 @@ fn main() {
     // Build task_id → spawn_loc_id from PollStart events (more complete than TaskSpawn alone)
     let mut task_locs: HashMap<u32, u16> = HashMap::new();
     for e in &events {
-        if let TelemetryEvent::PollStart { task_id, spawn_loc_id, .. } = e {
-            task_locs.entry(task_id.to_u32()).or_insert(spawn_loc_id.as_u16());
+        if let TelemetryEvent::PollStart {
+            task_id,
+            spawn_loc_id,
+            ..
+        } = e
+        {
+            task_locs
+                .entry(task_id.to_u32())
+                .or_insert(spawn_loc_id.as_u16());
         }
     }
     // Also include TaskSpawn mappings from the reader
     for (task_id, spawn_loc_id) in &reader.task_spawn_locs {
-        task_locs.entry(task_id.to_u32()).or_insert(spawn_loc_id.as_u16());
+        task_locs
+            .entry(task_id.to_u32())
+            .or_insert(spawn_loc_id.as_u16());
     }
 
     // Count wakes by waker spawn location
@@ -71,7 +80,9 @@ fn main() {
                 *wakes_by_loc.entry(Some("<non-task context>")).or_default() += 1;
                 resolved += 1;
             } else if let Some(loc_id) = task_locs.get(&id) {
-                let loc = reader.spawn_locations.get(&dial9_tokio_telemetry::telemetry::SpawnLocationId::from_u16(*loc_id));
+                let loc = reader
+                    .spawn_locations
+                    .get(&dial9_tokio_telemetry::telemetry::SpawnLocationId::from_u16(*loc_id));
                 *wakes_by_loc.entry(loc.map(|s| s.as_str())).or_default() += 1;
                 resolved += 1;
             } else {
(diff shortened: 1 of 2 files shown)


src/telemetry/recorder.rs
```diff
@@ -381,6 +436,7 @@ impl TelemetryRecorder {
         if task_tracking_enabled {
             let s5 = shared.clone();
             builder.on_task_spawn(move |meta| {
+                println!("meta: {:?}", meta.id());
```
jlizen: Needs cleanup

src/telemetry/recorder.rs
```diff
@@ -186,34 +200,99 @@ impl FlushState {
...
         self.intern_map.insert(ptr, id);
-        self.intern_strings.push(format!(
-            "{}:{}:{}",
-            location.file(),
-            location.line(),
-            location.column()
-        ));
+        if location == TOKIO_INTERNAL_LOCATION {
+            self.intern_strings
+                .push("<tokio internal (io driver, etc.>".to_string());
```
jlizen: nit: typo

also this was kind of spammy in the viewer. i know you've already iterated on this though, feel free to ignore.

src/traced.rs
```diff
@@ -0,0 +1,129 @@
...
+);
+
+unsafe fn traced_waker_clone(data: *const ()) -> RawWaker {
+    let arc = unsafe { Arc::from_raw(data as *const TracedWakerData) };
+    let cloned = arc.clone();
+    std::mem::forget(arc);
+    RawWaker::new(Arc::into_raw(cloned) as *const (), &VTABLE)
+}
+
+unsafe fn traced_waker_wake(data: *const ()) {
```
jlizen: Nit: you can avoid the inner arc clone with: 
```rust
unsafe fn traced_waker_wake(data: *const ()) {
    let arc = unsafe { Arc::from_raw(data as *const TracedWakerData) };
    record_wake_event(&arc);
    match Arc::try_unwrap(arc) {
        Ok(data) => data.inner.wake(),
        Err(arc) => arc.inner.wake_by_ref(),
    }
}

```

src/traced.rs
```diff
@@ -0,0 +1,129 @@
...
+    let arc = unsafe { Arc::from_raw(data as *const TracedWakerData) };
```
jlizen: I know this is all pretty experimental, but could we add some simple `// SAFETY` comments on the functions and inline unsafe calls? Just specifying the invariant around what type the pointer needs to be from.

src/traced.rs
```diff
@@ -0,0 +1,129 @@
...
+impl<F: Future> Future for Traced<F> {
+    type Output = F::Output;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        if !this.handle.shared.enabled.load(Ordering::Relaxed) {
+            return this.inner.poll(cx);
+        }
+
+        let traced_waker = make_traced_waker(cx.waker(), *this.task_id, &this.handle.shared);
```
jlizen: This allocates a fresh `Arc` every poll. But we can keep using the waker unless the task is stolen across tokio workers, right?

IE would something like this work?
```rust
pin_project! {
    pub struct Traced<F> {
        #[pin]
        inner: F,
        handle: TracedHandle,
        task_id: TaskId,
        cached_waker: Option<Waker>,
    }
}

// In poll():
let waker = cx.waker();
let traced_waker = match &this.cached_waker {
    Some(w) if w.will_wake(waker) => w.clone(),
    _ => {
        let w = make_traced_waker(waker, *this.task_id, &this.handle.shared);
        *this.cached_waker = Some(w.clone());
        w
    }
};
```

tests/echo_server.rs
jlizen, line 83: would be nice to use `guard.spawn()` here to flush out macro issues with the `Traced` impl

src/traced.rs
```diff
@@ -0,0 +1,129 @@
...
+/// Handle used by `Traced<F>` to emit events into the telemetry system.
+/// Obtained via `TelemetryHandle::traced_handle()`.
+#[derive(Clone)]
+pub struct TracedHandle {
+    pub(crate) shared: Arc<SharedState>,
+}
+
+pin_project! {
+    /// Future wrapper that captures wake events (and later, task dumps).
+    pub struct Traced<F> {
```
jlizen: Would be nice to have just a small unit test validating a wake / poll cycle, something like: 

```rust
#[tokio::test]
async fn traced_emits_wake_events() {
    // Build a TracedRuntime with an in-memory writer
    // Use guard.spawn() to run a task that awaits a Notify
    // Notify it from another task
    // Flush, read the trace, assert at least one WakeEvent exists
    // with the correct woken_task_id
}
```

dial9-tokio-telemetry/src/telemetry/task_metadata.rs
```diff
@@ -15,14 +15,22 @@ impl Serialize for TaskId {
...
+        id.hash(&mut extractor);
+        TaskId(extractor.0)
     }
 }
 
+/// A Hasher that captures the first u64 written to it.
+struct U64Extractor(u64);
+
+impl Hasher for U64Extractor {
+    fn write(&mut self, _bytes: &[u8]) {}
```
jlizen: Bit fragile, we could guard against tokio not calling `write_u64` a `debug_assert` here. I don't think that's part of their public contract?

PR #9, opened 2026-02-27
fix Add perf-self-profile crate

rcoh:
Standalone crate wrapping perf_event_open for:
- CPU sampling (frequency-based, sw/hw counters)
- Context-switch capture (per-thread sched events)
- Ring buffer mmap reader for zero-copy sample access
- Symbol resolution via blazesym

Includes integration tests for multi-thread profiling and sched event capture (sleep, lock acquisition stacks).


perf-self-profile/src/lib.rs
```diff
@@ -0,0 +1,42 @@
+//! # perf-self-profile
+//!
+//! Minimal crate for a program to capture its own perf events with stack traces
```
jlizen, line 3: Worth mentioning `perf_event_paranoid` here?

perf-self-profile/src/ring_buffer.rs
```diff
@@ -0,0 +1,186 @@
...
+//! parse records, then advance `data_tail` to tell the kernel we're done.
+
+use std::mem;
+use std::ptr;
+use std::sync::atomic::{Ordering, fence};
+
+use crate::sys::{PerfEventHeader, PerfEventMmapPage};
+
+/// A mapped perf ring buffer.
+pub struct RingBuffer {
```
jlizen: nit: i'd prefer `pub(crate)` to show it is internal API

perf-self-profile/src/ring_buffer.rs
```diff
@@ -0,0 +1,186 @@
...
+    base: *mut u8,
+    /// Size of the data region only (excluding the metadata page).
+    data_size: u64,
+    /// Total mmap size (metadata page + data pages), for munmap.
+    mmap_size: usize,
+    /// Our current read position.
+    position: u64,
+}
+
+// Safety: The ring buffer pointer is only accessed by the owner thread,
```
jlizen: We currently access it from multiple threads in our tests (`Arc<Mutex<PerfSampler(RingBuffer)>>`. 

It seems to be more about single threaded access rather than thread-local access? And kernel access is handled by the fences.

perf-self-profile/src/ring_buffer.rs
```diff
@@ -0,0 +1,186 @@
...
+        unsafe {
+            let page = &mut *(self.base as *mut PerfEventMmapPage);
+            fence(Ordering::Release);
+            ptr::write_volatile(&mut page.data_tail, value);
+        }
+    }
+
+    fn data_slice(&self) -> &[u8] {
+        unsafe {
+            let data_ptr = self.base.add(4096); // skip metadata page
```
jlizen, line 122: This isn't portable. Can we use `PerfEventMmapPage.data_offset` instead?

Cargo.lock
jlizen, line 1: this is stale, probably we should .gitignore it

perf-self-profile/Cargo.lock
jlizen, line 1: this is stale, probably we should .gitignore it

perf-self-profile/src/sys.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub time_enabled: u64,
+    pub time_running: u64,
+    pub capabilities: u64,
+    pub pmc_width: u16,
+    pub time_shift: u16,
+    pub time_mult: u32,
+    pub time_offset: u64,
+    pub time_zero: u64,
+    pub size: u32,
+    pub _reserved: [u8; 948], // pad to offset 0x400 = 1024
```
jlizen: This is fragile, it breaks if any new preceding fields mess with the offset.

We should at minimum add a compile-time assertion to catch drift:
```
  const _: () = assert!(std::mem::offset_of!(PerfEventMmapPage, data_head) == 0x400);
```

perf-self-profile/src/sys.rs
```diff
@@ -0,0 +1,172 @@
+//! Raw Linux perf_event types, constants, and syscall wrapper.
```
jlizen: Why not https://crates.io/crates/perf-event-open-sys ?

.github/actions/rust-build/action.yml
```diff
@@ -1,28 +1,28 @@
...
+        - uses: Swatinem/rust-cache@v2
+        - name: Build
+          shell: bash
+          env:
+              RUSTFLAGS: "--cfg tokio_unstable"
+          run: |
+              if [ "${{ inputs.toolchain }}" != stable ]; then
+                  rm -fv Cargo.lock
+              fi
+              cargo build
```
jlizen, line 23: good call

perf-self-profile/src/ring_buffer.rs
```diff
@@ -0,0 +1,186 @@
...
+// and the kernel side is synchronized via data_head/data_tail with fences.
+unsafe impl Send for RingBuffer {}
+
+impl RingBuffer {
+    /// Create a new RingBuffer from an already-mmap'd pointer.
+    ///
+    /// # Safety
+    /// `base` must be a valid mmap'd perf event region with `mmap_size` bytes.
+    pub unsafe fn new(base: *mut u8, data_size: u64, mmap_size: usize) -> Self {
+        RingBuffer {
```
jlizen, line 35: Should include a debug assert (on pointer being non-null, data and mmap size being compatible)

perf-self-profile/src/symbolize.rs
```diff
@@ -0,0 +1,152 @@
...
+const EMPTY: SymbolInfo = SymbolInfo {
+    name: None,
+    base_addr: 0,
+    object: None,
+    offset: 0,
+};
+
+/// Resolve an instruction pointer to a symbol name.
+pub fn resolve_symbol(addr: u64) -> SymbolInfo {
+    SYMBOLIZER.with(|cell| {
```
jlizen, line 86: This won't pick up `dlopen`'d libraries, probably worth a note in the doc comment / TODO

perf-self-profile/src/sampler.rs
```diff
@@ -0,0 +1,614 @@
...
+        (0, 0)
+    };
+
+    let time = if sample_type & PERF_SAMPLE_TIME != 0 {
+        read_u64!()
+    } else {
+        0
+    };
+
+    if sample_type & (1 << 3) != 0 {
```
jlizen: nit: the next few also should be constants

perf-self-profile/examples/test_sleep.rs
```diff
@@ -0,0 +1,34 @@
...
+use std::sync::{Arc, Mutex};
+use std::thread;
+
+#[inline(never)]
+fn do_sleep() {
+    thread::sleep(std::time::Duration::from_millis(50));
+}
+
+fn main() {
+    unsafe { libc::prctl(libc::PR_SET_DUMPABLE, 1) };
```
jlizen, line 11: nit: probably include the safety comment for any copy pasters

perf-self-profile/src/sys.rs
```diff
@@ -0,0 +1,172 @@
...
+
+// --- Record types ---
+
+pub const PERF_RECORD_SAMPLE: u32 = 9;
+pub const PERF_RECORD_LOST: u32 = 2;
+
+// --- Kernel address threshold ---
+// On x86_64, userspace virtual addresses are below 0x0000_8000_0000_0000.
+// Anything at or above this is kernel space, a context marker, or invalid.
+pub const USER_ADDR_LIMIT: u64 = 0x0000_8000_0000_0000;
```
jlizen: sooner or later we should probably set up target config, even if we only support x86 / little endian / etc for the time being. But anyway hopefully we don't need to manage perf bindings.

jlizen (approved):
I think we should probably avoid writing our own perf bindings, but fine to ship as is and adjust later.

perf-self-profile/src/ring_buffer.rs
rcoh: Oh weird I could have sworn I fixed this comment

PR #11, opened 2026-02-27
Integrate CPU profiling into Dial9

rcoh:
This brings CPU sampling into dial9 both on-cpu and off-cpu events are recorded.

These are properly correlated into the trace fail which lets you see what workers are actually doing during long polls (and why they were descheduled).

Along the way I refactored the front end and fixed a bunch of other stuff too


dial9-tokio-telemetry/src/telemetry/format.rs
```diff
@@ -12,23 +12,30 @@
...
 //! ```
 //!
 //! Timestamps are microseconds since trace start. u32 micros supports traces up to ~71 minutes.
 //! `cpu_us` is thread CPU time (from `CLOCK_THREAD_CPUTIME_ID`) in microseconds since thread start.
 //! `sched_wait_us` is the scheduling wait delta from schedstat during the park period.
 //! In-memory representation remains nanoseconds; conversion happens at the wire boundary.
 //!
-//! ### v8 → v9 changes
-//! - Added WakeEvent (code 7): emitted by Traced<F> waker wrapper, records who woke a task
+//! ### v10 → v11 changes
```
jlizen, line 26: should we split the changelog out into its own file? it's kind of weird how older notes are replaced by new

dial9-tokio-telemetry/design/cpu-sampling-and-worker-id-design.md
```diff
@@ -0,0 +1,183 @@
...
+`SpawnLocationDef`, `CallframeDef`, and `ThreadNameDef` are metadata events that must be re-emitted when the writer rotates to a new file. The system tracks this via:
+
+- `FlushState.emitted_this_file: HashSet<SpawnLocationId>` — cleared on rotation.
+- `callframe_emitted_this_file: HashSet<u64>` — cleared on rotation.
+- `thread_name_emitted_this_file: HashSet<u32>` — cleared on rotation.
+- `writer.take_rotated()` is checked before writing events and after `write_atomic` returns.
+
+---
+
+## 6. Wire Codes (v13)
```
jlizen, line 167: This numbering is inconsistent with `format.rs` notes and the viewer impl. Consider consolidating on a single source of truth.

dial9-tokio-telemetry/Cargo.toml
```diff
@@ -13,6 +13,12 @@ pin-project-lite = "0.2"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 smallvec = "1"
+perf-self-profile = { path = "../perf-self-profile", optional = true }
+
+[features]
+cpu-profiling = ["perf-self-profile"]
+
+default = ["cpu-profiling"]
```
jlizen: non-blocking: this should either be non-default or platform gated, probably. fits into the broader "target gating" story. I'd suggest we address that before crates.io publish.

dial9-tokio-telemetry/src/telemetry/events.rs
```diff
@@ -1,6 +1,27 @@
...
+    CpuProfile = 0,
+    /// Context switch captured by per-thread sched event tracking.
+    SchedEvent = 1,
+}
+
+impl CpuSampleSource {
+    pub fn from_u8(v: u8) -> Self {
+        match v {
+            1 => Self::SchedEvent,
+            _ => Self::CpuProfile,
```
jlizen, line 20: We should match on `CpuProfile` and return an option or result on unknown type, probably.

dial9-tokio-telemetry/src/telemetry/recorder.rs
```diff
@@ -365,18 +434,120 @@ impl TelemetryRecorder {
     }
 
     fn flush(&mut self) {
+        // Sync worker tid→id mappings into both profilers with a single lock
```
jlizen: This function is getting pretty huge, with interwoven conditional compilation for cpu profiling. Is there a way to extract cpu profiling into its own method?

dial9-tokio-telemetry/src/telemetry/recorder.rs
```diff
@@ -365,18 +434,120 @@ impl TelemetryRecorder {
...
+        // acquisition, then drain samples from each.
+        #[cfg(feature = "cpu-profiling")]
+        {
+            let tids = self.shared.worker_tids.lock().unwrap();
+            if let Some(ref mut profiler) = self.cpu_profiler {
+                for (&tid, &worker_id) in tids.iter() {
+                    profiler.register_worker(worker_id, tid);
+                }
+            }
+            let mut shared_profiler = self.shared.sched_profiler.lock().unwrap();
```
jlizen: I think it's correct currently, but this double locking is a future deadlock risk if anybody else starts locking the sched profiler  before worker. Consider at least highlighting this,or perhaps as a guard rail making the lock method for the sched profiler require an input demonstrating that the worker lock is already held. (IE turn the whole thing into a state machine)

dial9-tokio-telemetry/src/telemetry/recorder.rs
```diff
@@ -443,6 +631,28 @@ impl TelemetryRecorder {
             });
         }
 
+        // Wire per-thread sched event tracking into worker lifecycle.
+        #[cfg(feature = "cpu-profiling")]
+        {
+            let s_start = shared.clone();
+            let s_stop = shared.clone();
+            builder
+                .on_thread_start(move || {
```
jlizen: (also exists for existing behaviors)

This nukes anything that is set beforehand, probably we should document that at least. I don't see a great way to chain any pre-set ones on a user-provided runtime builder.

In the future it might be nice to accept a closure for user-provided hooks.

dial9-tokio-telemetry/trace_viewer/trace_parser.js
```diff
@@ -0,0 +1,244 @@
...
+                if (off + 11 > buffer.byteLength) break;
+                const tsUs = view.getUint32(off, true); off += 4;
+                const wid = view.getUint8(off); off += 1;
+                const tid = view.getUint32(off, true); off += 4;
+                const src = view.getUint8(off); off += 1; // 0=CpuProfile, 1=SchedEvent
+                const nf = view.getUint8(off); off += 1;
+                if (off + nf * 8 > buffer.byteLength) break;
+                const chain = [];
+                for (let i = 0; i < nf; i++) {
+                    const lo = view.getUint32(off, true);
```
jlizen, line 93: this math is lossy, it would mangle kernel addresses above 2^53.  Should use bigint.

dial9-tokio-telemetry/src/telemetry/recorder.rs
jlizen: In general the conditional compilation is fairly noisy with all its tendrils if there is a better way to consolidate it.

dial9-tokio-telemetry/design/cpu-sampling-and-worker-id-design.md
```diff
@@ -0,0 +1,183 @@
+# CPU Sampling & Worker ID Tracking — Current Design
```
jlizen, line 1: Thanks for this, was very helpful. Though I'm a bit worried it might rot over time. We can keep an eye on it.

dial9-tokio-telemetry/src/telemetry/format.rs
rcoh: probably eventually. for the moment, this is mostly to help AI update the html viewer.

dial9-tokio-telemetry/Cargo.toml
rcoh: ah yeah, good point.

dial9-tokio-telemetry/src/telemetry/recorder.rs
rcoh: yeah I think we probably will remove it internally, and condition it on the ingest side to simplify

PR #13, opened 2026-02-28
Release plz

rcoh:


release-plz.toml
```diff
@@ -0,0 +1,3 @@
+[workspace]
+git_release_enable = true
+changelog_update = true
```
jlizen, line 7: This gets us a changelog per crate rather than at root, is that you preference? I vaguely prefer root if we think we'll have a single facade don't care very much.

jlizen (approved):

release-plz.toml
rcoh: yeah that's a good point. I was initially thinking that the perf crate would have its own changelog but I think in practice we should basically keep it as a private inner crate

PR #30, opened 2026-03-01
Release 0.1.0

rcoh:


jlizen (approved):
🎉looking good!

Main feedback is, README could use a, "can I use this in production?" q&a with an answer of "probably not, it's had some basic benchmarking and has reasonable overhead, but don't blast it across your fleet. putting it on onebox is probably ok though"?

PR #39, opened 2026-03-02
Improve symbolization and include docs.rs links in call frames

rcoh:
before: `<...>::poll`

after: 
Screenshot 2026-03-02 at 10 59 10 AM

I am following up with another CR to include a demo trace that is checked in for demo purposes that will make reviewing changes like this a little easier as well


dial9-tokio-telemetry/trace_viewer/trace_parser.js
```diff
@@ -234,11 +234,137 @@
...
+        return m ? m[1] : null;
+    }
+
+    /**
+     * Format a stack frame for human-readable display.
+     * @param {{symbol: string, location: string|null}} frame
+     * @returns {{text: string, docsUrl: string|null}}
+     */
+    function formatFrame(frame) {
+        if (typeof frame === 'string') return { text: frame, docsUrl: null };
```
jlizen: Are there cases where we would expect a string input to this? The jsdocs describe an object. I'd probably throw an error if no (or else update document this case)

dial9-tokio-telemetry/trace_viewer/index.html
```diff
@@ -366,8 +366,14 @@
         <script src="trace_parser.js"></script>
         <script>
             // Use the external parser
-            const { parseTrace } = TraceParser;
+            const { parseTrace, formatFrame, symbolizeChain, deduplicateSamples } = TraceParser;
 
+            /** Render a raw frame {symbol, location} as HTML, with docs.rs link if available. */
+            function renderFrame(frame) {
+                const f = formatFrame(frame);
+                if (f.docsUrl) return `<a href="${f.docsUrl}" target="_blank" style="color:inherit;text-decoration:underline dotted">${f.text}</a>`;
```
jlizen: this would break from various characters (and is a xss risk). sample fix:

```
 function esc(s) {
      return s.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;');
  }

  function renderFrame(frame) {
      const f = formatFrame(frame);
      if (f.docsUrl) return `<a href="${esc(f.docsUrl)}" target="_blank" style="color:inherit;text-decoration:underline
  dotted">${esc(f.text)}</a>`;
      return esc(f.text);
  }
```

dial9-tokio-telemetry/trace_viewer/trace_parser.js
```diff
@@ -234,11 +234,137 @@
...
+            const m = location.match(/:(\d+)$/);
+            if (m) result += ` ${fileName || ''}:${m[1]}`;
+        }
+        return { text: result, docsUrl: _docsRsUrl(location) };
+    }
+
+    /**
+     * Resolve a callchain (array of address strings) to frame objects.
+     * @param {string[]} callchain - Address strings like "0x55cc6d053893"
+     * @param {Map<string, {symbol: string, location: string|null}>} callframeSymbols
```
jlizen, line 334: nit: bit odd that we require `callframeSymbols` to be passed into `symbolizeChain` and `duplicateSamples`, but then require it to be already processed for `formatFrame`, either is fine but it should be consistent

dial9-tokio-telemetry/trace_viewer/trace_parser.js
```diff
@@ -234,11 +234,137 @@
...
+            if (parts[i] === '{{closure}}') closures++; else break;
+        }
+        const meaningful = parts.length - closures;
+        if (meaningful <= 3) return s;
+        return parts.slice(meaningful - 3).join('::');
+    }
+
+    /**
+     * Try to build a docs.rs source link from a cargo registry location path.
+     * Location looks like: /home/user/.cargo/registry/src/index.crates.io-HASH/hyper-0.14.28/src/client/connect/http.rs:474
```
jlizen: this is pretty brittle, it breaks from vendored dependencies, custom registries, etc. i'd suggest instead broadening to just matching on the path segment:
```
  function _docsRsUrl(location) {
      if (!location) return null;
      // Match any path containing a crate-version directory segment
      const m = location.match(/\/([a-z][a-z0-9_-]*)-(\d+\.\d+[^/]*)\/(.*?)(?::(\d+))?$/);
      if (!m) return null;
      const [, crate_, version, rawPath, line] = m;
      const crateSrc = crate_.replace(/-/g, '_');
      const path = rawPath.replace(/^src\//, '');
      let url = `https://docs.rs/${crate_}/${version}/src/${crateSrc}/${path}.html`;
      if (line) url += `#${line}`;
      return url;
  }

```

PR #52, opened 2026-03-04
Track blocking pool threads in CPU profiler via ThreadRole

rcoh:
## Summary

Refactors the CPU profiler's tid→worker_id mapping to centralize it in SharedState and adds blocking pool thread tracking.

### Changes

**New: Blocking pool detection**
- `on_thread_start` registers all tokio-managed threads as `ThreadRole::Blocking`
- Worker threads overwrite this to `Worker(i)` on first poll event
- Blocking pool samples now get `worker_id=254` (`BLOCKING_WORKER`) instead of `255` (`UNKNOWN_WORKER`)

**Simplification: centralized tid→role mapping**
- Replace `worker_tids: HashMap<u32, usize>` with `thread_roles: HashMap<u32, ThreadRole>` in SharedState
- Remove `tid_to_worker` and `register_worker` from both CpuProfiler and SchedProfiler
- Profilers now yield raw `(tid, callchain, time)` samples — no worker_id knowledge
- `flush_cpu` reads `thread_roles` once per flush and maps tid→worker_id at write time
- Delete `sync_worker_tids` entirely

**Bug fix: child process sample filtering**
- CpuProfiler now filters out samples from child processes via pid check
- perf `inherit` flag causes child process samples to leak into parent's ring buffer
- These addresses are in the child's address space and can never be resolved

### Data flow (before → after)

Before:
```
worker_tids (SharedState) → sync_worker_tids → CpuProfiler.tid_to_worker
                                              → SchedProfiler.tid_to_worker
```

After:
```
thread_roles (SharedState) → flush_cpu reads once → maps tid→worker_id at write time
```

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/events.rs
+++ b/dial9-tokio-telemetry/src/telemetry/events.rs
@@ -2,9 +2,16 @@ use crate::telemetry::task_metadata::{SpawnLocationId, TaskId};
 use serde::Serialize;
 
 /// Sentinel worker_id for events from non-worker threads (encoded as u8 on the wire).
+///
+/// Collides with a real worker index if the runtime has 255 worker threads.
+/// In practice this only affects very large machines; a future wire-format change
+/// (e.g. u16 worker index) would remove the limitation.
 pub const UNKNOWN_WORKER: usize = 255;
 
-/// Sentinel worker_id for events from tokio's blocking thread pool.
+/// Sentinel worker_id for events from tokio's blocking thread pool (encoded as u8 on the wire).
+///
+/// Same collision caveat as [`UNKNOWN_WORKER`]: collides with worker index 254 on machines
+/// with ≥254 runtime worker threads.
 pub const BLOCKING_WORKER: usize = 254;
 
 /// Role of a thread known to the telemetry system.
--- a/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
+++ b/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
@@ -105,14 +105,18 @@ impl TelemetryRecorder {
                 .on_thread_start(move || {
                     // Register as Blocking initially; worker threads will
                     // overwrite this to Worker(i) in resolve_worker_id.
+                    // Note: there is a brief window between thread start and the
+                    // first poll event where worker threads appear as Blocking.
+                    // This is benign — resolve_worker_id corrects it on first poll.
+                    // We use insert (not or_insert) so that a recycled OS tid always
+                    // starts fresh rather than inheriting a stale Worker entry.
                     {
                         let tid = crate::telemetry::events::current_tid();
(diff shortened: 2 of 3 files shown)


dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -103,6 +103,17 @@ impl TelemetryRecorder {
             let s_stop = shared.clone();
             builder
                 .on_thread_start(move || {
+                    // Register as Blocking initially; worker threads will
```
jlizen, line 106: Minor: not a new issue, but this creates a brief race until the first poll event occurs on a worker where it will be misattributed (now as blocking, not unknown). Probably worth documenting.

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -103,6 +103,17 @@ impl TelemetryRecorder {
...
+                    // overwrite this to Worker(i) in resolve_worker_id.
+                    {
+                        let tid = crate::telemetry::events::current_tid();
+                        s_start
+                            .thread_roles
+                            .lock()
+                            .unwrap()
+                            .entry(tid)
+                            .or_insert(crate::telemetry::events::ThreadRole::Blocking);
```
jlizen: Two concerns:

- this map will grow without bounds for long-running services
- if a thread exits, its tid can be reused by OS, and then we might have it falsely attributed as a worker thread rather than blocking

It seems like both could be addressed by removing the entry in on_thread_stop? (Though the second could easily be addressed by removing the existence check before writing)

dial9-tokio-telemetry/src/telemetry/events.rs
```diff
@@ -4,6 +4,19 @@ use serde::Serialize;
 /// Sentinel worker_id for events from non-worker threads (encoded as u8 on the wire).
 pub const UNKNOWN_WORKER: usize = 255;
 
+/// Sentinel worker_id for events from tokio's blocking thread pool.
+pub const BLOCKING_WORKER: usize = 254;
```
jlizen, line 15: Not really a new problem, but both of these sentinel values would collide if there were 254/255 runtime threads (ie on very large machines). We could document this limitation, we could bump it to u16 on the wire, or we could add another `thread_role` field to `CpuSample` on the wire.

Probably just document it for now? And maybe add a `debug_assert!` (not that devo env likely has the same core count as prod anyway)?

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
jlizen, line 769: needs similar assertion for BLOCKING_WORKER

jlizen (approved):
Overall seems fine, main concerns are hygiene of the resolve id map, and then sentinel value collisions, but I'm ok to track separately

PR #54, opened 2026-03-04
Fix thread-local buffer never flushing on drop

rcoh:
## Problem

`on_task_spawn` runs on the calling thread (often the main thread in `block_on`), not a Tokio worker. The thread-local event buffer only flushed on `WorkerPark` or when full (1024 events), so the main thread's buffer — containing just a handful of `TaskSpawn` events — was never drained to the central collector.

This also affects `TaskTerminate` events on worker threads that don't park again after the terminate callback fires during runtime shutdown.

## Fix

- Add `Drop` impl on `ThreadLocalBuffer` that flushes remaining events to the `CentralCollector` when the thread exits
- Store an `Arc<CentralCollector>` in each buffer (set lazily on first `record_event` call) so `Drop` can reach the collector
- Explicitly flush the current thread's buffer in `TelemetryGuard::drop` before the final collector drain, since the main thread's `Drop` won't fire until process exit (after the writer is closed)
- Log a warning if a buffer is dropped with events but no collector
- Update park/unpark validation test to allow a diff of 1 in either direction (the flush can now capture a previously-lost unpark at shutdown)

### DIFF
--- a/dial9-tokio-telemetry/examples/blocking_pool_tracking.rs
+++ b/dial9-tokio-telemetry/examples/blocking_pool_tracking.rs
@@ -1,0 +1,60 @@
+//! Example demonstrating blocking pool tracking.
+//!
+//! Spawns work on tokio's blocking pool via `spawn_blocking`, then writes a
+//! trace. Blocking pool samples should appear with `worker_id = 254`
+//! (BLOCKING_WORKER) instead of 255 (UNKNOWN_WORKER).
+//!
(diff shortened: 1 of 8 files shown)


jlizen (approved):
It would be good to have a regression test that checks that `TaskSpawn` events from the main thread show up in the trace.

I'm approving under the assumption that you have manually validated it, but test would be better.

PR #57, opened 2026-03-05
Docs improvements & fix ui paper cuts

rcoh:
- Add alert when shift+drag is attempted with no CPU samples loaded, so users understand they need cpu-profiling enabled
- Add guard.enable() comment to README cpu-profiling example that uses .build(), and add doc comments to both build() methods noting they start disabled
- Document frame pointer requirement and perf_event_paranoid <= 1 for scheduler events under a new "Requirements" subsection
- Document how CPU samples on known workers help diagnose long polls
- Fix "2. 2." typo in CPU profiling section

### DIFF
--- a/dial9-tokio-telemetry/README.md
+++ b/dial9-tokio-telemetry/README.md
@@ -117,9 +117,7 @@ let (runtime, guard) = TracedRuntime::builder()
     .with_cpu_profiling(CpuProfilingConfig::default())
     .with_sched_events(SchedEventConfig { include_kernel: true })
     .with_inline_callframe_symbols(true)
-    .build(builder, Box::new(writer))?;
-// Note: `.build()` starts disabled — call `guard.enable()` to begin recording,
-// or use `.build_and_start()` to enable immediately.
+    .build_and_start(builder, Box::new(writer))?;
 # Ok(())
 # }
 # #[cfg(not(feature = "cpu-profiling"))]
@@ -138,13 +136,13 @@ This pulls in [`dial9-perf-self-profile`](/perf-self-profile) for `perf_event_op
 rustflags = ["-C", "force-frame-pointers=yes"]
 ```
 
-**Scheduler events** (`with_sched_events`) require `perf_event_paranoid` ≤ 1:
+**`perf_event_paranoid`**: CPU profiling features require `perf_event_paranoid` ≤ 2 for sampling, and ≤ 1 for scheduler event tracking (`with_sched_events`):
 
 ```bash
 # check current value
 cat /proc/sys/kernel/perf_event_paranoid
 
-# allow unprivileged users to collect scheduling events
+# allow CPU sampling and scheduler event tracking
 sudo sysctl kernel.perf_event_paranoid=1
 ```
 
--- a/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
+++ b/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
@@ -371,6 +371,10 @@ impl TracedRuntimeBuilder {
         Ok((runtime, guard))
     }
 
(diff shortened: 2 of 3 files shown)


dial9-tokio-telemetry/trace_viewer/index.html
```diff
@@ -1764,6 +1764,10 @@
             document
                 .getElementById("main-area")
                 .addEventListener("mousedown", (e) => {
+                    if (e.shiftKey && trace && trace.cpuSamples.length === 0) {
+                        alert("No CPU samples in this trace. Enable cpu-profiling to use shift+drag region select for flamegraphs.");
```
jlizen, line 1768: I recommend against using alerts like this, the modal is pretty jarring, would be spammy with multiple misaimed shift-clicks. Maybe show a toast instead?

dial9-tokio-telemetry/trace_viewer/index.html
```diff
@@ -1764,6 +1764,10 @@
...
+                        return;
+                    }
                     if (e.shiftKey && trace && trace.cpuSamples.length > 0) {
```
jlizen: nit: this if clause is now redundant

dial9-tokio-telemetry/README.md
```diff
@@ -118,6 +118,8 @@ let (runtime, guard) = TracedRuntime::builder()
     .with_sched_events(SchedEventConfig { include_kernel: true })
     .with_inline_callframe_symbols(true)
     .build(builder, Box::new(writer))?;
+// Note: `.build()` starts disabled — call `guard.enable()` to begin recording,
```
jlizen: I would suggest refactoring this code to be more copy-pastable, IE either call `guard.enable()` inline, or switch to using `build_and_start()`.

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
jlizen, line 428: Why no doc comment here?

dial9-tokio-telemetry/README.md
```diff
@@ -126,6 +128,30 @@ let (runtime, guard) = TracedRuntime::builder()
...
+
+**Frame pointers**: CPU profile stack traces rely on frame-pointer-based unwinding. Compile your application with frame pointers enabled, otherwise stack traces will be truncated or missing:
+
+```toml
+# .cargo/config.toml
+[build]
+rustflags = ["-C", "force-frame-pointers=yes"]
+```
+
+**Scheduler events** (`with_sched_events`) require `perf_event_paranoid` ≤ 1:
```
jlizen: technically cpu sampling requires <=2 as well? 

Maybe reframe to "CPU profiling features generally require perf_event_paranoid ≤ 1"

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -273,6 +273,9 @@ impl TracedRuntimeBuilder {
         self
     }
 
+    /// Build the traced runtime. Recording starts **disabled** — call
+    /// [`TelemetryGuard::enable`] to begin, or use
+    /// [`build_and_start`](Self::build_and_start).
     pub fn build(
```
jlizen, line 279: Somewhat separate, but I was poking around, and it looks like `build()` sets `task_tracking_enabled: false` and `build_and_start()` sets `task_tracking_enabled: true.`

Probably that should at least be documented, though I would probably unify the twos' defaults? I'm lacking context on why they differ currently.

jlizen (approved):
Good improvements, main concern is the alert UX seems like a bad idea, but not trying to block over it.

I'd also consider documenting `build_and_start()`, as well as the differing task tracking defaults.

PR #60, opened 2026-03-06
S3 reporter

rcoh:


dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -223,6 +232,12 @@ pub struct TelemetryGuard {
     handle: TelemetryHandle,
     stop: Arc<AtomicBool>,
     thread: Option<std::thread::JoinHandle<()>>,
+    #[cfg(feature = "worker")]
```
rcoh: Move these fields into one struct

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -223,6 +232,12 @@ pub struct TelemetryGuard {
     handle: TelemetryHandle,
     stop: Arc<AtomicBool>,
     thread: Option<std::thread::JoinHandle<()>>,
```
rcoh: Rename this to flush thread

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -258,7 +319,18 @@ impl Drop for TelemetryGuard {
...
-        self.handle.recorder.lock().unwrap().flush();
+        // 2. Seal the final segment (.active → .bin)
+        self.handle.recorder.lock().unwrap().seal();
+        // 3. Signal worker to stop and drain remaining segments
+        #[cfg(feature = "worker")]
+        {
+            if let Some(ref stop) = self.worker_stop {
+                stop.store(true, Ordering::Release);
+            }
+            if let Some(t) = self.worker_thread.take() {
```
rcoh: Hmmm. This will block the runtime waiting for shutdown — this isn't the right approach.

The worker should be a task not a thread

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -302,13 +376,21 @@ impl TracedRuntimeBuilder {
         self
     }
 
+    /// Configure an in-process worker that watches for sealed segments
+    /// and uploads them to S3.
+    #[cfg(feature = "worker")]
+    pub fn in_process_worker(mut self, config: crate::worker::WorkerConfig) -> Self {
```
rcoh: Calling this worker is confusing with the tokio worker. Needs to be renamed to something clearer

dial9-tokio-telemetry/src/worker/connection.rs
```diff
@@ -0,0 +1,151 @@
...
+
+#[cfg(feature = "worker-s3")]
+mod inner {
+    use std::time::{Duration, Instant};
+
+    const INITIAL_BACKOFF: Duration = Duration::from_secs(1);
+    const MAX_BACKOFF: Duration = Duration::from_secs(300); // 5 minutes
+
+    /// S3 connection health state.
+    #[derive(Debug)]
```
rcoh: I'm not positive we need this whole state machine. Maybe we should just use the built in s3 retries

dial9-tokio-telemetry/src/worker/connection.rs
```diff
@@ -0,0 +1,151 @@
...
+            match self {
+                Self::Healthy => None,
+                Self::Degraded { backoff, .. } => Some(*backoff),
+            }
+        }
+    }
+}
+
+#[cfg(feature = "worker-s3")]
+pub use inner::*;
```
rcoh: Avoid pub use *

dial9-tokio-telemetry/src/worker/identity.rs
```diff
@@ -0,0 +1,77 @@
...
+/// use dial9_tokio_telemetry::worker::identity::InstanceIdentity;
+///
+/// // Explicit string
+/// let id = InstanceIdentity::explicit("us-east-1/i-0abc123");
+/// assert_eq!(id.resolve(), "us-east-1/i-0abc123");
+/// ```
+pub enum InstanceIdentity {
+    /// User-provided identity string (e.g. "us-east-1/i-0abc123").
+    Explicit(String),
+    /// Auto-detected from hostname.
```
rcoh: What's the point of these variants?

Is this just a string? Should still be a new type though

dial9-tokio-telemetry/src/worker/mod.rs
```diff
@@ -0,0 +1,258 @@
+pub mod sealed;
+#[cfg(feature = "worker")]
```
rcoh: Probably don't need worker as a separate feature from worker-s3

S3 is a feature but worker isn't

dial9-tokio-telemetry/src/worker/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+        /// Set a pre-built S3 transfer manager client.
+        /// When provided, the worker uses this client directly instead of
+        /// building one from `aws_config::load_defaults`.
+        #[cfg(feature = "worker-s3")]
+        pub fn client(mut self, client: aws_sdk_s3_transfer_manager::Client) -> Self {
+            self.client = Some(client);
+            self
+        }
+
+        pub fn build(self) -> Result<WorkerConfig, &'static str> {
```
rcoh: Use bon?

dial9-tokio-telemetry/src/worker/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+}
+
+#[cfg(feature = "worker-s3")]
+fn process_segment(segment: &sealed::SealedSegment, state: &mut Option<S3WorkerState>) {
+    let Some(state) = state.as_mut() else { return };
+    if !state.connection.should_attempt_upload() {
+        return;
+    }
+    let timestamp = chrono_free_timestamp();
+    match state.runtime.block_on(state.uploader.upload_and_delete(segment, &timestamp)) {
```
rcoh: I'm a little worried about this pattern of a regular thread driving runtime but actually I guess it is fine. One problem is any background tasks won't make progress. Restructuring to be a task is probably better.

dial9-tokio-telemetry/src/worker/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+#[cfg(feature = "worker-s3")]
+fn chrono_free_timestamp() -> String {
+    use std::time::SystemTime;
+    let now = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .unwrap_or_default();
+    let secs = now.as_secs();
+    let days = secs / 86400;
+    let time_of_day = secs % 86400;
+    let hours = time_of_day / 3600;
```
rcoh: Just use a crate

dial9-tokio-telemetry/src/worker/s3.rs
```diff
@@ -0,0 +1,483 @@
...
+    #[derive(Default)]
+    pub struct S3ConfigBuilder {
+        bucket: Option<String>,
+        prefix: Option<String>,
+        service_name: Option<String>,
+        instance_path: Option<String>,
+        boot_id: Option<String>,
+    }
+
+    impl S3ConfigBuilder {
```
rcoh: Use bon

dial9-tokio-telemetry/src/background_task/s3.rs
```diff
@@ -0,0 +1,483 @@
...
+            .await
+            .unwrap();
+
+        assert_eq!(
+            key,
+            "traces/checkout-api/us-east-1/i-0abc123/2026-03-05T19-30-00Z-0.bin.gz"
+        );
+
+        // Local file should be deleted
+        assert!(!segment_path.exists());
```
rcoh: Attempt to download the file from s3s

dial9-tokio-telemetry/tests/worker_integration.rs
```diff
@@ -0,0 +1,267 @@
...
+#![cfg(feature = "worker-s3")]
+
+use dial9_tokio_telemetry::telemetry::{RotatingWriter, TracedRuntime};
+use dial9_tokio_telemetry::worker::s3::S3Config;
+use dial9_tokio_telemetry::worker::WorkerConfig;
+use flate2::read::GzDecoder;
+use std::io::Read;
+
+/// Create a transfer manager Client backed by s3s-fs.
+fn fake_s3_client(fs_root: &std::path::Path) -> aws_sdk_s3_transfer_manager::Client {
```
rcoh: Duplicated?

dial9-tokio-telemetry/tests/s3_integration.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    // Create the bucket directory for s3s-fs
+    std::fs::create_dir(s3_root.path().join("test-bucket")).unwrap();
+
+    let tm_client = fake_s3_client(s3_root.path());
+    let raw_client = fake_raw_s3_client(s3_root.path());
+
+    // Small max_file_size to force rotation quickly
+    let writer = RotatingWriter::new(&trace_path, 512, 50 * 1024).unwrap();
+
```
rcoh, line 118: These are nice tests. We should probably make a test harness to easily wrap a tokio runtime in a tokio test to simplify

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,259 @@
...
+**Why not inotify/fswatch?** Adds complexity and platform-specific code. Polling every 1s is simple and sufficient.
+
+### 2. Time-first S3 key layout
+
+**Problem:** The primary access pattern is incident correlation — "what was happening across all services at time T?" This means time should be the first index in the key hierarchy.
+
+**Decision:** Time (date-hour) is the first component after the optional prefix:
+
+```
+{prefix}/{date-hour}/{service}/{instance}/{epoch_secs}-{index}.bin.gz
```
rcoh: Is date-hour the right granularity? What about epohseconds in 10 second chunks e.g.?

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,259 @@
...
+
+**Tradeoff:** Requires reasonable clock sync, but we already need that for trace timestamps.
+
+### 3. Gzip compression
+
+**Problem:** Trace files are large (binary event streams).
+
+**Solution:** Gzip in memory before upload. Trace data is highly compressible (repetitive structures).
+
+**Why gzip not zip?** Simpler, standard `Content-Encoding` header, better compression ratio.
```
rcoh, line 80: Why gzip and not zstd

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,259 @@
...
+
+### 4. Connection state machine
+
+**Problem:** S3 outages shouldn't crash the worker or lose data.
```
rcoh, line 86: Losing data is actually "fine" but running out of disk space is not. How does the design handle this currently?

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,259 @@
...
+## Error Handling
+
+| Error | Action | State Change |
+|-------|--------|--------------|
+| Segment disappeared (evicted by RotatingWriter) | Skip, log debug | None |
+| S3 upload fails (500, timeout, 403) | Log warning, keep file | → Degraded |
+| S3 retry succeeds | Log info | → Healthy |
+| Compression fails | Log error, skip segment | None |
+
+**Never crash.** All errors are logged. Worker continues processing.
```
rcoh: We should also be producing metrics (and probably eventually also put this data into the dial9 traces)

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,259 @@
...
+        .bucket("my-traces")
+        .prefix("prod")
+        .service_name("checkout-api")
+        .instance_path("us-east-1/i-0abc123")
+        .boot_id("unique-boot-id")
+        .build()?)
+    .build()?;
+
+let (runtime, guard) = TracedRuntime::builder()
+    .in_process_worker(worker_config)
```
rcoh: we need to rename in_process_worker to something clearer

dial9-tokio-telemetry/src/worker/s3.rs
```diff
@@ -0,0 +1,423 @@
...
+        /// Upload a sealed segment to S3, then delete the local file on success.
+        ///
+        /// Returns the S3 key of the uploaded object.
+        pub async fn upload_and_delete(
+            &self,
+            segment: &SealedSegment,
+            epoch_secs: u64,
+        ) -> Result<String, Box<dyn std::error::Error + Send + Sync>> {
+            let data = std::fs::read(&segment.path)?;
+            let compressed = gzip_compress(&data)?;
```
rcoh: should we incrementally compress to keep memory usage manageable?

dial9-tokio-telemetry/src/worker/s3.rs
```diff
@@ -0,0 +1,423 @@
...
+
+        /// Upload a sealed segment to S3, then delete the local file on success.
+        ///
+        /// Returns the S3 key of the uploaded object.
+        pub async fn upload_and_delete(
+            &self,
+            segment: &SealedSegment,
+            epoch_secs: u64,
+        ) -> Result<String, Box<dyn std::error::Error + Send + Sync>> {
+            let data = std::fs::read(&segment.path)?;
```
rcoh: this is going to block the runtime, is that ok?

dial9-tokio-telemetry/src/background_task/instance_metadata.rs
```diff
@@ -0,0 +1,72 @@
+//! Machine identity detection for S3 object key paths.
```
rcoh, line 1: call this `instance_metadata`

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -330,11 +331,9 @@ impl Drop for TelemetryGuard {
...
         self.handle.recorder.lock().unwrap().seal();
         // 3. Signal worker to stop and drain remaining segments
-        if let Some(ref stop) = self.worker_stop {
-            stop.store(true, Ordering::Release);
-        }
-        if let Some(t) = self.worker_thread.take() {
-            let _ = t.join();
+        if let Some(w) = self.worker.take() {
+            w.stop.store(true, Ordering::Release);
+            let _ = w.thread.join();
```
rcoh: unify this into a single `worker.shutdown()` method

dial9-tokio-telemetry/src/worker/identity.rs
```diff
@@ -51,22 +41,20 @@ mod tests {
...
     }
 
     #[test]
-    fn explicit_identity_accepts_into_string() {
-        let id = InstanceIdentity::explicit(String::from("dc-west/rack4"));
-        assert_eq!(id.resolve(), "dc-west/rack4");
+    fn new_accepts_owned_string() {
+        let id = InstanceIdentity::new(String::from("dc-west/rack4"));
+        assert_eq!(id.as_str(), "dc-west/rack4");
     }
```
rcoh: these tests are kind of pointless

dial9-tokio-telemetry/README.md
```diff
@@ -195,6 +195,72 @@ See [TRACE_ANALYSIS_GUIDE.md](/dial9-tokio-telemetry/TRACE_ANALYSIS_GUIDE.md) fo
...
+
+    let s3_config = S3Config::builder()
+        .bucket("my-trace-bucket")
+        .prefix("traces")
+        .service_name("my-service")
+        .instance_path("us-east-1/i-0abc123")
+        .boot_id("unique-boot-id")
+        .build();
+
+    let worker_config = WorkerConfig::builder()
```
rcoh: It's still very confusing this is called WorkerConfig because of the pun with tokio worker

dial9-tokio-telemetry/README.md
```diff
@@ -195,6 +195,72 @@ See [TRACE_ANALYSIS_GUIDE.md](/dial9-tokio-telemetry/TRACE_ANALYSIS_GUIDE.md) fo
...
+    runtime.block_on(async {
+        // your async code here
+    });
+
+    // guard drop: flushes, seals final segment, worker drains remaining to S3
+    Ok(())
+}
+```
+
+Objects land at `s3://{bucket}/{prefix}/{date-hour}/{service_name}/{instance_path}/{epoch_secs}-{index}.bin.gz` with metadata headers (`x-amz-meta-service`, `x-amz-meta-boot-id`, etc.) for quick inspection via `HeadObject`.
```
rcoh: I think this needs to be finer grained than date-hour

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,263 @@
...
+
+- `{date-hour}`: `2026-03-07/20` (enables time-range queries across all services)
+- `{service}`: user-provided service name
+- `{instance}`: `us-east-1/i-0abc123` or `dc-west/rack4-host7` (opaque string)
+- `{epoch_secs}`: Unix epoch seconds
+- `{index}`: segment index from RotatingWriter
+
+**Metadata headers:**
+```
+x-amz-meta-service: checkout-api
```
rcoh: Do these all have to start with x-amz-meta?

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,261 @@
...
+                                           │
+                                           ▼
+                        S3: {bucket}/{prefix}/{date-time}/
+                            {service}/{instance}/
+                            {epoch_secs}-{index}.bin.gz
+```
+
+## Key Design Decisions
+
+### 1. Rename-on-seal for atomicity
```
jlizen, line 32: Filename + poll seems good

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,261 @@
...
+Example: `traces/2026-03-07/2030/checkout-api/us-east-1/i-0abc123/1741384542-3.bin.gz`
+
+**Why time-first instead of service-first?**
+
+| Layout | Incident query ("what happened at 8:30pm?") | Single-service query |
+|--------|------------------------------------------|---------------------|
+| `{time}/{service}/...` | `ListObjects(prefix=traces/2026-03-07/2030/)` — one call, all services | `ListObjects(prefix=traces/2026-03-07/2030/checkout-api/)` — still one call |
+| `{service}/{time}/...` | N calls, one per service — must know all service names upfront | `ListObjects(prefix=traces/checkout-api/2026-03-07/2030/)` — one call |
+
+Time-first is strictly better for incident correlation and no worse for single-service queries. The only case where service-first wins is "list all time ranges for one service" — but that's a rare access pattern compared to "what happened during this incident."
```
jlizen, line 64: Seems like we don't want a secondary bucket on service either, if we expect mostly narrow point timeframe queries? Goes to N files for most queries that would be 1?

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,261 @@
...
+
+**Benefits:**
+- Time-range queries across all services with a single `ListObjectsV2` prefix
+- Natural Athena partitioning if we add Parquet output later
+- Efficient S3 lifecycle policies (delete everything older than N days)
```
jlizen, line 69: This would be just reference config? I don't see how partition choices affect lifecycle policies much

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,261 @@
...
+| Layout | Incident query ("what happened at 8:30pm?") | Single-service query |
+|--------|------------------------------------------|---------------------|
+| `{time}/{service}/...` | `ListObjects(prefix=traces/2026-03-07/2030/)` — one call, all services | `ListObjects(prefix=traces/2026-03-07/2030/checkout-api/)` — still one call |
+| `{service}/{time}/...` | N calls, one per service — must know all service names upfront | `ListObjects(prefix=traces/checkout-api/2026-03-07/2030/)` — one call |
+
+Time-first is strictly better for incident correlation and no worse for single-service queries. The only case where service-first wins is "list all time ranges for one service" — but that's a rare access pattern compared to "what happened during this incident."
+
+**Benefits:**
+- Time-range queries across all services with a single `ListObjectsV2` prefix
+- Natural Athena partitioning if we add Parquet output later
```
jlizen, line 68: We should match default athena path structure, probably

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,261 @@
...
+
+### 3. Gzip compression
+
+**Problem:** Trace files are large (binary event streams).
+
+**Solution:** Gzip in memory before upload. Trace data is highly compressible (repetitive structures).
+
+**Why gzip not zip?** Simpler, standard `Content-Encoding` header, better compression ratio.
+
+**Why gzip and not zstd?** Zstd has better compression ratios and speed, but gzip has wider ecosystem support: S3 `Content-Encoding: gzip` is universally understood, every CLI tool can decompress it (`gunzip`, `zcat`), and the `flate2` crate is already a transitive dependency via the AWS SDK. Switching to zstd would add a native C dependency (`zstd-sys`) for marginal gains on files that are typically 1-5 MB. If compression becomes a bottleneck, zstd is a straightforward swap.
```
jlizen, line 82: `gzip` seems fine, we can start to optimize for compression speed if we need to. I prefer a simpler build for most customers.

Presumably users will have metrics if they fall behind on flushing, so can start a discussion about adding a zstd flag (or more likely other optimizations entirely)

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,261 @@
...
+Healthy: upload + delete
+   ↓ (upload fails)
+Degraded: skip uploads, keep files on disk, exponential backoff
+   ↓ (retry succeeds)
+Healthy
+```
+
+Backoff: 1s → 2s → 4s → ... → 5min cap
+
+**Why not crash?** Compressed files on disk are still valuable. Can be manually uploaded or recovered when S3 comes back.
```
jlizen, line 111: Also you want to keep trying in case s3 comes back online

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,261 @@
...
+**Problem:** Not everyone needs S3 upload. AWS SDK is a heavy dependency.
+
+**Solution:** Two tiers:
+
+```toml
+# Core only (no worker)
+dial9-tokio-telemetry = "0.1"
+
+# Worker with S3 upload
+dial9-tokio-telemetry = { version = "0.1", features = ["worker-s3"] }
```
jlizen, line 144: This name is a bit odd, are we expecting future usage of `s3` in separate components?

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,261 @@
...
+
+## API
+
+```rust
+// In-process worker with S3 upload
+let writer = RotatingWriter::new("/tmp/traces/trace.bin", 1_MB, 5_MB)?;
+
+let uploader_config = UploaderConfig::builder()
+    .trace_path("/tmp/traces/trace.bin")
+    .s3(S3Config::builder()
```
jlizen: No custom cred provider? A lot of people have central monitoring accounts. Setting up cross-account sinks is kind of annoying.

Ideally they could just provide their own client, probably, with timeouts etc customized.

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,261 @@
...
+    for segment in sealed {
+        let data = fs::read(&segment.path)?;
+        let compressed = gzip_compress(&data)?;
+        let key = s3_config.object_key(&segment, &timestamp());
+        
+        if connection.should_attempt_upload() {
+            match upload(&key, compressed).await {
+                Ok(_) => {
+                    connection.on_success();
+                    fs::remove_file(&segment.path)?;
```
jlizen: So we crash the worker on failed cleanup? Probably good, not worth the trouble of deduplicating stale files internally.

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,261 @@
...
+## Error Handling
+
+| Error | Action | State Change |
+|-------|--------|--------------|
+| Segment disappeared (evicted by RotatingWriter) | Skip, log debug | None |
+| S3 upload fails (500, timeout, 403) | Log warning, keep file | → Degraded |
+| S3 retry succeeds | Log info | → Healthy |
+| Compression fails | Log error, skip segment | None |
+
+**Never crash.** All errors are logged. Worker continues processing.
```
jlizen: Except failed writes? Or are those logged at another layer with worker loop carrying on?

We would want to make sure we don't spam s3 with a growing number of duplicate object PUTs if for some reason we have write but not rm permissions.

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,261 @@
...
+
+## S3 Object Layout
+
+```
+s3://{bucket}/{prefix}/{date-time}/{service}/{instance}/{epoch_secs}-{index}.bin.gz
+```
+
+- `{date-time}`: `2026-03-07/2030` — 10-minute bucket (enables time-range queries across all services)
```
jlizen: i would suggest hour as partition key definitely, and maybe bucket by minute (or 10 seconds, varies per system probably...)?

Might be nice to have configurable, but also fine to just pick a reasonable default and wait until we have parquet / real query patterns.

dial9-tokio-telemetry/design/s3-worker-design.md
jlizen: (Fine as later scope though)

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,261 @@
...
+- `{service}`: user-provided service name
+- `{instance}`: `us-east-1/i-0abc123` or `dc-west/rack4-host7` (opaque string)
```
jlizen, line 292: do we manage this or provide an api for users to set it?

presumably it automagically works on amazon services that have IMDS/TMDS, and we take CRs for others?

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,261 @@
...
+
+**Key tests:**
+1. End-to-end: RotatingWriter seals → worker uploads to s3s → verify object
+2. Fault injection: s3s returns 500s → worker enters degraded → retries → recovers
+3. Backpressure: slow s3s → RotatingWriter evicts old segments → worker skips gracefully
+4. Compression: upload → download from s3s → decompress → verify roundtrip
+
+## Future Work
+
+- **Metrics:** The worker should emit metrics for upload success/failure counts, upload latency, compression ratios, and circuit breaker state. Eventually these should also be recorded into the dial9 traces themselves for self-monitoring.
```
jlizen: This one could use an issue, seems important

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,261 @@
...
+- **Symbolization:** Embed `/proc/self/maps` in traces, symbolize in worker
+- **Sidecar mode:** Run worker as separate process for blast-radius isolation
+- **Cross-host indexing:** S3 event → Lambda → DynamoDB for "find all traces matching X"
+- **Parquet output:** Convert traces to Parquet for Athena queries
```
jlizen, line 346: Link #26 ?

dial9-tokio-telemetry/src/background_task/s3.rs
```diff
@@ -0,0 +1,473 @@
...
+pub async fn gzip_compress_file(path: &std::path::Path) -> std::io::Result<Vec<u8>> {
+    let mut file = tokio::fs::File::open(path).await?;
+    let mut encoder = GzEncoder::new(Vec::new(), Compression::fast());
+    let mut buf = [0u8; 64 * 1024];
+    loop {
+        let n = file.read(&mut buf).await?;
+        if n == 0 {
+            break;
+        }
+        encoder.write_all(&buf[..n])?;
```
jlizen, line 134: These are gzip operations are sync, should we just do the whole thing on a `spawn_blocking` thread?

dial9-tokio-telemetry/src/background_task/s3.rs
```diff
@@ -0,0 +1,473 @@
...
+
+/// Convert epoch seconds to `YYYY-MM-DD/HHM0` string for S3 key bucketing.
+/// Minutes are truncated to 10-minute buckets (00, 10, 20, 30, 40, 50).
+fn time_bucket_from_epoch(epoch_secs: u64) -> String {
+    let dt = time::OffsetDateTime::from_unix_timestamp(epoch_secs as i64)
+        .unwrap_or(time::OffsetDateTime::UNIX_EPOCH);
+    let format = time::format_description::parse("[year]-[month]-[day]/[hour][minute]")
+        .expect("static format");
+    let mut s = dt.format(&format).expect("format never fails for UTC");
+    // Truncate minutes to 10-minute bucket: replace last digit with '0'
```
jlizen: this is a bit fragile, it requires on single digit minutes having 0 padding.

maybe better as an integer calculation? 
```
  let minute_bucket = (dt.minute() / 10) * 10;  // 35 → 30, 7 → 0, etc.
  format!("{:04}-{:02}-{:02}/{:02}{:02}", dt.year(), dt.month() as u8, dt.day(), dt.hour(), minute_bucket)
```

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+
+    /// Directory containing trace segments.
+    pub fn trace_dir(&self) -> &Path {
+        self.trace_path.parent().unwrap_or(Path::new("."))
+    }
+
+    /// File stem used for segment matching (e.g. "trace" for "trace.0.bin").
+    pub fn trace_stem(&self) -> &str {
+        self.trace_path
+            .file_stem()
```
jlizen: I'd prefer an error if they pass in `/tmp/trace/` (no trailing file), right now it just always falls back to `trace` which seems confusing.

dial9-tokio-telemetry/README.md
```diff
@@ -195,6 +195,72 @@ See [TRACE_ANALYSIS_GUIDE.md](/dial9-tokio-telemetry/TRACE_ANALYSIS_GUIDE.md) fo
...
 - **`task-dump`** — Enables Tokio's `taskdump` feature for async stack traces. Required for the `long_sleep`, `completing_task`, `cancelled_task`, and `debug_timing` examples.
+- **`worker-s3`** — Enables S3 upload support. Adds `aws-sdk-s3`, `aws-sdk-s3-transfer-manager`, `aws-config`, and `flate2`.
+
+## S3 upload
+
+With the `worker-s3` feature, sealed trace segments are automatically gzip-compressed and uploaded to S3 by a background worker thread. The application process is unaffected — uploads happen asynchronously after segments are sealed.
+
+```rust,ignore
+use dial9_tokio_telemetry::telemetry::{RotatingWriter, TracedRuntime};
+use dial9_tokio_telemetry::background_task::UploaderConfig;
```
jlizen: In the code it is: BackgroundTaskConfig.

Can we `no_run` this?

dial9-tokio-telemetry/src/background_task/s3.rs
```diff
@@ -0,0 +1,473 @@
...
+use std::io::Write;
+use tokio::io::AsyncReadExt;
+
+/// Configuration for S3 uploads.
+#[derive(Clone, bon::Builder)]
+#[builder(on(String, into))]
+pub struct S3Config {
+    bucket: String,
+    service_name: String,
+    instance_path: String,
```
jlizen: Should accept `InstanceIdentity` right?

dial9-tokio-telemetry/src/background_task/s3.rs
```diff
@@ -0,0 +1,473 @@
...
+    let mut s = dt.format(&format).expect("format never fails for UTC");
+    // Truncate minutes to 10-minute bucket: replace last digit with '0'
+    let last = s.len() - 1;
+    s.replace_range(last..=last, "0");
+    s
+}
+
+/// Gzip-compress by streaming from an async reader in chunks.
+/// Avoids holding the entire uncompressed file in memory.
+pub async fn gzip_compress_file(path: &std::path::Path) -> std::io::Result<Vec<u8>> {
```
jlizen: should be pub(crate)

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+}
+
+/// Detect the region of an S3 bucket via HeadBucket.
+///
+/// On success the response includes the region directly. On a 301 redirect
+/// (wrong region) the `x-amz-bucket-region` header contains the correct one.
+#[cfg(feature = "worker-s3")]
+async fn detect_bucket_region(client: &aws_sdk_s3::Client, bucket: &str) -> String {
+    match client.head_bucket().bucket(bucket).send().await {
+        Ok(resp) => resp.bucket_region().unwrap_or("us-east-1").to_owned(),
```
jlizen: should we warn on this fallback too?

dial9-tokio-telemetry/tests/s3_integration.rs
```diff
@@ -0,0 +1,482 @@
...
+use dial9_tokio_telemetry::telemetry::{RotatingWriter, TracedRuntime};
+use flate2::read::GzDecoder;
+use std::io::Read;
+
+/// Create an aws_sdk_s3::Client backed by s3s-fs (in-memory fake S3).
+///
+/// NOTE: This helper is duplicated in src/background_task/s3.rs unit tests.
+/// Rust's test compilation model prevents sharing between unit tests (compiled
+/// with #[cfg(test)] in src/) and integration tests (compiled from tests/).
+/// A shared test-support crate would fix this but is overkill for now.
```
jlizen: presumably we would do `test-util` feature rather than shared support crate? fine to dupe for now

dial9-tokio-telemetry/src/background_task/sealed.rs
```diff
@@ -0,0 +1,280 @@
...
+    #[allow(dead_code)]
+    fn parse_segment_timestamp(&self) -> std::io::Result<u64> {
+        let file = File::open(&self.path)?;
+        let mut reader = BufReader::new(file);
+
+        // Read and validate header
+        let (_magic, _version) = format::read_header(&mut reader)?;
+
+        // Try to read the first event, which should be SegmentMetadata
+        if let Some(event) = format::read_event(&mut reader)?
```
jlizen, line 64: (relates to #67 )

dial9-tokio-telemetry/src/background_task/s3.rs
```diff
@@ -0,0 +1,473 @@
...
+    }
+
+    /// Optional region override for the S3 client.
+    pub(crate) fn region(&self) -> Option<&str> {
+        self.region.as_deref()
+    }
+
+    /// Build the S3 object key for a sealed segment.
+    ///
+    /// Format: `{prefix}/{date-hour}/{service}/{instance}/{epoch_secs}-{index}.bin.gz`
```
jlizen: should be {date}/{HHM0}

scripts/integ-test-s3-upload.sh
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env bash
+# Integration test: runs the metrics-service with S3 upload enabled,
+# then periodically checks that trace segments land in S3 and can be
+# analyzed.
+set -euo pipefail
+
+REPO_ROOT="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)"
+cd "$REPO_ROOT"
+
+BUCKET="dial9-integ-test-$$-$(date +%s)"
```
jlizen, line 10: I would rather the bucket be deterministic and from an ENV. Right now we are on the hook for making sure teardown logic always runs successfully, which seems fragile. Also it's kind of nice to have failed run data available to look at.

To me permanent bucket + fairly short lifecycle on objects seems like the sweet spot.

scripts/integ-test-s3-upload.sh
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env bash
```
jlizen, line 1: Looks good, adding to CI will be a separate PR?

dial9-tokio-telemetry/design/s3-worker-design.md
jlizen: Left feedback on this. My knee-jerk was also 10 second chunks (for bucketing) with hour as a partition key. I could see an argument for minute too. It seems like it depends a lot on sample rate / host count?

rcoh:
Load testing revealed a bug in the interaction between the rotating writer and the uploader I'm looking into it but will hold off merging

scripts/integ-test-s3-upload.sh
jlizen: Same feedback on this @rcoh  (though i recognize it's more annoying for local usage, maybe we should have a CI mode and a local mode)

Anyway not a blocking concern, we can see how this one goes...

dial9-tokio-telemetry/Cargo.toml
```diff
@@ -22,20 +22,40 @@ serde_json = "1"
...
 tracing = "0.1.44"
+hostname = "0.4"
+bon = "3"
+aws-sdk-s3-transfer-manager = { version = "0.1.3", optional = true, default-features = false }
+aws-sdk-s3 = { version = "1", default-features = false, optional = true, features = ["behavior-version-latest", "sigv4a", "rt-tokio", "default-https-client"] }
+aws-config = { version = "1", optional = true, default-features = false }
+flate2 = { version = "1", optional = true }
+time = { version = "0.3", features = ["formatting", "macros"], optional = true }
+metrique-writer = "0.1"
+metrique = { version = "0.1.20", features = ["local-format"] }
```
jlizen, line 34: we're adding these as non-optional because we expect to later use them outside of s3 worker context?

dial9-tokio-telemetry/src/background_task/s3.rs
```diff
@@ -0,0 +1,593 @@
...
+use flate2::write::GzEncoder;
+use std::collections::HashMap;
+use std::io::Write;
+use std::sync::Arc;
+
+/// Trait for custom S3 object key generation.
+///
+/// Implement this to control the S3 key layout. The default key layout is
+/// `{prefix}/{date}/{HHMM}/{service}/{instance}/{epoch}-{index}.bin.gz`.
+pub trait S3KeyFn: Send + Sync {
```
jlizen, line 20: This seems good, with a reasonable default. I'd suggest saying `epoch-secs` here like you do in the other comment.

It looks like no [bucketing](https://docs.aws.amazon.com/athena/latest/ug/ctas-partitioning-and-bucketing-what-is-bucketing.html) yet, huh? Seems fine. We can iterate on it. Path seems like the more important thing to get right currently IMO, and it LGTM.

dial9-tokio-telemetry/src/background_task/s3.rs
```diff
@@ -0,0 +1,593 @@
...
+        dt.month() as u8,
+        dt.day(),
+        dt.hour(),
+        dt.minute()
+    )
+}
+
+/// Gzip-compress a file synchronously. Intended for use with `spawn_blocking`.
+#[cfg(test)]
+pub(crate) fn gzip_compress_file_sync(path: &std::path::Path) -> std::io::Result<Vec<u8>> {
```
jlizen, line 124: This is only used in our tests, why not use `gzip_compress_bytes` for them instead and rip it out?

dial9-tokio-telemetry/src/background_task/connection.rs
```diff
@@ -0,0 +1,132 @@
...
+//! while this provides higher-level circuit-breaking across requests.
+
+use std::time::{Duration, Instant};
+
+const INITIAL_BACKOFF: Duration = Duration::from_secs(1);
+const MAX_BACKOFF: Duration = Duration::from_secs(300); // 5 minutes
+
+/// Circuit breaker for S3 upload attempts.
+#[derive(Debug, Default)]
+pub enum CircuitBreaker {
```
jlizen, line 15: The handrolled breaker seems fine for now, but if we start to want to have things like consecutive failures before opening, half-open state for probing failures, jitter, etc, we could look at something off the shelf with a light footprint like `recloser`.

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,346 @@
...
+
+**Problem:** Trace files are large (binary event streams).
+
+**Solution:** Gzip in memory before upload. Trace data is highly compressible (repetitive structures). Compression runs via `tokio::task::spawn_blocking` to keep the worker's event loop responsive.
+
+**Why gzip not zip?** Simpler, standard `Content-Encoding` header, better compression ratio.
+
+**Why gzip and not zstd?** Zstd has better compression ratios and speed, but gzip has wider ecosystem support: S3 `Content-Encoding: gzip` is universally understood, every CLI tool can decompress it (`gunzip`, `zcat`), and the `flate2` crate is already a transitive dependency via the AWS SDK. Switching to zstd would add a native C dependency (`zstd-sys`) for marginal gains on files that are typically 1-5 MB. If compression becomes a bottleneck, zstd is a straightforward swap.
+
+### 4. Circuit breaker
```
jlizen, line 84: Seems good to have but not seeing the benefit of handrolling it

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -0,0 +1,912 @@
...
+                        } else {
+                            tracing::warn!(target: "dial9_worker", error = %e.kind, cause = ?e.kind, path = %segment.path.display(), "processor failed, skipping segment");
+                        }
+                        continue 'next_segment;
+                    }
+                }
+            }
+
+            data.metrics.success = 1;
+            data.metrics.total_time.stop();
```
jlizen, line 434: It stops automatically on drop anyway right? Or does the struct live on somewhere?

dial9-tokio-telemetry/design/s3-worker-design.md
```diff
@@ -0,0 +1,346 @@
...
+| Compression fails | Log error, skip segment | None |
+| Circuit breaker open | Skip upload entirely | None (wait for backoff timer) |
+
+**Never crash.** All errors are logged via `tracing`. Worker continues processing.
+
+Per-segment metrics are recorded regardless of success or failure — the `SegmentData` carries a metrics guard that flushes on drop.
+
+## Metrics
+
+The worker emits per-segment metrics via `metrique`:
```
jlizen, line 268: Probably worth noting that the default destination is the dev null sink, unless opted into

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -0,0 +1,912 @@
...
+
+    async fn drain(&mut self) {
+        tracing::info!(target: "dial9_worker", "draining remaining segments");
+        if let Ok(segments) = sealed::find_sealed_segments(&self.dir, &self.stem) {
+            tracing::info!(target: "dial9_worker", count = segments.len(), "found segments to drain");
+            self.process_segments(&segments).await;
+        }
+    }
+
+    async fn process_segments(&mut self, segments: &[sealed::SealedSegment]) {
```
jlizen, line 358: This function claims to be async, but has `std::fs::metadata` and `std::fs::read` calls in it. Is that intentional?

dial9-tokio-telemetry/src/background_task/connection.rs
jlizen: I might consider more like 2 or 3 failures before opening the breaker, as a simple tweak that doesn't involve using a library. 1 failures seems fragile. I guess, for that, we are expecting s3 client retries to add some resilience?

jlizen (approved):
Looks good. The refactor seems reasonable. Some small feedback.

dial9-tokio-telemetry/src/background_task/s3.rs
jlizen: Oh, one question is, do we have some guarantee that all of a file will fall within a specific minute? Might get fairly confusing to query if not... In that case, we might want to stick to only hour. 

Certainly some users might want the narrower partitions and would eagerly check ahead, but it's much easier to do that within a partition than across.

dial9-tokio-telemetry/src/background_task/connection.rs
rcoh: there are already actually retries happening in the sdk so this only triggers when we have already seen a lot of consecutive failures

jlizen: Gotcha, seems fine, might merit a note explaining that. Sounds like half-open state and jitter would be the main reasons to cut to a library with batteries, but doesn't seem warranted yet.

dial9-tokio-telemetry/src/background_task/mod.rs
rcoh: will add comment - we are inside a current thread runtime so it doesn't actually matter. if we used spawn_blocking its just going to spawn another thread. worth reconsidering at some point. The main impact is that graceful shutdown can't cancel these fs operations

PR #61, opened 2026-03-06
Implement trace format

rcoh:
Adds format crate for dial9

### DIFF
--- a/dial9-tokio-telemetry/Cargo.toml
+++ b/dial9-tokio-telemetry/Cargo.toml
@@ -21,7 +21,6 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 smallvec = "1"
 dial9-perf-self-profile = { workspace = true, optional = true }
-dial9-trace-format = { path = "../dial9-trace-format" }
 tracing = "0.1.44"
 
 [features]


dial9-trace-format/src/encoder.rs
```diff
@@ -0,0 +1,154 @@
...
+    registry: SchemaRegistry,
+    string_pool: HashMap<String, u32>,
+    next_pool_id: u32,
+    // Small linear cache: faster than HashMap+SipHash for the typical <20 event types.
+    type_ids: Vec<(TypeId, u16)>,
+}
+
+impl Encoder {
+    pub fn new() -> Self {
+        let mut buf = Vec::with_capacity(1024 * 1024);
```
rcoh: This should be configurable

dial9-trace-format/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+
+pub use dial9_trace_format_derive::TraceEvent;
+pub use types::InternedString;
+pub use types::StackFrames;
+pub use types::TraceField;
+
+use schema::{FieldDef, SchemaEntry};
+
+/// Trait implemented by `#[derive(TraceEvent)]` for compile-time event types.
+pub trait TraceEvent {
```
rcoh, line 18: Add a decode method may need an associated type for borrowing from the stream

dial9-trace-format/src/schema.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+impl SchemaRegistry {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn register(&mut self, entry: SchemaEntry) -> Result<(), String> {
+        if self.schemas.contains_key(&entry.type_id) {
+            return Err(format!("type_id {} already registered", entry.type_id));
```
rcoh: This probably shouldn't error if the type is the same

dial9-trace-format/src/types.rs
```diff
@@ -0,0 +1,591 @@
...
+    I64 = 1,
+    F64 = 2,
+    Bool = 3,
+    String = 4,
+    Bytes = 5,
+    U64Array = 6,
+    PooledString = 7,
+    StackFrames = 8,
+    Varint = 9,
+    StringMap = 10,
```
rcoh, line 19: StringMap needs to be added to readme and possibly also the spec

dial9-trace-format/src/types.rs
```diff
@@ -0,0 +1,591 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u8)]
+pub enum FieldType {
+    U64 = 0,
+    I64 = 1,
+    F64 = 2,
+    Bool = 3,
+    String = 4,
+    Bytes = 5,
+    U64Array = 6,
```
rcoh: We have bytes we should drop U64Array

dial9-trace-format/src/types.rs
```diff
@@ -0,0 +1,591 @@
...
+pub struct StackFrames(pub Vec<u64>);
+
+/// An interned string reference (pool ID). Created by [`Encoder::intern_string`].
+/// On the wire this is a `PooledString` (u32 LE).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct InternedString(pub(crate) u32);
+
+impl InternedString {
+    /// Returns the raw pool ID for wire encoding.
+    pub fn pool_id(self) -> u32 { self.0 }
```
rcoh: Do we actually need this accessor to be pub? Seems like we should just use the new type everywhere

dial9-trace-format/src/types.rs
```diff
@@ -0,0 +1,591 @@
...
+}
+#[derive(Debug, Clone, PartialEq)]
+pub enum FieldValue {
```
rcoh, line 46: Non exhaustive

dial9-trace-format/src/types.rs
```diff
@@ -0,0 +1,591 @@
...
+                    buf.extend_from_slice(k);
+                    buf.extend_from_slice(&(v.len() as u32).to_le_bytes());
+                    buf.extend_from_slice(v);
+                }
+            }
+        }
+    }
+
+    /// Decode a value of the given type from the buffer at offset. Returns (value, bytes_consumed).
+    pub fn decode(field_type: FieldType, data: &[u8], offset: usize) -> Option<(FieldValue, usize)> {
```
rcoh: Alternatively can this return a new slice?

dial9-trace-format/src/types.rs
```diff
@@ -0,0 +1,591 @@
...
+}
+
+/// Zero-copy field value that borrows from the input buffer.
+#[derive(Debug, Clone, PartialEq)]
+pub enum FieldValueRef<'a> {
+    U64(u64),
+    I64(i64),
+    F64(f64),
+    Bool(bool),
+    String(&'a [u8]),
```
rcoh: We should decode into str

dial9-trace-format/src/types.rs
```diff
@@ -0,0 +1,591 @@
...
+    Bytes(&'a [u8]),
+    U64Array(&'a [u8]),
+    PooledString(u32),
+    /// Raw delta-encoded stack frame bytes. Use [`StackFrameIter`] to decode.
+    StackFrames(&'a [u8], u32),
```
rcoh: Can you put a new type in here?

dial9-trace-format/src/types.rs
```diff
@@ -0,0 +1,591 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub enum FieldValueRef<'a> {
+    U64(u64),
+    I64(i64),
+    F64(f64),
+    Bool(bool),
+    String(&'a [u8]),
+    Bytes(&'a [u8]),
+    U64Array(&'a [u8]),
+    PooledString(u32),
```
rcoh: Can you put a new type in here?

dial9-trace-format/src/types.rs
```diff
@@ -0,0 +1,591 @@
...
+    /// Raw delta-encoded stack frame bytes. Use [`StackFrameIter`] to decode.
+    StackFrames(&'a [u8], u32),
+    Varint(u64),
+    StringMap(&'a [u8], u32),
```
rcoh: This should also be a new type wrapper so you can decode easily

dial9-trace-format/src/types.rs
```diff
@@ -0,0 +1,591 @@
...
+    }
+
+    #[test]
+    fn encode_decode_stack_frames() {
+        let addrs = vec![0x5555_5555_1234u64, 0x5555_5555_0a00, 0x5555_5555_0800, 0x5555_5555_0100];
+        let val = FieldValue::StackFrames(addrs.clone());
+        let mut buf = Vec::new();
+        val.encode(&mut buf);
+        // Should be much smaller than 4 + 4*8 = 36 bytes
+        assert!(buf.len() < 36, "stack frames should be compact, got {} bytes", buf.len());
```
rcoh: This should be a deterministic number

progress/trace-format-progress.md
```diff
@@ -0,0 +1,48 @@
+# dial9-trace-format Progress
+
+## Status: COMPLETE ✅
```
rcoh: Delete this

dial9-tokio-telemetry/src/telemetry/analysis.rs
```diff
@@ -235,6 +237,7 @@ pub fn analyze_trace(events: &[TelemetryEvent]) -> TraceAnalysis {
             }
             TelemetryEvent::SpawnLocationDef { .. }
             | TelemetryEvent::TaskSpawn { .. }
+            | TelemetryEvent::TaskTerminate { .. }
```
rcoh, line 249: Please merge from main to remove these unrelated changes

dial9-trace-format/README.md
```diff
@@ -0,0 +1,142 @@
...
+let mut enc = Encoder::new();
+enc.register_schema_for::<HttpRequest>("HttpRequest", vec![
+    FieldDef { name: "timestamp_ns".into(), field_type: FieldType::Varint },
+    FieldDef { name: "method".into(), field_type: FieldType::String },
+    FieldDef { name: "status".into(), field_type: FieldType::Varint },
+]);
+
+enc.write_event_for::<HttpRequest>(&[
+    FieldValue::Varint(1_000_000),
+    FieldValue::String(b"GET".to_vec()),
```
rcoh: You shouldn't need an owned value here, need to be able to pass &str

dial9-trace-format/README.md
```diff
@@ -0,0 +1,142 @@
...
+```rust
+use dial9_trace_format::encoder::Encoder;
+use dial9_trace_format::codec::SymbolEntry;
+
+let mut enc = Encoder::new();
+// Intern strings for deduplication (returns a pool ID)
+let id = enc.intern_string("my_function");
+
+// Attach symbol table for stack frame symbolization
+enc.write_symbol_table(&[
```
rcoh, line 142: Confused by this

dial9-trace-format/SPEC.md
```diff
@@ -0,0 +1,229 @@
+# dial9-trace-format Binary Specification
+
+Version: 2
```
rcoh: Version one was never released call it v1

dial9-trace-format/SPEC.md
```diff
@@ -0,0 +1,229 @@
...
+| Field | Type | Description |
+|-------|------|-------------|
+| tag | u8 | `0x02` |
+| type_id | u16 | References a schema's `type_id` |
+| timestamp_delta_ns | u24 | Nanosecond delta from the current timestamp base (3 bytes LE) |
+| values | ... | Field values, encoded in schema field order |
+
+The `timestamp_delta_ns` is a 24-bit unsigned integer (0–16,777,215) representing nanoseconds elapsed since the most recent Timestamp Reset frame (or 0 if none has been seen). This gives ~16.7 ms of range per reset. The encoder **must** emit a Timestamp Reset frame before any event whose delta would exceed 16,777,215 ns.
+
+Because the timestamp is packed into the event header rather than delta-encoded across events, there is **no accumulation error** — each event's timestamp is independently computed as `base + delta_ns`.
```
rcoh: This doesn't make sense

dial9-trace-format/benches/leb128.rs
```diff
@@ -0,0 +1,151 @@
+use criterion::{Criterion, black_box, criterion_group, criterion_main};
+use dial9_trace_format::leb128;
+
+// Also test alternative impls directly in the bench
+mod alt {
```
rcoh: We can delete this do we even leb anything anymore?

dial9-trace-format-derive/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+        "u32" => quote! { enc.write_u32(self.#field_name); },
+        "u64" => quote! { enc.write_u64(self.#field_name); },
+        "i64" => quote! { enc.write_i64(self.#field_name); },
+        "f64" => quote! { enc.write_f64(self.#field_name); },
+        "bool" => quote! { enc.write_bool(self.#field_name); },
+        "String" => quote! { enc.write_string(&self.#field_name); },
+        "Vec<u8>" => quote! { enc.write_bytes(&self.#field_name); },
+        "InternedString" => quote! { enc.write_interned(self.#field_name); },
+        "StackFrames" => quote! { enc.write_stack_frames(&self.#field_name); },
+        "Timestamp" => quote! { enc.write_timestamp(&self.#field_name); },
```
rcoh: Right now these are brittle slightly because they match field names instead of using a trait. At least add a todo

dial9-trace-format/README.md
```diff
@@ -0,0 +1,153 @@
...
+        }
+        DecodedFrame::StringPool(entries) => println!("{} pool entries", entries.len()),
+        DecodedFrame::SymbolTable(entries) => println!("{} symbols", entries.len()),
+    }
+}
+```
+
+The `#[traceevent(timestamp)]` attribute marks a `u64` field as the event's timestamp. It is encoded as a u24 nanosecond delta in the event header (not as a regular field), giving nanosecond precision with no accumulation error. The encoder automatically emits `TimestampReset` frames when the delta exceeds ~16.7 ms.
+
+Integer fields (`u8`, `u16`, `u32`, `u64`) are encoded as LEB128 varints. `StackFrames` fields are delta-encoded. The derive macro handles the mapping automatically.
```
rcoh: This isn't true anymore right?

dial9-trace-format/README.md
```diff
@@ -0,0 +1,170 @@
+# dial9-trace-format
+
+A self-describing binary trace format. Schemas are embedded in the stream so readers don't need out-of-band type definitions. Designed for tokio runtime telemetry but usable for any structured event stream.
+
+See [SPEC.md](SPEC.md) for the wire format specification.
+
+## Design principles
+
```
jlizen, line 8: I might move the self-describing pitch from the top-level, into design principals

dial9-trace-format/README.md
```diff
@@ -0,0 +1,170 @@
...
+1. **Relatively compact.** Events average ~15 bytes raw and ~3 bytes gzipped on a real-world tokio trace. The format isn't the absolute smallest possible — it trades a few bytes of overhead for the properties below.
```
jlizen: Too many emdashes in this.

dial9-trace-format/README.md
```diff
@@ -0,0 +1,170 @@
...
+2. **Extremely fast to write.** The encoder does no allocations per event and uses fixed-width or LEB128 fields with no framing overhead beyond a 3-byte header. Benchmarks show ~48M events/s encode throughput on a single core.
```
jlizen: `LEB128` is a bit opaque without a link or a non-acronym first-usage

dial9-trace-format/README.md
```diff
@@ -0,0 +1,170 @@
...
+3. **Compressible.** Schemas are written once; events are pure data with no repeated field names or tags. Timestamps are delta-encoded as 3-byte offsets. This structure compresses well — gzip reduces a typical trace to ~20% of its raw size (and beats a hand-tuned bespoke format by 1.4x after compression).
```
jlizen: Nice. Sounds like these strategies would transfer nicely to off the shelf parquet. (Aside from the concerns about lack of streaming writes)

dial9-trace-format/README.md
```diff
@@ -0,0 +1,170 @@
...
+## Numbers
+
+On a 42k-event tokio runtime trace (`format_comparison` example, release mode):
+
+| | Raw | Gzipped |
+|---|---|---|
+| dial9-trace-format | 632 KB (14.8 B/event) | 129 KB (3.0 B/event) |
+| Hand-tuned bespoke format | 586 KB (13.7 B/event) | 177 KB (4.1 B/event) |
+
+The self-describing format is ~8% larger raw but **37% smaller after gzip** because its regular structure compresses better than the bespoke format's variable-length tag soup.
```
jlizen, line 24: awesome to see size go down even as we add more metadata, nice job

dial9-trace-format/README.md
```diff
@@ -0,0 +1,170 @@
...
+1. **Relatively compact.** Events average ~15 bytes raw and ~3 bytes gzipped on a real-world tokio trace. The format isn't the absolute smallest possible — it trades a few bytes of overhead for the properties below.
+2. **Extremely fast to write.** The encoder does no allocations per event and uses fixed-width or LEB128 fields with no framing overhead beyond a 3-byte header. Benchmarks show ~48M events/s encode throughput on a single core.
+3. **Compressible.** Schemas are written once; events are pure data with no repeated field names or tags. Timestamps are delta-encoded as 3-byte offsets. This structure compresses well — gzip reduces a typical trace to ~20% of its raw size (and beats a hand-tuned bespoke format by 1.4x after compression).
+4. **Simple enough to port.** The entire wire format is ~5 frame types, LEB128 integers, and little-endian fixed-width fields. A [JavaScript decoder](js/decode.js) is under 200 lines. No code generation or IDL compiler is needed to read a stream — schemas are inline.
+
+## Numbers
+
+On a 42k-event tokio runtime trace (`format_comparison` example, release mode):
+
+| | Raw | Gzipped |
```
jlizen, line 19: might be nice to add a rough disk space estimation at X default rotation speed

dial9-trace-format/README.md
```diff
@@ -0,0 +1,170 @@
...
+| Encode | ~48M |
+| Decode (visitor, zero-alloc) | ~30M |
+| Decode (zero-copy ref) | ~7M |
+| Decode (owned) | ~6M |
+
+## Usage
+
+### Derive macro
+
+For event types known at compile time, use `#[derive(TraceEvent)]`:
```
jlizen, line 41: Glad to see this macro, it will make it way more practical to manage.

dial9-trace-format/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+```rust
+use dial9_trace_format::{TraceEvent, StackFrames};
+use dial9_trace_format::encoder::Encoder;
+use dial9_trace_format::decoder::{Decoder, DecodedFrame};
+
+#[derive(TraceEvent)]
```
jlizen: The order of these fields is load bearing, i assume? So that we can add new columns?

Might be good to have snapshot testing to assert against changes, and otherwise document it more clearly.

dial9-trace-format/README.md
```diff
@@ -0,0 +1,170 @@
...
+```rust
+use dial9_trace_format::encoder::Encoder;
+use dial9_trace_format::schema::FieldDef;
+use dial9_trace_format::types::{FieldType, FieldValue};
+
+struct HttpRequest; // marker type
+
+let mut enc = Encoder::new();
+enc.register_schema_for::<HttpRequest>("HttpRequest", vec![
+    FieldDef { name: "timestamp_ns".into(), field_type: FieldType::Varint },
```
jlizen, line 107: might be a bit better example with anything dynamic to show why you want runtime initialization

dial9-trace-format/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+### String interning
+
+Frequently-repeated strings can be interned to avoid encoding them multiple times:
+
+```rust
+use dial9_trace_format::encoder::Encoder;
+
+let mut enc = Encoder::new();
+let id = enc.intern_string("my_function");
```
jlizen, line 128: Does this mean we are only interning on a per-batch basis? I would have expected globally scoped interning being supported too (at which point there could maybe be derive macro support?) Or, is that intended to be added in another layer and passed to the encoder by hand?

Also we probably want to support interning dynamic `String`s, added a note on the fn impl.

dial9-trace-format/src/encoder.rs
```diff
@@ -0,0 +1,411 @@
...
+        }
+        let mut enc = EventEncoder::new(&mut self.state);
+        for v in field_values {
+            enc.write_field_value(v)?;
+        }
+        Ok(())
+    }
+
+    /// Intern a string, emitting a pool frame if new. Returns an [`InternedString`] handle.
+    pub fn intern_string(&mut self, s: &str) -> io::Result<crate::types::InternedString> {
```
jlizen, line 207: Shouldn't this support interning dynamic strings too? A lot of repetitive ones are runtime-generated, I assume

dial9-trace-format/README.md
jlizen: it might be nice to have a `test-util` helper with some heuristic based tests to suggest new interned fields or something.

dial9-trace-format/SPEC.md
```diff
@@ -0,0 +1,213 @@
...
+
+Each **FieldDef**:
+
+| Field | Type | Description |
+|-------|------|-------------|
+| name_len | u16 | Length of field name in bytes |
+| name | [u8; name_len] | UTF-8 field name |
+| field_type | u8 | Field type tag (see Field Types) |
+
+A `type_id` **must not** be registered more than once in a stream.
```
jlizen: Currently in the impl you can register it again IFF it uses the same type_id as before. Is that intended? It seems like it probably should be an error.

If not, we should update docs to match.

dial9-trace-format/SPEC.md
```diff
@@ -0,0 +1,213 @@
...
+
+The `has_timestamp` flag indicates whether events of this type include a packed nanosecond timestamp in the event frame header. When set, the timestamp is encoded in the event header (see Event Frame) and is **not** included in the field list. The schema's `field_count` and `fields` describe only the non-timestamp payload fields.
```
jlizen, line 70: > When set, the timestamp is encoded in the event header (see Event Frame) and is **not** included in the field list

Does that mean that it is injected in as a field if `has_timestamp` is false?

This is a bit confusing, I might prefer to just have a timestamp field type that is hoisted to the header when set. That way user can set the name, etc.

dial9-trace-format/SPEC.md
```diff
@@ -0,0 +1,213 @@
...
+| fields | [FieldDef; field_count] | Field definitions (see below) |
+
+Each **FieldDef**:
+
+| Field | Type | Description |
+|-------|------|-------------|
+| name_len | u16 | Length of field name in bytes |
+| name | [u8; name_len] | UTF-8 field name |
+| field_type | u8 | Field type tag (see Field Types) |
+
```
jlizen, line 67: These two notes belong underneath the envelope layout table.

dial9-trace-format/SPEC.md
```diff
@@ -0,0 +1,213 @@
...
+| tag | u8 | `0x02` |
+| type_id | u16 | References a schema's `type_id` |
+| timestamp_delta_ns | u24 | Nanosecond delta from the current timestamp base (3 bytes LE) |
+| values | ... | Field values, encoded in schema field order |
+
+The `timestamp_delta_ns` is a 24-bit unsigned integer (0–16,777,215) representing nanoseconds elapsed since the current timestamp base. This gives ~16.7 ms of range per reset. The encoder **must** emit a Timestamp Reset frame before any event whose delta would exceed 16,777,215 ns or whose timestamp is earlier than the current base.
+
+Each event's absolute timestamp is computed as `base + delta_ns`. After decoding a timestamped event, the decoder **must** set `timestamp_base_ns = base + delta_ns` (i.e., advance the base to the event's absolute timestamp). This keeps inter-event deltas small, which is critical for compression.
+
+The decoder **must** know the schema for `type_id` to determine how many fields to read and their types. If the schema is unknown, decoding **must** fail.
```
jlizen, line 97: Currently we are enforcing with a `debug_assert!` that the encoder is properly writing the right number of fields for a given event's schema.

I guess we don't care to pay the runtime cost of the comparison? I probably would prefer to panic otherwise, it would cause corruption on the reader side unless we add extra overhead by length-prefixing events.

Probably ok for now, just thinking about the edge cases of this approach. Might be good to document the invariant more clearly on the encoder side (you MUST write values for every registered field)

dial9-trace-format/src/types.rs
```diff
@@ -0,0 +1,878 @@
...
+            timestamp_base_ns: 0,
+        }
+    }
+
+    /// Compute the timestamp delta, emitting a TimestampReset frame if needed
+    /// (delta overflow or backwards timestamp).
+    ///
+    /// The base advances to `ts_ns` after every event so that consecutive
+    /// inter-event deltas stay small — critical for gzip compressibility.
+    pub(crate) fn encode_timestamp_delta(&mut self, ts_ns: u64) -> io::Result<u32> {
```
jlizen, line 443: This could cause an overflow on the reader side, do we care? Maybe worth a debug assert as a signpost?

dial9-trace-format/src/types.rs
```diff
@@ -0,0 +1,878 @@
...
+            }
+            FieldType::F64 => {
+                let v = f64::from_le_bytes(data.get(..8)?.try_into().ok()?);
+                Some((FieldValue::F64(v), &data[8..]))
+            }
+            FieldType::Bool => Some((FieldValue::Bool(*data.first()? != 0), &data[1..])),
+            FieldType::String => {
+                let len = u32::from_le_bytes(data.get(..4)?.try_into().ok()?) as usize;
+                let bytes = data.get(4..4 + len)?;
+                let s = String::from_utf8(bytes.to_vec()).ok()?;
```
jlizen: Do we want to surface uf8 failures differently? Spec requires UTF-8 but still could be confusing to silently drop.

dial9-trace-format/src/encoder.rs
```diff
@@ -0,0 +1,411 @@
...
+        self.registry
+            .register(id, entry)
+            .expect("auto-register failed");
+        self.type_ids.push((rust_id, id));
+        Ok(id)
+    }
+
+    /// Register a schema for a marker type `T`. The encoder assigns the wire type_id.
+    /// Returns the assigned type_id. Panics if `T` is already registered.
+    pub fn register_schema_for<T: 'static>(
```
jlizen, line 86: Why panic here, but return a result on `SchemaRegistry::register()`? Probably should be consistent.

dial9-trace-format/src/encoder.rs
```diff
@@ -0,0 +1,411 @@
...
+        &mut self,
+        name: &str,
+        fields: Vec<crate::schema::FieldDef>,
+    ) -> io::Result<WireTypeId> {
+        self.register_schema_for_with_timestamp::<T>(name, false, fields)
+    }
+
+    /// Register a schema with explicit timestamp flag.
+    pub fn register_schema_for_with_timestamp<T: 'static>(
```
jlizen, line 98: Same feedback as `register_schema_for` (and also needs doc comment if keeping the panic)

dial9-trace-format/src/leb128.rs
```diff
@@ -0,0 +1,198 @@
...
+            return Some((result, pos));
+        }
+        if pos >= 10 {
+            return None;
+        } // u64 LEB128 is at most 10 bytes
+    }
+}
+
+#[inline]
+pub fn encode_signed(mut value: i64, w: &mut impl Write) -> io::Result<()> {
```
jlizen: The signed ones don't appear to be used anywhere besides tests?

dial9-trace-format/examples/format_comparison.rs
```diff
@@ -0,0 +1,801 @@
+//! Compare old (TOKIOTRC v16) vs new (TRC\0 v1) trace format.
```
jlizen, line 1: This is a pretty heavyweight example, do we plan to keep it around and up to date for newer schemas? Should it just be a separate binary?

PR #65, opened 2026-03-09
Remove SimpleBinaryWriter, use RotatingWriter everywhere

rcoh:
Replace all usages of `SimpleBinaryWriter` with `RotatingWriter::single_file()`. This reduces the API surface and ensures all writers go through the rename-on-seal path that the worker pipeline depends on.

Changes:
- Remove `SimpleBinaryWriter` struct and impl
- Add `RotatingWriter::single_file(path)` convenience constructor (no rotation, no eviction)
- Add `impl TraceWriter for Box<W>` delegation so callers no longer need `Box::new()`
- Update all examples, tests, benches, and docs
- Change `build()` / `build_and_start()` to accept `impl TraceWriter + 'static` instead of `Box<dyn TraceWriter>`

### DIFF
--- a/dial9-tokio-telemetry/tests/end_to_end.rs
+++ b/dial9-tokio-telemetry/tests/end_to_end.rs
@@ -1,7 +1,9 @@
 mod common;
 mod validation;
 
-use dial9_tokio_telemetry::telemetry::{RotatingWriter, TelemetryEvent, TraceReader, TracedRuntime, analyze_trace};
+use dial9_tokio_telemetry::telemetry::{
+    RotatingWriter, TelemetryEvent, TraceReader, TracedRuntime, analyze_trace,
+};
 use std::time::Duration;
 
 /// Run a known workload under TracedRuntime, read the trace back, and verify


dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -25,10 +25,10 @@ pub struct TelemetryRecorder {
 }
 
 impl TelemetryRecorder {
-    pub fn new(writer: Box<dyn TraceWriter>) -> Self {
+    pub fn new(writer: impl TraceWriter + 'static) -> Self {
```
jlizen, line 28: Good ergonomic improvement. Why not `impl Into<Box<T>>` to avoid double-boxing though if passed a boxed writer?

This could use a doc comment though. Probably we should add a lint for missing pub doc comments? Want a GH issue?

dial9-tokio-telemetry/tests/echo_server.rs
```diff
@@ -1,8 +1,6 @@
 mod validation;
 
-use dial9_tokio_telemetry::telemetry::{
-    SimpleBinaryWriter, TraceReader, TracedRuntime, analyze_trace,
-};
+use dial9_tokio_telemetry::telemetry::{RotatingWriter, TraceReader, TracedRuntime, analyze_trace};
```
jlizen, line 3: nit: this looks like it is missing a rustfmt? (Haven't checked CI step)

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
rcoh: the right way which I'll add eventually is to use a downcast trick to skip a double box

jlizen: Seems good, might be worth tossing in GitHub Issue given the new contributors?

PR #66, opened 2026-03-09
Add SegmentMetadata event (wire code 11)

rcoh:
New `TelemetryEvent::SegmentMetadata { entries: Vec<(String, String)> }` for embedding key-value metadata (service, host, boot_id, etc.) at the start of each trace segment. Makes trace files self-describing regardless of where they are stored.

Changes:
- New `SegmentMetadata` variant in `TelemetryEvent`
- Wire format: `code(u8) + num_entries(u16) + (key_len(u16) + key + val_len(u16) + val)*`
- Format read/write with roundtrip tests
- Updated analysis, `trace_to_fat_jsonl`, and JS trace viewer parser

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/format.rs
+++ b/dial9-tokio-telemetry/src/telemetry/format.rs
@@ -254,6 +254,12 @@ pub fn write_event(w: &mut impl Write, event: &TelemetryEvent) -> Result<()> {
         }
         TelemetryEvent::SegmentMetadata { entries } => {
             w.write_all(&[WIRE_SEGMENT_METADATA])?;
+            if entries.len() > u16::MAX as usize {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::InvalidInput,
+                    "SegmentMetadata: too many entries",
+                ));
+            }
             let num = entries.len() as u16;
             w.write_all(&num.to_le_bytes())?;
             for (key, val) in entries {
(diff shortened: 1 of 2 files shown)


dial9-tokio-telemetry/src/telemetry/format.rs
```diff
@@ -239,6 +231,19 @@ pub fn write_event(w: &mut impl Write, event: &TelemetryEvent) -> Result<()> {
             w.write_all(&woken_task_id.to_u32().to_le_bytes())?;
             w.write_all(&[*target_worker])?;
         }
+        TelemetryEvent::SegmentMetadata { entries } => {
+            w.write_all(&[WIRE_SEGMENT_METADATA])?;
+            let num = entries.len() as u16;
```
jlizen, line 263: This should have bounds checking and return `io::Error`, this metadata is user-supplied so while implausible it is more likely to overflow than other casts.

Probably could to do it in other places too, like `SpawnLocationDef`

dial9-tokio-telemetry/src/telemetry/format.rs
```diff
@@ -355,6 +347,34 @@ pub fn read_event(r: &mut impl Read) -> Result<Option<TelemetryEvent>> {
 
         return Ok(Some(TelemetryEvent::ThreadNameDef { tid, name }));
     }
+    if tag[0] == WIRE_SEGMENT_METADATA {
+        let mut num_bytes = [0u8; 2];
+        r.read_exact(&mut num_bytes)?;
+        let num = u16::from_le_bytes(num_bytes) as usize;
```
jlizen, line 393: nit: i know our trace input is generally trusted, but this is a slight DOS vector for unreasonably large input. The other readers already are capping it at 255, might be good to set an upper bounds below 65k

dial9-tokio-telemetry/trace_viewer/trace_parser.js
```diff
@@ -144,16 +139,27 @@
...
+                    const kLen = view.getUint16(off, true); off += 2;
+                    if (off + kLen > buffer.byteLength) { truncated = true; break; }
+                    off += kLen; // skip key
+                    if (off + 2 > buffer.byteLength) { truncated = true; break; }
+                    const vLen = view.getUint16(off, true); off += 2;
+                    if (off + vLen > buffer.byteLength) { truncated = true; break; }
+                    off += vLen; // skip value
+                }
+                if (truncated) break; // break outer loop
+                // TODO: Store metadata entries for display in the viewer (service name, host, etc.)
```
jlizen: GH issue for this would be good

PR #74, opened 2026-03-13
Update trace-format API with StringPool and serde support

rcoh:
cleanups in the trace-format API to smooth adoption into the main crate

### DIFF
--- a/dial9-trace-format/src/decoder.rs
+++ b/dial9-trace-format/src/decoder.rs
@@ -389,7 +389,8 @@ impl<'a> Decoder<'a> {
                     self.pos += 9;
                 }
                 _ => {
-                    // Use next_frame_ref for non-event frames (schema, pool, symbol table, reset)
+                    // Use next_frame_ref for non-event frames (schema, pool, symbol table)
+                    // `next_frame_ref` will update the decoder state as we read the frames (e.g. the pooled strings)
                     match self.next_frame_ref() {
                         Ok(Some(_)) => {}
                         Ok(None) => {


dial9-trace-format/src/codec.rs
```diff
@@ -163,7 +163,7 @@ pub fn encode_symbol_table(entries: &[SymbolEntry], w: &mut impl Write) -> io::R
     for e in entries {
         w.write_all(&e.base_addr.to_le_bytes())?;
         w.write_all(&e.size.to_le_bytes())?;
-        w.write_all(&e.symbol_id.to_le_bytes())?;
+        w.write_all(&e.symbol_id.0.to_le_bytes())?;
```
jv1i, line 166: very minor nit: more of a question (for reference for myself) than a suggestion, do we prefer `symbol_id.0` over `symbol_id.raw_id()` (the new method) everywhere, for a specific reason?

dial9-trace-format/src/decoder.rs
```diff
@@ -321,7 +365,28 @@ impl<'a> Decoder<'a> {
...
+                                pos: self.pos,
+                                message: "truncated timestamp reset".into(),
+                            });
+                        }
+                    };
+                    self.timestamp_base_ns = ts;
+                    self.pos += 9;
                 }
                 _ => {
                     // Use next_frame_ref for non-event frames (schema, pool, symbol table, reset)
```
jv1i: minor nit: timestamp reset now handled above, can update this comment:
```suggestion
                    // Use next_frame_ref for non-event frames (schema, pool, symbol table)
```

dial9-trace-format/src/codec.rs
rcoh: yeah no good reason.

PR #77, opened 2026-03-13
Support kernel frames in callchains when include_kernel=true

rcoh:
Add support for Kernel addresses when permissions exist

### DIFF
--- a/.github/actions/rust-build/action.yml
+++ b/.github/actions/rust-build/action.yml
@@ -21,6 +21,11 @@ runs:
                   rm -fv Cargo.lock
               fi
               cargo build
+        - name: Enable perf_event_open for tests
+          shell: bash
+          run: |
+              sudo sysctl kernel.perf_event_paranoid=1
+              cat /proc/sys/kernel/perf_event_paranoid
         - name: Run tests
           shell: bash
           env:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1443,6 +1443,7 @@ dependencies = [
  "criterion",
  "dial9-trace-format-derive",
  "flate2",
+ "serde",
  "serde_json",
  "tempfile",
 ]
--- a/dial9-tokio-telemetry/tests/common.rs
+++ b/dial9-tokio-telemetry/tests/common.rs
@@ -2,12 +2,6 @@ use dial9_tokio_telemetry::telemetry::events::TelemetryEvent;
 use dial9_tokio_telemetry::telemetry::writer::TraceWriter;
 use std::sync::{Arc, Mutex};
 
-/// Returns true when running in CI (GitHub Actions sets CI=true).
-#[allow(dead_code)]
-pub fn is_ci() -> bool {
-    std::env::var("CI").is_ok()
-}
-
 /// A [`TraceWriter`] that accumulates all events into a shared `Vec`.
 ///
 /// Construct with [`CapturingWriter::new`] and pass the returned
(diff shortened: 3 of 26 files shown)


perf-self-profile/src/lib.rs
```diff
@@ -41,5 +41,15 @@ mod ring_buffer;
 mod sampler;
 mod symbolize;
 
+/// Upper bound of userspace virtual addresses. Addresses at or above this limit
+/// are kernel addresses.
+///
+/// - x86_64: canonical address hole starts at bit 47
+/// - aarch64: TTBR0 (user) vs TTBR1 (kernel) selected by bit 63
+#[cfg(target_arch = "x86_64")]
```
jlizen, line 49: we'll get better errors if we have a fall-through `compile_error!`:
```
     #[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
     compile_error!("perf-self-profile: USER_ADDR_LIMIT not defined for this architecture");
```

perf-self-profile/src/sampler.rs
```diff
@@ -368,13 +369,17 @@ impl PerfSampler {
...
                         let callchain = s
                             .callchain()
                             .unwrap_or(&[])
                             .iter()
                             .copied()
-                            // TODO: imrpove this to allow capturing kernel traces when permissions are available
-                            .filter(|&a| a < USER_ADDR_LIMIT && a != 0)
+                            .filter(|&a| {
+                                a != 0
+                                    && a < PERF_CONTEXT_MAX
```
jlizen: This naming is a bit confusing. maybe: `PERF_CONTEXT_START_MARKER`? Since eg `PERF_CONTEXT_HV` / ++ are ascending beyond it.

perf-self-profile/src/symbolize.rs
```diff
@@ -93,6 +95,28 @@ pub fn resolve_symbol(addr: u64) -> SymbolInfo {
...
+                    name: Some(sym.name.to_string()),
+                    base_addr: sym.addr,
+                    code_info: None,
+                    offset: addr.saturating_sub(sym.addr),
+                };
+            }
+            return SymbolInfo {
+                name: Some(format!("[kernel] {:#x}", addr)),
+                base_addr: addr,
+                code_info: None,
```
jlizen, line 116: we should have code_info available from DWARF debug symbols, right? np to not impl it yet though

dial9-tokio-telemetry/examples/kernel_sched_events.rs
```diff
@@ -0,0 +1,137 @@
...
+async fn blocking_task(id: usize) {
+    for _ in 0..5 {
+        std::thread::sleep(Duration::from_millis(10));
+        tokio::task::yield_now().await;
+    }
+    eprintln!("Task {id} done");
+}
+
+fn main() {
+    let trace_path = "kernel_sched_trace.bin";
```
jlizen: might be nice to start a `example-traces/` folder to avoid cluttering

dial9-tokio-telemetry/examples/kernel_sched_events.rs
```diff
@@ -0,0 +1,137 @@
...
+            if printed < 3 {
+                printed += 1;
+                eprintln!("\n--- SchedEvent sample #{printed} (worker {worker_id}) ---");
+                for addr in callchain {
+                    let name = reader
+                        .callframe_symbols
+                        .get(addr)
+                        .cloned()
+                        .unwrap_or_else(|| format!("{:#x}", addr));
+                    if name.starts_with("[kernel]")
```
jlizen: this is fragile, false positives based on userspace symbols.

consider: `>= USER_ADDR_LIMIT`

perf-self-profile/src/sampler.rs
```diff
@@ -368,13 +369,17 @@ impl PerfSampler {
 
                 let sample = match parsed {
```
jlizen, line 371: The filtering logic is subtle enough that we should probably extract and unit test it

PR #78, opened 2026-03-13
Enable perf_event_open tests in CI

rcoh:
Set `kernel.perf_event_paranoid=1` in the CI workflow before running tests. This allows `perf_event_open` to work on GitHub Actions x64 runners, enabling 7 tests that were previously skipped when `CI=true`.

**Tests enabled:**
- `cpu_sample_clock_alignment` (2 assertion blocks)
- `sched_clock_alignment`
- `sched_events`
- `sched_triggers`: `captures_lock_acquisition_stack`, `captures_sleep_stack`
- `multithread`: `profiles_spawned_threads`

**Evidence this works:** [actions/runner-images#11689](https://github.com/actions/runner-images/issues/11689) confirms `sudo sysctl kernel.perf_event_paranoid=1` works on x64 GitHub Actions runners (Feb 2025).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1443,6 +1443,7 @@ dependencies = [
  "criterion",
  "dial9-trace-format-derive",
  "flate2",
+ "serde",
  "serde_json",
  "tempfile",
 ]
--- a/dial9-trace-format-derive/src/lib.rs
+++ b/dial9-trace-format-derive/src/lib.rs
@@ -41,7 +41,7 @@ fn derive_trace_event_impl(input: DeriveInput) -> proc_macro2::TokenStream {
         let field_name = field.ident.as_ref().unwrap();
         let ty = &field.ty;
 
-        // Skip the timestamp field in schema/encode/decode — it's in the event header
+        // Skip the timestamp field in schema/encode — it's in the event header
         if timestamp_field_name.as_ref() == Some(field_name) {
             continue;
         }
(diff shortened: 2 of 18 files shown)


.github/actions/rust-build/action.yml
```diff
@@ -21,6 +21,11 @@ runs:
                   rm -fv Cargo.lock
               fi
               cargo build
+        - name: Enable perf_event_open for tests
+          shell: bash
+          run: |
+              sudo sysctl kernel.perf_event_paranoid=1
+              cat /proc/sys/kernel/perf_event_paranoid
```
Fluzko, line 28: Is this `cat` really needed? seems that's something used to test if the previous command worked, but pretty sure we can remove it.

dial9-tokio-telemetry/tests/cpu_sample_clock_alignment.rs
```diff
@@ -185,7 +182,12 @@ fn cpu_sample_timestamps_align_with_wall_clock() {
             // Every sample that falls inside this burn window must come from
             // `expected_worker`.  A sample from a different worker would mean
             // the profiler mis-attributed the sample.
+            // Skip UNKNOWN_WORKER (255) samples — on CI the profiler can fire
+            // before the thread-to-worker mapping is visible.
             for &(t, w) in &in_window {
+                if w == UNKNOWN_WORKER {
+                    continue;
+                }
```
Fluzko, line 190: Can I get some context around why this can happen in the CI? just couriousity

Fluzko (approved):
Minor comment, LGTM

dial9-tokio-telemetry/tests/cpu_sample_clock_alignment.rs
rcoh: yeah, its not a CI issue. There's a general race condition where we can get perf events before we can accurately attribute a worker to the thread.

I think its solvable, especially with the `worker_index` API that I just added to Tokio though.

.github/actions/rust-build/action.yml
rcoh: yeah just helpful for debugging if it fails

PR #81, opened 2026-03-15
Add kernel tracepoint support to perf-self-profile

rcoh:
to support this, I needed to tweak the custom even codepath in dial9-trace-format (which is not used set so this is not an issue).

We won't be able to take advantage of this until we fully integrate the trace format into dial9-tokio-telemetry.

As part of this PR, I made some simplifications in the trace format:
1. All events (other than traceformat frames) have a timestamp. I think this is a good simplification and helps generally make stuff cleaner
2. The encode API is now a little easier to "hold" — you don't have to worry about pre-registering the schema.

There are only two direct code paths to write events:
1. `write` which accepts a `TraceEvent` implementing type
2. `write_event` which accepts a runtime-defined event where you must pass the `Schema`. You can only get a `Schema` by registering it, so its hard to misuse.

### DIFF
--- a/perf-self-profile/tests/common.rs
+++ b/perf-self-profile/tests/common.rs
@@ -1,3 +1,0 @@
-pub fn is_ci() -> bool {
-    std::env::var("CI").is_ok()
-}
--- a/perf-self-profile/tests/tracepoint_e2e.rs
+++ b/perf-self-profile/tests/tracepoint_e2e.rs
@@ -1,5 +1,3 @@
-mod common;
-
 use dial9_perf_self_profile::tracepoint::TracepointDef;
 use dial9_perf_self_profile::{PerfSampler, SamplerConfig};
 use dial9_trace_format::decoder::{DecodedFrame, Decoder};
@@ -11,11 +9,6 @@ use dial9_trace_format::types::FieldValue;
 /// dial9-trace-format, decode, and verify fields.
 #[test]
 fn tracepoint_sched_switch_e2e() {
-    if common::is_ci() {
-        eprintln!("skipping tracepoint e2e in CI");
-        return;
-    }
-
     // 1. Parse the kernel format file (includes tracepoint ID)
     let tp = match TracepointDef::from_event("sched", "sched_switch") {
         Ok(tp) => tp,


scripts/check-docsrs.sh
```diff
@@ -14,12 +14,16 @@ check_package() {
...
     echo "→ Checking docs.rs build for $pkg_name..."
     
-    # Because of workspace unification, we need to actually package the individual packages,
-    # then attempt to document the package itself to detect failure in the presence of
-    # of some bugs.
-    cargo package -p "$pkg_name" --allow-dirty
-    (cd "target/package/$pkg_name-$pkg_version" && \
-        cargo +nightly docs-rs --target "$TARGET")
+    # cargo package + docs-rs on the packaged crate catches workspace unification bugs.
+    # Falls back to building directly from the workspace when cargo package fails
```
jlizen, line 18: Thanks, I had just run into same in metrique

dial9-trace-format/src/encoder.rs
```diff
@@ -5,17 +5,59 @@ use crate::codec::{self, PoolEntry, SymbolEntry, WireTypeId};
...
+    /// Schema field definitions.
+    pub fn fields(&self) -> &[crate::schema::FieldDef] {
+        &self.0.fields
+    }
+}
+
+/// Key for schema lookup — either by name (manual registration) or by Rust
+/// `TypeId` (derive macro path).
+#[derive(Clone, PartialEq, Eq, Hash)]
+enum SchemaKey {
```
jlizen, line 59: Good refactor

dial9-trace-format/src/encoder.rs
```diff
@@ -5,17 +5,59 @@ use crate::codec::{self, PoolEntry, SymbolEntry, WireTypeId};
...
 use crate::types::{EncodeState, EventEncoder};
 use std::any::TypeId;
+use std::collections::HashMap;
 use std::io::{self, Write};
+use std::sync::Arc;
+
+/// A schema handle returned by [`Encoder::register_schema`] or created via
+/// [`Schema::new`].
+///
+/// Carries the full schema definition (name + fields) so it can auto-register
```
jlizen, line 15: Much nicer external API with this

perf-self-profile/src/tracepoint.rs
```diff
@@ -0,0 +1,820 @@
...
+//! use dial9_perf_self_profile::tracepoint::TracepointDef;
+//!
+//! let tp = TracepointDef::from_event("sched", "sched_switch").unwrap();
+//! println!("tracepoint {} has id {}", tp.name, tp.id);
+//! for field in &tp.fields {
+//!     println!("  {}: {} (offset={}, size={})", field.name, field.field_type, field.offset, field.size);
+//! }
+//! ```
+
+use std::io::{self, Write};
```
jlizen, line 19: imports need cleanup, i think

perf-self-profile/src/tracepoint.rs
```diff
@@ -0,0 +1,820 @@
...
+                        end,
+                        raw.len()
+                    ),
+                ));
+            }
+
+            if field.is_dynamic {
+                // __data_loc: 4 bytes at field.offset encode (offset << 16) | length
+                let loc =
+                    u32::from_ne_bytes(raw[field.offset..field.offset + 4].try_into().unwrap());
```
jlizen, line 163: not portable to big endian, we should probably highlight for later fixing even if we aren't adding config guards now

dial9-trace-format/src/encoder.rs
```diff
@@ -5,17 +5,59 @@ use crate::codec::{self, PoolEntry, SymbolEntry, WireTypeId};
...
+
+impl Schema {
+    /// Create a schema handle without an encoder.
+    ///
+    /// The schema will be lazily registered the first time it is passed to
+    /// [`Encoder::write_event`].
+    pub fn new(name: &str, fields: Vec<crate::schema::FieldDef>) -> Self {
+        Self(Arc::new(SchemaEntry {
+            name: name.to_string(),
+            has_timestamp: true,
```
jlizen: We should update the `SchemaEntry` struct in `schema.rs` tests as well

perf-self-profile/src/tracepoint.rs
```diff
@@ -0,0 +1,820 @@
...
+    pub fn encode_raw<W: Write>(
+        &self,
+        encoder: &mut Encoder<W>,
+        schema: &Schema,
+        timestamp_ns: u64,
+        raw: &[u8],
+    ) -> io::Result<()> {
+        let extracted = self.extract_fields(raw)?;
+        let field_values = self.to_trace_format_values(&extracted);
+        let mut values = Vec::with_capacity(1 + field_values.len());
```
jlizen, line 239: Fine for now, but some alternative optimizations available here:
- smallvec
- reuse the vec in some `TracepointEncoder` struct
 
We could also decompose into lower-level encoder methods that write directly to the encoder, but that seems like a bad external api

dial9-trace-format/src/encoder.rs
```diff
@@ -67,147 +109,128 @@ impl<W: Write> Encoder<W> {
...
+            cached
         } else {
-            None
+            let entry = T::schema_entry();
+            let schema = Schema::new(&entry.name, entry.fields);
+            let id = self.ensure_registered(&schema)?;
+            self.schema_ids.insert(key, id);
+            id
         };
+        let ts_ns = event.timestamp().expect("TraceEvent must have a timestamp");
```
jlizen: This would cause a runtime panic on a bad `TraceEvent` implementation that doesn't set a timestamp.

It would be nicer to instead enforce it at the type level by changing `timestamp()` to return `u64` directly

dial9-trace-format/src/encoder.rs
```diff
@@ -67,147 +109,128 @@ impl<W: Write> Encoder<W> {
...
-        let rust_id = TypeId::of::<T>();
-        for &(tid, wire_id) in &self.type_ids {
-            if tid == rust_id {
+    /// Ensure a schema is registered with this encoder. Returns the wire type
+    /// ID for this encoder's output stream.
+    ///
+    /// Idempotent if the schema matches. Errors if a different schema was
+    /// already registered under the same name.
+    fn ensure_registered(&mut self, schema: &Schema) -> io::Result<WireTypeId> {
+        let key = SchemaKey::Name(schema.name().to_string());
```
jlizen: Nit: this is a string allocation in the `write_event` path. Consider Arcing

perf-self-profile/src/tracepoint.rs
```diff
@@ -0,0 +1,820 @@
...
+                && let Some(field) = parse_field_line(rest)
+            {
+                // Skip common fields (kernel bookkeeping).
+                if !field.name.starts_with("common_") {
+                    fields.push(field);
+                }
+            }
+        }
+
+        if name.is_empty() {
```
jlizen, line 123: nit: would be good to have similar for tracepoint id

perf-self-profile/src/sampler.rs
```diff
@@ -291,16 +302,28 @@ impl PerfSampler {
...
+        );
 
         attr.sample_type = PERF_SAMPLE_IP as u64
             | PERF_SAMPLE_CALLCHAIN as u64
             | PERF_SAMPLE_TID as u64
             | PERF_SAMPLE_TIME as u64
             | PERF_SAMPLE_CPU as u64
-            | PERF_SAMPLE_PERIOD as u64;
+            | PERF_SAMPLE_PERIOD as u64
+            | if matches!(config.event_source, EventSource::Tracepoint(_)) {
```
jlizen, line 327: nit: mention why only tracepoints are getting data?

dial9-trace-format/src/encoder.rs
rcoh: Good call. Ideally we will set it up so you can just use a reference as the hashmap key

dial9-trace-format/src/encoder.rs
rcoh: Yeah good call. We should have the macro make timestamp mandatory

PR #85, opened 2026-03-16
feat: add enabled() builder method for conditional telemetry install

### LINKED ISSUE #51: Add a constructor so that you can conditionally install dial9 on a runtime
Make it possible to have simple code, that e.g.:

```rust
let enabled = std::env::var("ENABLE_DIAL9");
TraceRuntime::builder().conditionally_install(enabled).<....>
```

if you pass `false` into conditionally install, everything should be a total no-op -- it shouldn't modify the tokio runtime at all, start the flush thread, etc.

jlizen:
# Add `enabled()` builder method for conditional telemetry install

Closes #51

### Summary

Adds `TracedRuntimeBuilder::enabled(bool)` (defaults to `true`). When set to `false`, `build()` produces a plain Tokio runtime with no telemetry installed and a dummy `TelemetryGuard`.

```rust
let enabled = std::env::var("ENABLE_DIAL9").is_ok();
let (runtime, guard) = TracedRuntime::builder()
    .enabled(enabled)
    .build_and_start(builder, writer)?;
```

### Testing

Added a unit test that builds with `enabled(false)` and verifies the runtime works normally with no threads spawned.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,9 @@ exclude = [
 
 [workspace.package]
 version = "0.1.1"
+edition = "2024"
+license = "Apache-2.0"
+repository = "https://github.com/dial9-rs/dial9-tokio-telemetry"
 
 [workspace.dependencies]
 libc = "0.2"
--- a/dial9-tokio-telemetry/Cargo.toml
+++ b/dial9-tokio-telemetry/Cargo.toml
@@ -1,10 +1,14 @@
 [package]
 name = "dial9-tokio-telemetry"
-edition = "2024"
-version = "0.1.1"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
 publish = true
+readme = "README.md"
 description = "Low-overhead runtime telemetry for Tokio with poll timing, wake events, and CPU profiling"
-license = "Apache-2.0"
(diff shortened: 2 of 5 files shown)


dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -344,6 +345,24 @@ pub struct TracedRuntimeBuilder {
 }
 
 impl TracedRuntimeBuilder {
+    /// Set to `false` to build a plain runtime with no telemetry
```
rcoh, line 348: I wonder if we should give this a stronger name and clarify how it differs from the start/stop handle

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -586,6 +631,31 @@ mod tests {
         let _recorder = TelemetryRecorder::new(NullWriter);
     }
 
+    #[test]
+    fn build_disabled_produces_working_runtime_with_noop_guard() {
+        let builder = tokio::runtime::Builder::new_multi_thread();
+        let (runtime, guard) = TracedRuntime::builder()
```
rcoh, line 641: Can you test handle.spawn when disabled?

Issue #24, opened 2026-03-01
move symbolizing to a background process 


rcoh:
At least as an option, trace files should be stored with addresses then when they are flushed, the background worker should be triggered that turns the addresses into symbols.

This requires a few changes:
- new record type that records the current state of /proc/maps
- design how symbols will be stored (e.g. as a footer, as an additional event type etc.)
- Update parsers to handle this. 

this is somewhat blocked on fully migrating to the new format though

PR #87, opened 2026-03-16
feat: Add ProcMaps frame and offline symbolizer for background symbolization

jlizen:
## Add offline symbolizer with schema-based proc maps and symbol table

Part of https://github.com/dial9-rs/dial9-tokio-telemetry/issues/24

### Summary

Adds the building blocks for moving symbolization off the flush thread and into a background process. Today `inline_callframe_symbols` resolves every callframe address synchronously via blazesym (DWARF parsing, file I/O), blocking trace flushing.

This PR adds:

1. `ProcMapsEntry` and `SymbolTableEntry` event types defined with `derive(TraceEvent)` in `perf-self-profile`, using the existing schema/event system rather than dedicated wire frame types. Any decoder that understands schemas automatically handles them without custom parsing code.
2. An offline `symbolize_trace()` function in `perf-self-profile` that reads a trace with `ProcMapsEntry` events and `StackFrames` fields, resolves addresses via blazesym (including inlined functions), and writes only the new `SymbolTableEntry` frames to the output. The caller controls how the original trace and symbol data are combined (e.g. by opening the file in append mode).
3. `resolve_symbols_with_maps()` in `perf-self-profile`, which returns all symbols at an address including inlined callees (the existing `resolve_symbol_with_maps` delegates to it for backwards compatibility). Both userspace and kernel paths handle inlined functions.

Does not change the telemetry crate or `TOKIOTRC` format. Wiring into the background worker pipeline and format migration are follow-up work.

#### Design: schema-based events over dedicated frame types

ProcMaps and SymbolTable data are represented as regular `derive(TraceEvent)` structs rather than custom wire frame types. The core wire format now only has Schema, Event, StringPool, and TimestampReset.

The main motivation is backwards compatibility. Dedicated frame types require every decoder to add custom parsing code for each new type, and there is currently no way to evolve them additively. Schema-based events only require decoders to understand the generic schema/event mechanism. New data shapes (like proc maps or symbol tables) can be introduced without any decoder changes. Consumers that care about the data match on the schema name; those that don't simply see it as another event and skip it.

These events don't actually carry timestamps, so we instead trigger a refresh frame to grab the current time (offline).

### Testing

- Round-trip tests for `ProcMapsEntry` and `SymbolTableEntry` schema-based events
- Symbolizer tests: empty trace, missing proc maps, missing stack frames, unresolvable addresses, full end-to-end flow with real addresses
- `parse_proc_maps_filters_correctly` test for the maps parser
- JS decoder, fuzz target, and integration tests updated

### DIFF
--- a/perf-self-profile/src/offline_symbolize.rs
+++ b/perf-self-profile/src/offline_symbolize.rs
@@ -63,11 +63,13 @@ pub fn encode_proc_maps(
     Ok(())
 }
 
-/// Symbolize a trace: read the input, resolve addresses from `ProcMapsEntry`
-/// events and `StackFrames` data, and write the original trace plus
-/// `SymbolTableEntry` events to the output.
+/// Symbolize a trace: read `input` for `ProcMapsEntry` events and `StackFrames`
+/// fields, resolve addresses via blazesym, and write only the new
+/// `SymbolTableEntry` frames (with a `StringPool`) to `output`.
 ///
-/// The input trace is copied verbatim to the output, with symbol data appended.
+/// The caller is responsible for the original trace data. In the typical
+/// file-based workflow, open the trace file in append mode and pass it as
+/// `output` so symbols are appended in place without copying.
 pub fn symbolize_trace(input: &[u8], output: &mut impl Write) -> io::Result<()> {
     let proc_maps_name = ProcMapsEntry::event_name();
 
@@ -111,7 +113,6 @@ pub fn symbolize_trace(input: &[u8], output: &mut impl Write) -> io::Result<()> 
     }
 
     if addresses.is_empty() || maps.is_empty() {
-        output.write_all(input)?;
         return Ok(());
     }
 
@@ -145,9 +146,7 @@ pub fn symbolize_trace(input: &[u8], output: &mut impl Write) -> io::Result<()> 
         }
     }
 
-    // Write: original trace verbatim, then append symbol data.
-    output.write_all(input)?;
-
+    // Append symbol data.
     if !symbol_events.is_empty() {
         codec::encode_string_pool(&pool_entries, output)?;
 
@@ -201,11 +200,11 @@ mod tests {
         let input = Encoder::new().finish();
         let mut output = Vec::new();
         symbolize_trace(&input, &mut output).unwrap();
-        assert_eq!(input, output);
+        assert!(output.is_empty());
     }
 
     #[test]
-    fn symbolize_no_proc_maps_copies_verbatim() {
+    fn symbolize_no_proc_maps_writes_nothing() {
         let mut buf = Encoder::new().finish();
         let schema = SchemaEntry {
             name: "Ev".into(),
@@ -227,11 +226,11 @@ mod tests {
 
         let mut output = Vec::new();
         symbolize_trace(&buf, &mut output).unwrap();
-        assert_eq!(buf, output);
+        assert!(output.is_empty());
     }
 
     #[test]
-    fn symbolize_no_stack_frames_copies_verbatim() {
+    fn symbolize_no_stack_frames_writes_nothing() {
         let input = make_encoder_with_proc_maps(&[MapsEntry {
             start: 0x1000,
             end: 0x2000,
@@ -240,7 +239,7 @@ mod tests {
         }]);
         let mut output = Vec::new();
         symbolize_trace(&input, &mut output).unwrap();
-        assert_eq!(input, output);
+        assert!(output.is_empty());
     }
 
     #[test]
@@ -277,8 +276,10 @@ mod tests {
 
         let mut output = Vec::new();
         symbolize_trace(&buf, &mut output).unwrap();
-        assert!(output.len() >= buf.len());
-        let mut dec = Decoder::new(&output).unwrap();
+        // Unresolvable addresses produce no symbol events
+        let mut combined = buf.clone();
+        combined.extend_from_slice(&output);
+        let mut dec = Decoder::new(&combined).unwrap();
         let frames = dec.decode_all();
         assert!(frames.len() >= 4);
     }
@@ -310,13 +311,12 @@ mod tests {
         let mut output = Vec::new();
         symbolize_trace(&buf, &mut output).unwrap();
 
-        assert!(
-            output.len() > buf.len(),
-            "expected symbol data to be appended"
-        );
+        assert!(!output.is_empty(), "expected symbol data to be written");
 
         let symbol_table_name = SymbolTableEntry::event_name();
-        let mut dec = Decoder::new(&output).unwrap();
+        let mut combined = buf.clone();
+        combined.extend_from_slice(&output);
+        let mut dec = Decoder::new(&combined).unwrap();
         let frames = dec.decode_all();
         let has_string_pool = frames
             .iter()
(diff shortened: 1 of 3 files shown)


dial9-trace-format/SPEC.md
```diff
@@ -147,6 +148,30 @@ Total: **9 bytes**.
...
+
+Only executable (`r-xp`) mappings with file-backed paths are included.
+
+| Field | Type | Description |
+|-------|------|-------------|
+| tag | u8 | `0x06` |
+| count | u32 | Number of entries |
+| entries | [ProcMapsEntry; count] | Mapping entries (see below) |
+
+Each **ProcMapsEntry**:
```
rcoh: Not as much a question for this PR but we should have a solid mental model for when to put types into the format directly.

There is considerable cost to doing this — the backwards compat story is much harder and all decoders need custom code to support new frame types.

For example we could support this by having each row be a "regular event"

It's not totally clear that we shouldn't do the same thing with symbolframe as well.

jlizen: Yeah i considered just making it a regular `Event` like `ProcMapsEntry { start_addr: u64, end_addr: u64, file_offset: u64, path: String }`.

It didn't feel quite right as an event because:
a) it had no timestamp and isn't really a trace event
b) a dedicated frame makes it simpler for decoders to find and extract rather than relying on convention
c) it matches the `SymbolTable` approach

Glad to change it though. I agree that we should have some guidance on this decisionmaking.

dial9-trace-format/src/symbolize.rs
```diff
@@ -0,0 +1,255 @@
...
+    // Resolve symbols.
+    let symbolizer = Symbolizer::new();
+    let mut pool_entries: Vec<crate::codec::PoolEntry> = Vec::new();
+    let mut symbol_entries: Vec<SymbolEntry> = Vec::new();
+    let mut string_to_id: HashMap<String, u32> = HashMap::new();
+    let mut seen_base_addrs: HashSet<u64> = HashSet::new();
+    let mut next_pool_id: u32 = 0;
+
+    for &addr in &addresses {
+        let info = dial9_perf_self_profile::resolve_symbol_with_maps(addr, &symbolizer, &maps);
```
rcoh: We need to refactor this to be able to return multiple symbols for inlining

dial9-trace-format/SPEC.md
rcoh: Yeah. Backwards compat is the main issue I see. Right now there is no way to easily modify the core events even in a way that is purely additive.

Whereas if we emit schema frames for them, only the names of the fields and their types are load bearing.

We could also explore enhancing the format to have better compatibility for internal types but I feel like avoiding internal types is probably cleaner.

One option that could be nice is having the crates re-export their trace types (e.g. moving the symbolization types into perf-self-profile)

Future work could also emit JS schemas for the shapes or some other very clean and straightforward way for code to pull in the schemas and parse the trace.

I guess as a result of this conversation I've sort of convinced myself we should move these fields out of the format itself.

dial9-trace-format/fuzz/fuzz_targets/fuzz_structured_round_trip.rs
```diff
@@ -120,12 +118,10 @@ struct String8 {
     len: u8,
 }
 
-#[derive(Arbitrary, Debug)]
-struct FuzzSymbol {
-    base_addr: u64,
-    size: u32,
-    symbol_id: u32,
-}
+struct S0;
```
rcoh: I think this is a result of a bad merge, these aren't used anymore

perf-self-profile/src/symbolize.rs
```diff
@@ -83,78 +101,116 @@ const EMPTY: SymbolInfo = SymbolInfo {
...
+    addr: u64,
+    symbolizer: &Symbolizer,
+    mappings: &[MapsEntry],
+) -> Vec<SymbolInfo> {
+    // Kernel addresses are >= USER_ADDR_LIMIT
+    if addr >= USER_ADDR_LIMIT {
+        let src = source::Source::Kernel(source::Kernel::default());
+        if let Ok(results) = symbolizer.symbolize(&src, Input::AbsAddr(&[addr]))
+            && !results.is_empty()
+            && let Some(sym) = results[0].as_sym()
```
rcoh, line 158: while you are here, we should read all the results, not just `results[0]`

perf-self-profile/src/offline_symbolize.rs
```diff
@@ -0,0 +1,452 @@
...
+                name: "size".into(),
+                field_type: FieldType::Varint,
+            },
+            FieldDef {
+                name: "symbol_name".into(),
+                field_type: FieldType::PooledString,
+            },
+        ],
+    }
+}
```
rcoh: why did you do it this way instead of `derive(TraceEvent)`? Because of no timestamp? It seems like sticking an arbitrary timestamp on this is pretty harmless IMO then you can derive.

The other big change we should tackle here (which is not trivial) is dealing with multiple symbols per address

perf-self-profile/src/offline_symbolize.rs
```diff
@@ -0,0 +1,452 @@
...
+    }
+    Ok(())
+}
+
+/// Symbolize a trace: read the input, resolve addresses from `ProcMapsEntry`
+/// events and `StackFrames` data, and write the original trace plus
+/// `SymbolTableEntry` events to the output.
+///
+/// The input trace is copied verbatim to the output, with symbol data appended.
+pub fn symbolize_trace(input: &[u8], output: &mut impl Write) -> io::Result<()> {
```
rcoh, line 73: if we are going to rewrite the whole trace, I wonder if we should put the symbols at the beginning? Otherwise, it would be nice if we could literally just append to the file (instead of doing a rewrite)

perf-self-profile/src/offline_symbolize.rs
jlizen: Yeah, just to avoid the dummy timestamp. But I can use the higher level API if we don't mind it.

Actually, i think also not too bad just to tweak the derive macro to allow no timestamp.

jlizen: ACK re: multiple symbols

jlizen:
@rcoh this is ready for more eyes, i swapped timestamp to just use `0`. Clean rebase against #91, it looks like.

jlizen:
@rcoh i went switched to writing a reset frame. 

AIUI, writing 0 just means, use the current state of the wall-clock (since the 0 is handled as a diff), which is less useful than a reset.

perf-self-profile/src/offline_symbolize.rs
```diff
@@ -0,0 +1,405 @@
...
+
+            symbol_events.push((info.base_addr, InternedString::from_raw(pool_id)));
+        }
+    }
+
+    // Append symbol data.
+    if !symbol_events.is_empty() {
+        codec::encode_string_pool(&pool_entries, output)?;
+
+        let type_id = WireTypeId(0xFFFE);
```
rcoh, line 153: WireTypeId needs to be assigned by the encoder — we shouldn't be using the codec directly.

I'm guessing it is this way because that's the only way to append to a file right now.

Guess we should enhance the encoder.

PR #95, opened 2026-03-18
feat: Wire background symbolization into the flush/worker pipeline

jlizen:
closes: #24 

### Summary

Today, when `cpu-profiling` is enabled with `with_inline_callframe_symbols(true)`, the flush thread does synchronous symbolization: for every CPU sample, it calls `resolve_symbol()` (which invokes blazesym) inline, blocking the flush thread. This is expensive (blazesym parses DWARF/ELF on each new address) and adds unpredictable latency to the flush path, which runs every 250ms on a dedicated thread shared with queue sampling.

PR #87 landed the building blocks for offline symbolization:
- `encode_proc_maps()`: writes `/proc/self/maps` entries as `ProcMapsEntry` schema events into the trace
- `symbolize_trace()`: reads a trace with raw `StackFrames` addresses + `ProcMapsEntry` events, resolves symbols via blazesym, and appends `SymbolTableEntry` events

This PR wires those building blocks into the actual runtime:

1. **Flush path writes raw addresses.** When background symbolization is active, `resolve_cpu_event_symbols` emits only the raw `CpuSample` event (with unresolved `StackFrames` addresses), skipping the per-address `resolve_symbol()` calls entirely.

2. **Background worker symbolizes sealed segments.** A new `SymbolizeProcessor` pipeline stage runs in the existing background worker before gzip compression and S3 upload. It captures `/proc/self/maps` from the current process, appends `ProcMapsEntry` events to the segment bytes, then calls `symbolize_trace()` to resolve addresses and append `SymbolTableEntry` events. This moves all blazesym work off the flush thread and onto the worker thread, where latency is not critical.

3. **Automatic, zero-config activation.** There is no new public API knob. Background symbolization activates automatically when `cpu-profiling` is enabled, a `CpuProfilingConfig` is configured, and the writer provides a `trace_path()` (i.e. writes to disk via `RotatingWriter`). Setting `with_inline_callframe_symbols(true)` forces synchronous inline symbolization on the flush thread and disables background symbolization. We keep this around because our tests use it and refactoring would be pretty annoying. Perhaps some other custom writer will want it too.

4. **Works without S3.** When S3 is not configured, the builder auto-spawns the worker thread and the pipeline uses `WriteBackProcessor` to write the symbolized bytes back to disk. When S3 is configured, symbolized data is uploaded to S3 but local `.bin` files are not updated (they are temporary staging). The pipeline composition is: `SymbolizeProcessor` → `GzipCompressor` → `S3PipelineUploader` (with S3), or `SymbolizeProcessor` → `WriteBackProcessor` (without S3).

#### Breaking changes

- `BackgroundTaskConfig.s3` changed from `S3Config` to `Option<S3Config>`. Existing `.s3(config)` builder calls still work (bon auto-wraps in `Some`), but `config.s3()` now returns `Option<&S3Config>`.

#### Design rationale

- **Proc maps captured in the worker, not the flush thread.** The worker runs in the same process and processes segments within seconds of sealing. For typical Tokio services, memory mappings are fixed at startup (no dlopen), so capturing maps at worker processing time is equivalent to capturing at flush time. This avoids any changes to `TraceWriter`, `RawEvent`, or the flush path beyond the symbolization skip.
- **Symbolize before compress**: the symbolizer needs to decode the trace, so it must run on uncompressed data. Pipeline order: symbolize, gzip, S3 upload.
- **Append-only output**: `symbolize_trace()` writes only new `SymbolTableEntry` + `StringPool` frames. The worker appends symbol data to the in-memory byte buffer, then passes it downstream. No temp files needed.
- **No changes to `TraceWriter` or `RawEvent`**: the `SymbolizeProcessor` operates on the sealed segment bytes in the worker pipeline, completely decoupled from the flush path. The only flush-path change is the conditional skip in `CpuFlushState`.
- **Unified pipeline**: symbolization is just another `SegmentProcessor` stage in the existing worker pipeline, not a separate system. The worker thread is shared between symbolization and S3 upload.

### Testing

- `SymbolizeProcessor` unit tests: `symbolize_processor_resolves_real_address` verifies that a trace with a real function address from the test binary gets `SymbolTableEntry` events appended. `symbolize_processor_no_stack_frames_is_noop` verifies that traces without `StackFrames` fields pass through with only `ProcMapsEntry` additions.
- Integration test `background_worker_symbolizes_sealed_segments`: builds a `TracedRuntime` with cpu-profiling and `RotatingWriter`, burns CPU to generate samples, performs graceful shutdown, then reads sealed `.bin` segments and verifies they contain `SymbolTableEntry` events (not just the schema).

### DIFF
--- a/dial9-tokio-telemetry/src/background_task/mod.rs
+++ b/dial9-tokio-telemetry/src/background_task/mod.rs
@@ -194,15 +194,16 @@ async fn build_pipeline(config: &mut BackgroundTaskConfig) -> Vec<Box<dyn Segmen
         pipeline.push(Box::new(SymbolizeProcessor));
     }
 
+    let mut has_s3 = false;
     #[cfg(feature = "worker-s3")]
     if let Some(s3_config) = config.s3.take() {
         let s3_uploader = S3PipelineUploader::new(s3_config, config.client.take()).await;
         pipeline.push(Box::new(GzipCompressor));
         pipeline.push(Box::new(s3_uploader));
+        has_s3 = true;
     }
 
-    // Without S3: gzip and write back to disk.
-    if !pipeline.iter().any(|p| p.name() == "S3Upload") {
+    if !has_s3 {
         pipeline.push(Box::new(GzipCompressor));
         pipeline.push(Box::new(WriteBackProcessor));
     }


dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -232,6 +262,122 @@ pub(crate) fn run_background_task(
...
+        mut data: SegmentData,
+    ) -> Pin<Box<dyn Future<Output = Result<SegmentData, ProcessError>> + Send + '_>> {
+        Box::pin(async move {
+            let input = std::mem::take(&mut data.bytes);
+            let result = tokio::task::spawn_blocking(move || {
+                // Capture current proc maps and append them to the trace data
+                // so symbolize_trace can resolve addresses.
+                let maps = dial9_perf_self_profile::read_proc_maps();
+                let mut enriched = input;
+                if let Err(e) = dial9_perf_self_profile::offline_symbolize::encode_proc_maps(
```
rcoh: There isn't really value in encoding proc maps at this point.

There is already a dump, we are in the process.

The place where we would actually get value from this is encoding them when we sealed the file.

That would allow fully off process symbolization eventually.

This is not harmful right now but I don't think we gain anything from encoding then reparsing the proc maps currently

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -232,6 +262,122 @@ pub(crate) fn run_background_task(
...
+        })
+    }
+}
+
+// ---------------------------------------------------------------------------
+// WriteBackProcessor — writes processed bytes back to the segment file
+// ---------------------------------------------------------------------------
+
+#[cfg(feature = "cpu-profiling")]
+struct WriteBackProcessor;
```
rcoh: Might was well gzip before we write it back?

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -390,9 +390,20 @@ impl TracedRuntimeBuilder {
...
+    /// blazesym and emits `CallframeDef` events inline. This disables
+    /// background symbolization even when a `trace_path()` is available.
+    ///
+    /// In most cases you should leave this off (the default). Background
+    /// symbolization is faster and produces richer output
+    /// (`SymbolTableEntry` + `ProcMapsEntry` events). Use this only if
+    /// you need symbols in a custom `TraceWriter` that does not write to
+    /// disk.
     #[cfg(feature = "cpu-profiling")]
     pub fn with_inline_callframe_symbols(mut self, enabled: bool) -> Self {
```
rcoh: We should probably deprecate this

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -390,9 +390,20 @@ impl TracedRuntimeBuilder {
...
-        self.inline_callframe_symbols = enabled;
+        self.callframe_symbols = enabled;
```
rcoh: Rename this `symbolize_on_flush`?

I am fine just removing this. We will release 0.2 anyway

dial9-tokio-telemetry/tests/background_symbolization.rs
```diff
@@ -0,0 +1,111 @@
...
+#[cfg(feature = "cpu-profiling")]
+#[test]
+fn background_worker_symbolizes_sealed_segments() {
+    use dial9_tokio_telemetry::telemetry::{CpuProfilingConfig, RotatingWriter, TracedRuntime};
+    use dial9_trace_format::TraceEvent as _;
+    use dial9_trace_format::decoder::{DecodedFrame, Decoder};
+    use std::time::Duration;
+
+    let trace_dir = tempfile::tempdir().unwrap();
+    let trace_path = trace_dir.path().join("trace.bin");
```
rcoh, line 33: Worth validating in the s3 integration test as well since there is some interaction of those features

examples/metrics-service/src/main.rs
```diff
@@ -113,7 +113,6 @@ fn main() -> std::io::Result<()> {
     {
         traced_builder = traced_builder
             .with_cpu_profiling(CpuProfilingConfig::default())
-            .with_inline_callframe_symbols(true)
             .with_sched_events(SchedEventConfig {
                 include_kernel: true,
```
rcoh, line 119: does the `record-demo-trace.sh` script still work?

dial9-tokio-telemetry/tests/background_symbolization.rs
```diff
@@ -0,0 +1,111 @@
...
+            break;
+        }
+    }
+
+    assert!(
+        found_symbol_event,
+        "expected at least one SymbolTableEntry event in sealed segments, \
+         checked {} .bin files",
+        bin_files.len()
+    );
```
rcoh: any reason not to fully parse these and actually look for the symbolized stack traces you are expecting?

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -518,7 +539,27 @@ impl TracedRuntimeBuilder {
 
         #[allow(unused_mut)]
         let mut worker = None;
-        if let Some(config) = self.worker_config {
+        // Determine the worker config: either user-provided (with_s3_uploader)
```
rcoh: how did this work before we added `trace_path`?

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -182,15 +188,39 @@ pub(crate) trait SegmentProcessor: Send {
...
 async fn build_pipeline(_config: &mut BackgroundTaskConfig) -> Vec<Box<dyn SegmentProcessor>> {
-    Vec::new()
+    #[allow(unused_mut)]
+    let mut pipeline: Vec<Box<dyn SegmentProcessor>> = Vec::new();
+    #[cfg(feature = "cpu-profiling")]
+    if _config.symbolize {
+        pipeline.push(Box::new(SymbolizeProcessor));
+        pipeline.push(Box::new(WriteBackProcessor));
+    }
+    pipeline
```
rcoh, line 211: can we unify these into one method now? Also, we might as well gzip when we write back? that is what the design doc says I think?

examples/metrics-service/src/main.rs
jlizen: will check

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
jlizen: The thing is, it's kind of a pain to test without some functionality along these lines. We could move it behind test-util if we don't expect further encoders to want to have self-contained inline symbols?

Or, would you rather refactor our tests to not rely on this, and just rip out the inline logic? I guess it might be the more sustainable approach?

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
jlizen: it was part of `with_s3_uploader`, I refactored to unify since anyway we are releasing a breaking release

dial9-tokio-telemetry/tests/background_symbolization.rs
jlizen: updated

dial9-tokio-telemetry/Cargo.toml
```diff
@@ -39,7 +39,7 @@ metrique-writer = "0.1"
 metrique = { version = "0.1.20", features = ["local-format"] }
 
 [features]
-cpu-profiling = ["dep:dial9-perf-self-profile"]
+cpu-profiling = ["dep:dial9-perf-self-profile", "dep:flate2"]
```
rcoh, line 42: What does cpu profiling have to do with flate2? Presumably I'll find out

dial9-tokio-telemetry/README.md
rcoh, line 127: Oh I see...the writer is built on the trace path. So then we end up duplicating it. I see why we put it on the trait.

dial9-tokio-telemetry/design/cpu-sampling-and-worker-id-design.md
```diff
@@ -155,29 +159,24 @@ All non-worker CPU samples carry `worker_id = 255`, which makes them indistingui
...
+## 6. Wire Format
+
+The trace format has migrated from fixed wire codes to `dial9-trace-format`, a self-describing schema-based binary format. Events are identified by schema name rather than numeric codes. The format uses `StringPool` frames for string interning and schema frames to define event layouts.
+
+Key event types (as schema names):
+- `PollStart`, `PollEnd`, `WorkerPark`, `WorkerUnpark`
+- `QueueSample`, `TaskSpawn`, `WakeEvent`
+- `CpuSample` (carries worker_id, tid, source, and stack frames)
+- `ThreadNameDef` (maps OS tid to thread name)
+- `SymbolTableEntry` (resolved symbol: base_addr, size, symbol_name)
```
rcoh: Not that we need to handle it right now but does this handle >1 symbol per address?

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -181,16 +185,33 @@ pub(crate) trait SegmentProcessor: Send {
...
-async fn build_pipeline(_config: &mut BackgroundTaskConfig) -> Vec<Box<dyn SegmentProcessor>> {
-    Vec::new()
+    // Symbolize without S3: gzip and write back to disk.
+    #[cfg(feature = "cpu-profiling")]
+    if config.symbolize
+        && !pipeline
+            .iter()
+            .any(|p| p.name() == "Gzip" || p.name() == "S3Upload")
+    {
+        pipeline.push(Box::new(GzipWriteBackProcessor));
```
rcoh: It seems like we should gzip as a stage then writeback as a stage? Somewhat surprising to see them combined?

I guess there are three main parts:

- processing (currently symbol processing, optional)
- compression (should probably just always do it?)
- storage (disk vs. s3)

dial9-tokio-telemetry/src/telemetry/recorder/cpu_flush_state.rs
```diff
@@ -1,65 +1,25 @@
...
 #[cfg(feature = "cpu-profiling")]
-pub(super) struct CpuFlushState {
-    /// When true, symbolicate callframe addresses and emit CallframeDef events.
-    pub(super) inline_callframe_symbols: bool,
-    /// Addresses already symbolicated (across all files). Maps addr → (symbol, location).
-    pub(super) callframe_intern: HashMap<u64, (String, Option<String>)>,
-    /// Addresses whose CallframeDef has been emitted in the current file.
-    callframe_emitted_this_file: HashSet<u64>,
-}
+pub(super) struct CpuFlushState;
```
rcoh: Presumably we can now delete this empty struct?

dial9-tokio-telemetry/src/telemetry/recorder/cpu_flush_state.rs
```diff
@@ -1,65 +1,25 @@
...
-                        symbol,
-                        location,
-                    })));
-                    self.callframe_emitted_this_file.insert(addr);
-                }
-            }
-        }
-        batch.push(RawEvent::CpuSample(Box::new(data.clone())));
-        batch
+        vec![RawEvent::CpuSample(Box::new(data.clone()))]
```
rcoh: This function is now a no op?

dial9-tokio-telemetry/design/cpu-sampling-and-worker-id-design.md
jlizen: Currently just takes the first. I am thinking to emit multiple symbol table entries per address with an `inline_depth` field.

It isn't actually much work, though would scope creep a bit since viewer needs support. I'll take a timeboxed stab at it.

dial9-tokio-telemetry/README.md
jlizen: I'm fine with the new approach though. Semantics are pretty clear with the typestate pattern, and we just pass that through to the writer.

dial9-tokio-telemetry/Cargo.toml
jlizen: The worker gzip-compresses segments before writing back to disk, even without s3. We have the dependency enabled on both cpu profiling and s3-worker separately.

dial9-tokio-telemetry/src/background_task/mod.rs
jlizen: Yeah, that seems reasonable to move compression into the middle like you suggested.

jlizen:
@rcoh addressed your feedback:

## What changed

**Deleted `CpuFlushState` entirely.** The struct was empty, `on_rotate` was a no-op, and `resolve_cpu_event_symbols` just wrapped data in a vec. Inlined the `RawEvent::CpuSample(...)` construction directly in `write_cpu_event` and removed the file.

**Split `GzipWriteBackProcessor` into `GzipCompressor` + `WriteBackProcessor`.** The pipeline now follows the model you suggested: processing (symbolize, optional) → compression (gzip) → storage (disk or S3). `GzipCompressor` is no longer gated behind `worker-s3`. `WriteBackProcessor` writes bytes back to the segment path on disk. The non-S3 pipeline is now `Symbolize → Gzip → WriteBack` instead of `Symbolize → GzipWriteBack`.

**Added `SymbolTableEntry` support to the trace viewer.** The viewer was only consuming the old `CallframeDefEvent` path, so background symbolization output was silently ignored. Added a `SymbolTableEntryEvent` case to `trace_parser.js` that populates `callframeSymbols` from the new schema.

**Added inline depth to `SymbolTableEntry`.** New `inline_depth` field (0 = outermost, 1+ = inlined). `write_symbol_data` no longer deduplicates away inlined frames. The viewer expands inline chains when resolving callchains, so inlined functions show up in flamegraphs and stack traces.

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -181,16 +185,29 @@ pub(crate) trait SegmentProcessor: Send {
...
+        let s3_uploader = S3PipelineUploader::new(s3_config, config.client.take()).await;
+        pipeline.push(Box::new(GzipCompressor));
+        pipeline.push(Box::new(s3_uploader));
+    }
 
-#[cfg(not(feature = "worker-s3"))]
-async fn build_pipeline(_config: &mut BackgroundTaskConfig) -> Vec<Box<dyn SegmentProcessor>> {
-    Vec::new()
+    // Without S3: gzip and write back to disk.
+    if !pipeline.iter().any(|p| p.name() == "S3Upload") {
```
rcoh: Maybe just drop a todo in here, seems brittle

jlizen: I'll just clean it up

PR #97, opened 2026-03-18
add: Harden flush path with ArrayQueue & emit metrics

rcoh:
## What

- Replace `Mutex<Vec<Vec<RawEvent>>>` in `CentralCollector` with crossbeam `ArrayQueue` (bounded, lock-free). Capacity 1024 batches; excess batches are dropped with a `tracing::warn`.
- Remove park-triggered flush from `record_event` — was causing excessive tiny batch pushes. Events still reach the collector when the thread-local buffer fills (1024 events) or on thread exit.
- Flush every 5ms instead of every 250ms. Since ArrayQueue drain is lock-free, there's no contention cost.
- Nice the flush thread (`nice +10`) on Linux so it doesn't compete with worker threads for CPU time.
- Add `flush_path` benchmark for measuring flush-path performance outside the runtime.

## Flush metrics via metrique

- Move `metrics_sink` from `BackgroundTaskConfig` to `TracedRuntimeBuilder` (`.with_metrics_sink()`) so it's shared by both the flush thread and the background worker.
- Add `src/metrics.rs` with `Operation` enum (`Flush`, `ProcessSegment`) and two `#[metrics]` structs: `FlushMetrics` and `SegmentProcessMetrics`.
- Flush thread publishes `FlushMetrics` each cycle when there's activity (event_count, flush_duration, dropped_batches).
- Both structs carry an `Operation` field for filtering in dashboards.

Sample output:
```
---
  Operation: Flush
  EventCount: 9216
  FlushDuration: 11.095ms
  DroppedBatches: 0
---
  Operation: Flush
  EventCount: 1
  FlushDuration: 16.270μs
  DroppedBatches: 0
```

## Testing

All 276 tests pass. `cargo fmt` and `cargo clippy --all-targets --all-features` clean.


dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -406,6 +431,14 @@ impl TracedRuntimeBuilder {
         self
     }
 
+    /// Set the metrics sink for telemetry operational metrics (flush stats,
+    /// pipeline metrics, etc.). Defaults to
+    /// [`DevNullSink`](metrique_writer::sink::DevNullSink).
+    pub fn with_metrics_sink(mut self, sink: metrique_writer::BoxEntrySink) -> Self {
```
jlizen: We'll add eg a emf or json over stdout helper later? To avoid forcing caller to take a metrique dependency?

dial9-tokio-telemetry/src/telemetry/collector.rs
```diff
@@ -13,25 +20,41 @@ impl Default for CentralCollector {
...
-        std::mem::take(&mut *self.buffers.lock().unwrap())
+        let mut batches = Vec::new();
+        while let Some(batch) = self.queue.pop() {
+            batches.push(batch);
+        }
+        batches
     }
 
     pub fn len(&self) -> usize {
-        self.buffers.lock().unwrap().iter().map(|b| b.len()).sum()
```
jlizen, line 30: This went from total events, to total batches, is that desired? Should we track both?

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -485,12 +518,18 @@ impl TracedRuntimeBuilder {
             let rec = recorder.clone();
             let shared = recorder.lock().unwrap().shared.clone();
             let stop = stop.clone();
+            let flush_metrics_sink = self.metrics_sink.clone();
             std::thread::Builder::new()
                 .name("telemetry-flush".into())
                 .spawn(move || {
-                    let flush_interval = Duration::from_millis(250);
```
jlizen, line 595: This means that we're now busy looping every 5ms (down from 250ms). I know that arrayqueue doesn't allow blocking waits, but maybe we could have an adaptive approach to keep the wakes less frequent?: 
```
  let sleep_ms = if stats.event_count > 0 { 5 } else { min(sleep_ms * 2, 100) };
```

dial9-tokio-telemetry/src/telemetry/collector.rs
```diff
@@ -13,25 +20,41 @@ impl Default for CentralCollector {
...
     pub fn accept_flush(&self, buffer: Vec<RawEvent>) {
-        self.buffers.lock().unwrap().push(buffer);
+        if let Some(_evicted) = self.queue.force_push(buffer) {
+            self.dropped_batches.fetch_add(1, Ordering::Relaxed);
+        }
     }
 
     pub fn drain(&self) -> Vec<Vec<RawEvent>> {
-        std::mem::take(&mut *self.buffers.lock().unwrap())
+        let mut batches = Vec::new();
```
jlizen, line 44: if we keep the 5ms loop, might be good to reuse the allocation

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -505,9 +544,17 @@ impl TracedRuntimeBuilder {
                             }
                         }
 
-                        if now.duration_since(last_flush) >= flush_interval {
-                            last_flush = now;
-                            rec.lock().unwrap().flush();
+                        let mut flush_timer = Timer::start_now();
+                        let stats = rec.lock().unwrap().flush();
```
jlizen, line 649: this skews the timer with mutex acquisition time, should we make it inwards of the guard?

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -485,12 +518,18 @@ impl TracedRuntimeBuilder {
...
             let stop = stop.clone();
+            let flush_metrics_sink = self.metrics_sink.clone();
             std::thread::Builder::new()
                 .name("telemetry-flush".into())
                 .spawn(move || {
-                    let flush_interval = Duration::from_millis(250);
+                    // Lower this thread's scheduling priority so it doesn't
+                    // compete with worker threads for CPU time.
+                    #[cfg(target_os = "linux")]
+                    unsafe {
```
jlizen, line 629: Needs safety comment. Probably also good to handle a failure in the `nice` call.

dial9-tokio-telemetry/src/telemetry/collector.rs
```diff
@@ -1,8 +1,15 @@
 use crate::telemetry::events::RawEvent;
-use std::sync::Mutex;
+use crossbeam_queue::ArrayQueue;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+/// Maximum number of batches (each up to 1024 events) that can be buffered.
+/// Beyond this, the oldest batch is evicted — the queue acts as a ring buffer
+/// so the most recent data is always preserved.
+const DEFAULT_CAPACITY: usize = 1024;
```
jlizen, line 8: Note that currently our queue samples will be batches of 1 event, and fire 100 times/second. So in case of backpressure, the queue could quickly become consumed by samples.

Would be good to use thread local storage to buffer or otherwise offset.

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
rcoh: we are waking up anyway though, so I don't think draining the queue will have a big impact

dial9-tokio-telemetry/src/telemetry/collector.rs
rcoh: fixed by pushing queue samples into TL buffer

dial9-tokio-telemetry/src/telemetry/collector.rs
rcoh: fixed by adding `fn next` to pop individual events without draining into buffer first

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -173,24 +215,6 @@ impl TelemetryRecorder {
             }
         })
     }
-
-    pub fn start_sampler_task(
```
rcoh, line 177: this was dead code

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
rcoh: lock acquisition time seems relevant, we could eventually track both I guess

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -609,9 +645,18 @@ impl TracedRuntimeBuilder<HasTracePath> {
...
-                            last_flush = now;
-                            rec.lock().unwrap().flush();
+                        let mut flush_timer = Timer::start_now();
+                        let stats = rec.lock().unwrap().flush();
+                        flush_timer.stop();
+                        if stats.event_count > 0 || stats.dropped_batches > 0 {
+                            let _guard = FlushMetrics {
+                                operation: Operation::Flush,
+                                event_count: stats.event_count,
+                                flush_duration: flush_timer,
```
jlizen, line 655: still includes time spent acquiring mutex, which might be misleading since the name reads like it is just wall-lock for the flush cycle

dial9-tokio-telemetry/src/telemetry/collector.rs
```diff
@@ -1,8 +1,15 @@
 use crate::telemetry::events::RawEvent;
-use std::sync::Mutex;
+use crossbeam_queue::ArrayQueue;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+/// Maximum number of batches (each up to 1024 events) that can be buffered.
+/// Beyond this, the oldest batch is evicted — the queue acts as a ring buffer
```
jlizen, line 6: nit: `ArrayQueue` isn't a ringbuffer, though the semantics are similar

dial9-tokio-telemetry/benches/flush_path.rs
```diff
@@ -0,0 +1,198 @@
...
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc, Barrier};
+
+/// Build a realistic batch simulating a worker thread's activity.
+///
+/// A typical worker cycle is: unpark → (poll_start, poll_end) × N → park.
+/// We simulate ~170 polls per batch (340 events) plus park/unpark and a few
+/// spawns and wakes, totalling ~350 events. The batch is repeated ~3× to fill
+/// close to BUFFER_CAPACITY (1024).
+fn make_batch(worker: usize) -> Vec<RawEvent> {
```
jlizen, line 21: Might be nice to also cover the `force_psh` eviction. This won't produce enough batches currently to contend on the TLS.

PR #98, opened 2026-03-19
fix(js-decoder): handle truncated frames gracefully, read symbol frames even if >= MAX_EVENTS

jlizen:
# [dial9-trace-format] Fix JS decoder crash on truncated trace segments

### Summary

The JS trace decoder (`decode.js`) throws `RangeError: offset is outside the bounds of the DataView` when parsing traces from non-demo runs. This happens because the rotating writer may not flush the final segment cleanly, leaving a truncated frame at the end of the file. The decoder had no bounds checking, so any read past the end of the buffer was an unhandled exception that killed the entire parse.

The fix wraps the frame-decoding path in `nextFrame()` with a `RangeError` catch. When a truncated frame is encountered, the decoder stops gracefully (returns `null`, ending iteration) instead of throwing. This preserves all successfully decoded frames up to the truncation point.

This approach was chosen over sprinkling individual bounds checks before every `getUint8`/`getUint32`/etc. call because: the truncation only happens at segment boundaries (not mid-stream corruption), so catching at the `nextFrame` level is the right granularity; and it keeps the decode methods clean and fast for the common case.

### Testing

Existing Rust-driven JS parser integration tests (`dial9-trace-format::js_parser`) pass. The original bug was reproduced by running the metrics-service in non-demo mode and loading the resulting trace in the viewer.

It now loads properly.

### DIFF
--- a/dial9-tokio-telemetry/tests/js_parser.rs
+++ b/dial9-tokio-telemetry/tests/js_parser.rs
@@ -95,3 +95,80 @@ fn test_js_parser_matches_rust() {
         String::from_utf8_lossy(&test_output.stderr)
     );
 }
+
+/// Verify that SymbolTableEntry frames at the end of a trace are still resolved
+/// even when the event cap is reached (i.e., the parser doesn't break out of
+/// the frame loop early and skip trailing metadata).
+#[cfg(feature = "cpu-profiling")]
+#[test]
+fn test_js_parser_resolves_symbols_past_event_cap() {
+    use dial9_perf_self_profile::offline_symbolize::SymbolTableEntry;
+    use dial9_tokio_telemetry::telemetry::format::{PollEndEvent, WorkerId};
+    use dial9_trace_format::encoder::Encoder;
+
+    let temp_dir = TempDir::new().unwrap();
+    let trace_path = temp_dir.path().join("capped_trace.bin");
+
+    {
+        let mut enc = Encoder::new();
+        for i in 0..10u64 {
+            enc.write(&PollEndEvent {
+                timestamp_ns: i * 1_000_000,
+                worker_id: WorkerId::from(0usize),
+            })
+            .unwrap();
+        }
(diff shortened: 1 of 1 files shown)


rcoh (approved):
Can you add a test to the existing js tests? You can create a truncated trace.

I am somewhat surprised we had a partially written frame because although we can obviously drop entire records 'm not sure how we would write a partial one

If the data is in the BufWrite, we should make sure we flush that data during shutdown

dial9-tokio-telemetry/trace_viewer/trace_parser.js
```diff
@@ -44,22 +45,22 @@
...
 
+        const capped = () => events.length >= maxEvents;
+
         for (const frame of frames) {
-            if (events.length >= MAX_EVENTS) break;
             if (frame.type !== 'event') continue;
             const v = frame.values;
             const ts = num(frame.timestamp_ns);
 
             switch (frame.name) {
```
rcoh, line 55: probably cleaner to do `if (capped() && !is_meta(frame.name)) { ... }` instead of handling it on each branch

dial9-tokio-telemetry/trace_viewer/trace_parser.js
```diff
@@ -129,21 +130,13 @@
...
                         tid: num(v.tid), source: num(v.source), callchain: chain,
                     });
-                    // thread_name is a PooledString, already resolved
                     const tn = v.thread_name;
                     if (tn && tn !== '<no thread name>') {
                         threadNames.set(num(v.tid), tn);
                     }
                     break;
                 }
-                case 'CallframeDefEvent': {
```
jlizen, line 139: dead code

dial9-tokio-telemetry/trace_viewer/trace_parser.js
jlizen: Ah yeah, I'll fix it up tomorrow

PR #100, opened 2026-03-19
feat(metrics-service): Add env-gated FD table pre-warming

jlizen:
# [metrics-service] Add env-gated FD table pre-warming

Related: https://github.com/tokio-rs/tokio/issues/7970

### Summary

Adds an opt-in FD table pre-warming step to the metrics-service example, mitigating kernel RCU-synchronization stalls that cause latency spikes when the FD table grows under load in multi-threaded processes.

When `PREWARM_FD_TABLE_SIZE=N` is set, the service opens `/dev/null` and uses `fcntl(fd, F_DUPFD_CLOEXEC, N)` to force the kernel to expand the FD table to at least `N` entries before the tokio runtime starts. Both FDs are closed immediately; the table capacity persists for the process lifetime since the kernel never shrinks it.

`F_DUPFD_CLOEXEC` was chosen over `dup`/`dup2` because it has "lowest free fd >= target" semantics (no loop needed, single syscall), won't clobber existing FDs, and sets `CLOEXEC` to avoid leaking to the child client process.

Linux-only (`#[cfg(target_os = "linux")]`). When the env var is unset, behavior is unchanged.

### Testing

Compiled and verified with `cargo check`, `cargo clippy`, and `cargo fmt`. This is an example binary, so the way to validate the effect is to run the service with and without the env var and compare traces visually:

```bash
# baseline
cargo run --release -p metrics-service --bin metrics-service -- --demo

# with pre-warming
PREWARM_FD_TABLE_SIZE=10000 cargo run --release -p metrics-service --bin metrics-service -- --demo
```

Without the warming, I observed the same ~80ms long-poll due to heavy `alloc_fd` calls that @rcoh previously observed.

Screenshot 2026-03-18 at 9 17 40 PM

With warming, I saw no such blocking calls.

Screenshot 2026-03-18 at 9 19 06 PM

### DIFF
--- a/examples/metrics-service/README.md
+++ b/examples/metrics-service/README.md
@@ -37,3 +37,5 @@ curl http://localhost:3001/metrics/cpu
 Everything is controlled via CLI flags—run `cargo run -- --help` for the full list. The load profile constants (`MAX_WORKERS`, `THUNDERING_HERD`, `BASELINE`, etc.) live at the top of `src/client.rs`.
 
 On Linux, CPU profiling and scheduler event tracing are enabled automatically.
+
+Set `PREWARM_FD_TABLE_SIZE=N` to pre-warm the kernel FD table before the runtime starts, mitigating RCU-synchronization latency spikes from FD table growth under load (see [tokio#7970](https://github.com/tokio-rs/tokio/issues/7970)).


jlizen:
I think that trace was partially corrupted.

Here is a new one that is more illustrative, I'll update the description as well:
Screenshot 2026-03-18 at 9 10 58 PM

PR #102, opened 2026-03-19
fix: Fix worker spin-loop on gzip-compressed and permanently failing segments

jlizen:
### Summary

The background worker pipeline (Symbolize → Gzip → WriteBack) could spin-loop on segments left over from a previous run. `WriteBackProcessor` overwrites `.bin` files with gzip-compressed content, so on the next start the worker reads them back expecting raw `TRC\0` trace data, fails header validation, and immediately retries every poll cycle.

Fixes:

1. `SymbolizeProcessor` and `GzipCompressor` detect gzip magic bytes (`1f 8b`) and pass through unchanged.
2. Segments that permanently fail processing are deleted from disk, preventing spin-loops on corrupted or unrecoverable files.

### Testing

- `failed_segment_deleted`: a permanently failing processor triggers deletion of the segment file.
- `gzip_segment_not_double_compressed`: gzip input passes through `GzipCompressor` unchanged.

### DIFF
--- a/dial9-tokio-telemetry/trace_viewer/index.html
+++ b/dial9-tokio-telemetry/trace_viewer/index.html
@@ -355,7 +355,7 @@
                     Drop a <code>.bin</code> trace file here or click to open
                 </div>
                 <div style="font-size: 0.8em; margin-top: 8px; color: #666">
-                    Expects TOKIOTRC binary format
+                    Expects D9TF binary format
                 </div>
                 <div style="margin-top: 12px">
                     <a href="#" id="load-demo" style="color: #6c63ff; font-size: 0.9em">or load demo trace</a>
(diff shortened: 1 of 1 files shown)


dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -542,7 +553,10 @@ impl WorkerLoop {
                         {
                             tracing::debug!(target: "dial9_worker", path = %segment.path.display(), "segment evicted during processing, skipping");
                         } else {
-                            tracing::warn!(target: "dial9_worker", error = %e.kind, cause = ?e.kind, path = %segment.path.display(), "processor failed, skipping segment");
+                            if let Err(remove_err) = std::fs::remove_file(&segment.path) {
+                                tracing::warn!(target: "dial9_worker", error = %remove_err, path = %segment.path.display(), "failed to remove corrupted segment");
```
rcoh, line 557: We should eventually publish a metric here

PR #111, opened 2026-03-19
fix: offline symbolization cleanups and optimizations

rcoh:
- the old symbolize path had a bunch of dead code that was unused from the previous work that serialized the ProcMaps into the stream
- the symbolizer can actually batch symbolize (so I switched to use that)
- avoid re-allocating strings prior to interning

there are still some improvements that can be made (documented)

### DIFF
--- a/dial9-trace-format/src/encoder.rs
+++ b/dial9-trace-format/src/encoder.rs
@@ -3,7 +3,7 @@
 use crate::TraceEvent;
 use crate::codec::{self, PoolEntry, WireTypeId};
 use crate::schema::{SchemaEntry, SchemaRegistry};
-use crate::types::{EncodeState, EventEncoder};
+use crate::types::{EncodeState, EventEncoder, InternedString};
 use std::any::TypeId;
 use std::collections::HashMap;
 use std::io::{self, Write};
@@ -281,9 +281,9 @@ impl<W: Write> Encoder<W> {
     }
 
     /// Intern a string, emitting a pool frame if new. Returns an [`InternedString`] handle.
-    pub fn intern_string(&mut self, s: &str) -> io::Result<crate::types::InternedString> {
+    pub fn intern_string(&mut self, s: &str) -> io::Result<InternedString> {
         if let Some(&id) = self.string_pool.get(s) {
-            return Ok(crate::types::InternedString(id));
+            return Ok(InternedString(id));
         }
         let id = self.next_pool_id;
         self.next_pool_id += 1;
@@ -295,7 +295,7 @@ impl<W: Write> Encoder<W> {
             }],
             &mut self.state.writer,
         )?;
-        Ok(crate::types::InternedString(id))
+        Ok(InternedString(id))
     }
 
     pub fn write_string_pool(&mut self, entries: &[PoolEntry]) -> io::Result<()> {
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,175 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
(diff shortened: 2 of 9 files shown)


perf-self-profile/src/offline_symbolize.rs
```diff
@@ -162,109 +78,141 @@ fn collect_stack_frame_addresses(values: &[FieldValueRef<'_>], addresses: &mut B
...
-                Some(ci) => (ci.file, ci.line.unwrap_or(0) as u64),
-                None => (String::new(), 0),
-            };
+    // Batch-resolve per ELF mapping.
+    for (map_idx, offsets_and_addrs) in &user_groups {
+        let entry = &maps[*map_idx];
+        let offsets: Vec<u64> = offsets_and_addrs.iter().map(|(o, _)| *o).collect();
+        let addrs: Vec<u64> = offsets_and_addrs.iter().map(|(_, a)| *a).collect();
+        let src = source::Source::Elf(source::Elf::new(&entry.path));
+        if let Ok(results) = symbolizer.symbolize(&src, Input::FileOffset(&offsets)) {
```
jlizen: We should probably write out placeholder entries, similar to kernel addresses, rather than silently dropping the whole batch. I found that placeholder behavior pretty useful earlier.

perf-self-profile/src/offline_symbolize.rs
```diff
@@ -162,109 +78,141 @@ fn collect_stack_frame_addresses(values: &[FieldValueRef<'_>], addresses: &mut B
...
+            }
+        }
+    }
 
-            let (source_file_str, source_line) = match info.code_info {
-                Some(ci) => (ci.file, ci.line.unwrap_or(0) as u64),
-                None => (String::new(), 0),
-            };
+    // Batch-resolve per ELF mapping.
+    for (map_idx, offsets_and_addrs) in &user_groups {
```
jlizen, line 142: Do we care about deterministic ordering at all? Old approach used a `BTreeSet`.

dial9-trace-format/src/encoder.rs
```diff
@@ -108,6 +108,57 @@ impl<W: Write> Encoder<W> {
...
+    /// writes new frames into `writer` **without** emitting a file header.
+    /// Use this when you need to append events (e.g. symbol tables) to a trace
+    /// that already has a header.
+    ///
+    /// Returns `None` if `existing` does not contain a valid trace header.
+    pub fn extend(existing: &[u8], writer: W) -> Option<Self> {
+        use crate::decoder::Decoder;
+
+        let mut decoder = Decoder::new(existing)?;
+        while decoder.next_frame_ref().ok().flatten().is_some() {}
```
jlizen: This would result in a decode error midway through ending the iteration, but making it look like happy path. Which then will produce weird corrupted states.

Consider bubbling up the error instead.

perf-self-profile/src/offline_symbolize.rs
```diff
@@ -162,109 +78,141 @@ fn collect_stack_frame_addresses(values: &[FieldValueRef<'_>], addresses: &mut B
...
-                data: s.into_bytes(),
-            });
-            id
-        });
-        InternedString::from_raw(id)
-    };
+    // Partition addresses into kernel vs userspace, group userspace by mapping.
+    let mut kernel_addrs: Vec<u64> = Vec::new();
+    // (mapping_index, file_offset, original_addr)
+    let mut user_groups: std::collections::HashMap<usize, Vec<(u64, u64)>> =
```
jlizen: nit: probably import hashmap up top

perf-self-profile/src/offline_symbolize.rs
```diff
@@ -273,174 +221,34 @@ mod tests {
...
+        types::{FieldType, FieldValue},
     };
 
-    fn make_encoder_with_proc_maps(maps: &[MapsEntry]) -> Vec<u8> {
-        let mut buf = Encoder::new().finish();
-        encode_proc_maps(WireTypeId(100), maps, &mut buf).unwrap();
-        buf
-    }
-
-    #[test]
```
jlizen, line 286: It seems like a good chunk of these test cases are still relevant, should we adapt them?

perf-self-profile/src/offline_symbolize.rs
rcoh: done, now reports `[symbolize-failed]` when the whole batch failed

perf-self-profile/src/offline_symbolize.rs
rcoh: I'm not sure why we would...

dial9-trace-format/src/encoder.rs
rcoh: code unused outside of tests, deleted

perf-self-profile/src/offline_symbolize.rs
rcoh: backported some

PR #115, opened 2026-03-19
fix: write segment metadata at the beginning of the file and add RotatingWriterBuilder

rcoh:
I noticed that the invalid segment metadata metric was firing. Turns out we were failing to write segment metata on the first segment.

While I was in the neighborhood, I converted RotatingFileWriter to into a builder. The old API was preserved.

### DIFF
--- a/.github/workflows/stress-test.yml
+++ b/.github/workflows/stress-test.yml
@@ -1,0 +1,36 @@
+name: Stress Test
+on:
+  workflow_dispatch:
+    inputs:
+      duration_minutes:
+        description: "Stress test duration in minutes"
+        required: false
+        default: "60"
+        type: string
+  schedule:
+    - cron: "0 6 * * *"
+permissions:
+  contents: read
+
+jobs:
+  stress-test:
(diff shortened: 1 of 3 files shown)


dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -1090,7 +1095,7 @@ mod tests {
                 let dir = tempfile::TempDir::new().unwrap();
                 let base = dir.path().join("trace");
 
-                let writer = RotatingWriter::new(&base, max_file_size, 1_000_000).unwrap();
+                let writer = RotatingWriter::builder().base_path(&base).max_file_size(max_file_size).max_total_size(1_000_000).build().unwrap();
```
jlizen: nit: fmt

dial9-tokio-telemetry/src/telemetry/writer.rs
jlizen, line 164: Probably worth documenting that this doesn't allow custom segment metadata

PR #121, opened 2026-03-20
cleanup code and benches

rcoh:
Simplify flush_path bench, clean up buffer/collector/recorder code.

- Reduce flush_path bench boilerplate
- Clean up buffer and collector modules
- Remove unused code from recorder

PR #122, opened 2026-03-20
refactor: eliminate recorder mutex + encapsulate BUFFER thread-local

rcoh:
Stacked on #121

**Commit 1: eliminate Arc<Mutex<TelemetryRecorder>>**

Replace the shared mutex around TelemetryRecorder with a channel-based design where the flush thread is the sole owner of EventWriter.

**Commit 2: hide BUFFER thread-local behind free functions**

Make the BUFFER thread_local private to buffer.rs and expose two free functions:
- `buffer::record_event(event, collector)` — record + auto-flush when full
- `buffer::drain_to_collector(collector)` — force-drain at shutdown/flush

Removes BUFFER.with boilerplate from 4 call sites.

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
+++ b/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
@@ -223,8 +223,6 @@ struct WorkerHandle {
 }
 
 /// RAII guard returned by [`TracedRuntimeBuilder::build`].
-///
-/// Controls the flush thread via a channel — no shared mutex.
 pub struct TelemetryGuard {
     handle: TelemetryHandle,
     flush_thread: Option<std::thread::JoinHandle<()>>,


dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -224,7 +196,13 @@ impl TelemetryHandle {
 
     pub fn disable(&self) {
```
jlizen, line 195: This needs a doc comment. That comment should mention that this will block if in mid-flush, since we only have capacity of 1.

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
jlizen, line 774: Would be good to have tests for:
- calling disable after events are already in the buffer (make sure they are flushed)
- graceful shutdown path doesn't seem to have coverage

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
rcoh: possibly since you reviewed it, I changed this so it doesn't actually block anymore. it just sets the flag. that seemed better anyway

PR #129, opened 2026-03-23
breaking: Cleanup public interfaces

rcoh:
Remove lots of types that shouldn't be pub. There are more, but we'll get there when we get there.

### DIFF
--- a/dial9-tokio-telemetry/src/analysis_unstable.rs
+++ b/dial9-tokio-telemetry/src/analysis_unstable.rs
@@ -9,11 +9,14 @@
 //! # Quick start
 //!
 //! ```no_run
+//! # fn main() -> std::io::Result<()> {
 //! use dial9_tokio_telemetry::analysis_unstable::{TraceReader, analyze_trace, print_analysis};
 //!
 //! let reader = TraceReader::new("trace.0.bin")?;
 //! let analysis = analyze_trace(&reader.runtime_events);
 //! print_analysis(&analysis, &reader.spawn_locations);
+//! # Ok(())
+//! # }
 //! ```
 
 pub use crate::telemetry::analysis::{
--- a/dial9-tokio-telemetry/README.md
+++ b/dial9-tokio-telemetry/README.md
@@ -126,7 +126,7 @@ Both of these events are tied to the precise instant and thread that they happen
 # #[cfg(feature = "cpu-profiling")]
 # fn main() -> std::io::Result<()> {
 # use dial9_tokio_telemetry::telemetry::{RotatingWriter, TracedRuntime};
-use dial9_tokio_telemetry::telemetry::{CpuProfilingConfig, SchedEventConfig};
+use dial9_tokio_telemetry::telemetry::cpu_profile::{CpuProfilingConfig, SchedEventConfig};
 
 # let writer = RotatingWriter::new("/tmp/t.bin", 1024, 4096)?;
 # let builder = tokio::runtime::Builder::new_multi_thread();


dial9-tokio-telemetry/src/background_task/instance_metadata.rs
```diff
@@ -7,7 +7,7 @@
 /// component in S3 object keys.
 ///
 /// # Examples
-/// ```
+/// ```ignore
```
jlizen: no_run ?

dial9-tokio-telemetry/Cargo.toml
```diff
@@ -40,11 +40,13 @@ metrique-writer = "0.1"
 metrique = { version = "0.1.20", features = ["local-format"] }
 
 [features]
+analysis = []
```
jlizen, line 43: Does analysis side use the rest of the API? Do we want a `dial9-viewer` crate?

dial9-tokio-telemetry/src/analysis_unstable.rs
```diff
@@ -0,0 +1,24 @@
...
+//!
+//! This module is feature-gated behind `analysis` and re-exports every public
+//! type and function from the internal `telemetry::analysis` implementation.
+//!
+//! The API surface here is still evolving — hence the `_unstable` suffix.
+//! Expect breaking changes between minor versions.
+//!
+//! # Quick start
+//!
+//! ```rust,ignore
```
jlizen: no_run?

dial9-tokio-telemetry/src/analysis_unstable.rs
```diff
@@ -0,0 +1,24 @@
+//! Trace analysis utilities (**unstable**).
+//!
+//! This module is feature-gated behind `analysis` and re-exports every public
+//! type and function from the internal `telemetry::analysis` implementation.
+//!
+//! The API surface here is still evolving — hence the `_unstable` suffix.
```
jlizen: nit: style

jlizen (approved):
re: semver bump:

This would break anybody using the analysis API, right? Both due to new feature + type locations moving around?

I think we probably could get away with it due to how clear we've been that this is early/experimental software. Certainly people are unlikely to use this API anyway, instead the html viewer.

My preference is a deprecation cycle before next breaking version, since we might have closed source consumers we can't go fix, but it's ok if you don't have the time

dial9-tokio-telemetry/src/telemetry/mod.rs
```diff
@@ -1,27 +1,27 @@
...
-    ActivePeriod, LongPoll, SampledPoll, SchedDelay, SpawnLocationStats, TraceAnalysis,
-    TraceReader, WakeDelay, WorkerStats, analyze_trace, compute_active_periods,
-    compute_wake_to_poll_delays, detect_idle_workers, detect_long_polls, detect_sampled_polls,
-    detect_sched_delays, detect_wake_delays, print_analysis,
-};
 #[cfg(feature = "cpu-profiling")]
 pub use cpu_profile::CpuProfilingConfig;
 #[cfg(feature = "cpu-profiling")]
 pub use cpu_profile::SchedEventConfig;
 pub use dial9_trace_format::InternedString;
```
jlizen: InternedString feels a bit exposed in our public API, at least our re-exported one.

dial9-tokio-telemetry/src/background_task/instance_metadata.rs
jlizen: Nit: no_run, at least? Adding `ignore` makes me nervous given the interface changes

dial9-trace-format/src/codec.rs
```diff
@@ -12,41 +17,44 @@ use std::io::{self, Write};
...
 #[inline]
-pub fn decode_u24_le(data: &[u8]) -> Option<u32> {
+pub(crate) fn decode_u24_le(data: &[u8]) -> Option<u32> {
     let b = data.get(..3)?;
     Some(b[0] as u32 | (b[1] as u32) << 8 | (b[2] as u32) << 16)
 }
 
+/// An owned string pool entry.
 #[derive(Debug, Clone, PartialEq)]
 pub struct PoolEntry {
```
jlizen, line 49: Non_exhaustive if wide open?

PR #133, opened 2026-03-24
Add FxHash + encoder infrastructure for thread-local encoding

rcoh:
This is part of series of PRs working on setting up #109 that moves the design from keeping Rust structs in memory to serializing the events basically the instant we get them.
 
### Encoder changes for the thread-local encoding pipeline:

- **FxHash**: replace std HashMap with FxHashMap for string_pool and schema_ids. Includes write_u32/u16/usize specializations to avoid byte-by-byte fallback for common key types.
- **reset_to(W)**: swap writer and clear internal state while preserving HashMap allocations, enabling encoder reuse across flush cycles.
- **write_raw()/reset_state()**: raw byte passthrough and state reset for the reset-frame pattern.
- **write_infallible()/intern_string_infallible()**: panic-on-error convenience methods for `Encoder<Vec<u8>>` where writes cannot fail.
- **Schema.entry** promoted to `pub(crate)`.
- **schema.rs**: SchemaRegistry fields promoted to `pub(crate)`.
- **types.rs**: add `write_field_value_ref_typed()` and `raw_data()` accessors on StackFramesRef/StringMapRef for zero-copy encoding.

Part 1 of 3 for the thread-local encoding feature. PRs 2 and 3:
- #134 : Decoder reset-frame support
- #135 : Buffer encode-inline + raw-copy flush (stacked on 1+2)

PR #134, opened 2026-03-24
Add reset-frame support + try_for_each_event to decoder

rcoh:
Decoder changes for the thread-local encoding pipeline:

- **Reset frames**: when a valid TRC\0 header appears mid-stream, the decoder resets schemas, string pool, and timestamp base, then continues. This enables raw-copy concatenation of independently encoded batches. Handled in `next_frame`, `next_frame_ref`, and `try_for_each_event`.
- **try_for_each_event**: like `for_each_event` but the callback may return `Err` to stop early. Includes `TryForEachError` enum.
- **for_each_event** simplified to delegate to `try_for_each_event`, eliminating ~100 lines of duplicated event-decoding logic.
- **schema_cache** switched from `HashMap<WireTypeId, _>` to `Vec<Option<_>>` indexed by `WireTypeId.0` for faster lookup.
- **JS decoder** (decode.js): mid-stream header detection and state reset.

Part 2 of 3 for the thread-local encoding feature. Independent of PR #133.
- #133: Encoder infrastructure
- #3: Buffer encode-inline + raw-copy flush (stacked on 1+2)

### DIFF
--- a/dial9-trace-format/src/decoder.rs
+++ b/dial9-trace-format/src/decoder.rs
@@ -171,7 +171,7 @@ impl<'a> Decoder<'a> {
     /// If the current position starts with a valid header, reset state and
     /// skip past it, returning true.
     fn try_consume_reset_header(&mut self) -> bool {
-        if self.data.len() - self.pos >= HEADER_SIZE
+        if self.pos + HEADER_SIZE <= self.data.len()
             && codec::decode_header(&self.data[self.pos..]).is_some()
         {
             self.reset_state();


dial9-trace-format/src/decoder.rs
```diff
@@ -140,6 +158,30 @@ impl<'a> Decoder<'a> {
...
+        self.registry = SchemaRegistry::new();
+        self.schema_cache.clear();
+        self.string_pool = StringPool::new();
+        self.timestamp_base_ns = 0;
+    }
+
+    /// If the current position starts with a valid header, reset state and
+    /// skip past it, returning true.
+    fn try_consume_reset_header(&mut self) -> bool {
+        if self.data.len() - self.pos >= HEADER_SIZE
```
jlizen:
```suggestion
        if self.pos + HEADER_SIZE <= self.data.len()
```

Issue #43, opened 2026-03-02
design and implement support for runtime-per-core setups


rcoh:
Currently the design has several limitations:
- cpu profiling is global. If you instrument each runtime and enable CPU profiling you will end up with duplicates
- you can only install it on one runtime. We actually want to install it in all runtimes and act as if each runtime is its own worker

Schedule instrumentation should work ok since that is per thread.

PR #141, opened 2026-03-26
#43: support multiple runtimes

vavrusa:
Hi! My use case is I need to trace several runtimes at the same time to get a better idea of how this is working - e.g. socket receivers in separate runtimes and task processing in separate runtime. I split this change into two parts - adding a way to attach additional runtimes onto an existing TelemetryGuard, and then write runtime metadata into trace metadata. I tried to do it in a way that is compatible with the existing format (see reasoning in f7638450e96e9623b1078debc4235609b2b34c93). The only main change is that worker-ids are globally unique in the service/trace so if you have e.g. 2 runtimes with 2 workers each, the first runtime is going to have worker-ids [0, 1] and the second runtime [2, 3]. In the UI the workers are then organized by the assigned runtime name. Not sure if the metadata is the best place for this, but it seemed to me because - 1. it's already used for service name, 2. it doesn't require format changes so existing trace files are compatible.

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
+++ b/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
@@ -1304,6 +1304,78 @@ mod tests {
         );
     }
 
+    /// Wake events from runtime B's workers must carry global worker IDs (≥ num_workers_a),
+    /// not local indices that collide with runtime A's workers.
+    #[test]
+    fn wake_events_use_global_worker_id_in_multi_runtime() {
+        use crate::telemetry::events::RawEvent;
+        use std::sync::{Arc, Mutex};
+
+        struct CapturingWriter(Arc<Mutex<Vec<RawEvent>>>);
+        impl crate::telemetry::writer::TraceWriter for CapturingWriter {
+            fn write_event(&mut self, event: &RawEvent) -> std::io::Result<()> {
+                self.0.lock().unwrap().push(event.clone());
+                Ok(())
+            }
+            fn flush(&mut self) -> std::io::Result<()> {
+                Ok(())
+            }
+        }
+
+        let events = Arc::new(Mutex::new(Vec::new()));
+
+        let mut builder_a = tokio::runtime::Builder::new_multi_thread();
+        builder_a.worker_threads(2);
+        let (runtime_a, guard) = TracedRuntime::builder()
+            .with_task_tracking(true)
+            .build_and_start_with_writer(builder_a, CapturingWriter(events.clone()))
+            .unwrap();
+
+        let mut builder_b = tokio::runtime::Builder::new_multi_thread();
+        builder_b.worker_threads(2);
+        let runtime_b = TracedRuntime::builder()
+            .with_task_tracking(true)
+            .build_with_reuse(builder_b, &guard)
+            .unwrap();
+
+        // Use handle.spawn on runtime B to get Traced waker wrapping → wake events.
+        let handle = guard.handle();
+        runtime_b.block_on(async {
+            let mut handles = Vec::new();
+            for _ in 0..50 {
+                handles.push(handle.spawn(async {
+                    tokio::task::yield_now().await;
+                }));
+            }
+            for h in handles {
+                h.await.unwrap();
+            }
+        });
+
+        drop(runtime_a);
+        drop(runtime_b);
+        drop(guard);
+
+        let captured = events.lock().unwrap();
+        let wake_workers: Vec<u8> = captured
+            .iter()
+            .filter_map(|e| match e {
+                RawEvent::WakeEvent { target_worker, .. } => Some(*target_worker),
+                _ => None,
+            })
+            .collect();
+        assert!(!wake_workers.is_empty(), "expected at least one WakeEvent");
+
+        // Runtime A has workers 0,1. Runtime B has workers 2,3.
+        // Wakes issued from runtime B's workers must have target_worker >= 2.
+        let has_global_id = wake_workers.iter().any(|&w| w >= 2 && w != 255);
+        assert!(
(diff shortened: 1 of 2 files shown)


dial9-tokio-telemetry/src/telemetry/recorder/shared_state.rs
```diff
@@ -86,8 +117,17 @@ pub(crate) struct SharedState {
...
+    /// Used to resolve the current thread's worker identity.
+    pub(crate) metrics: ArcSwap<HashMap<u64, RuntimeMetrics>>,
+    /// Counter for allocating unique runtime indices.
+    pub(crate) next_runtime_index: AtomicU64,
+    /// Global worker ID counter. Each worker gets a unique ID on first resolution.
+    pub(crate) next_worker_id: AtomicU64,
+    /// Maps runtime_index → human-readable runtime name (for segment metadata).
+    pub(crate) runtime_names: RwLock<HashMap<u64, String>>,
+    /// Maps worker_id → runtime_index. Populated lazily as workers are resolved.
+    pub(crate) worker_runtimes: RwLock<HashMap<u64, u64>>,
```
rcoh: I'm in the process of splitting the tokio stuff out from a dial9 core that isn't coupled to Tokio. I think as part of this PR, splitting out the _tokio specific state_ into its own struct would probably be helpful

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
rcoh, line 132: rather than push the complexity of multi-runtime into shared state, what if we captured the runtime id during hook registration?

then all of these methods could do utilize the runtime id when _creating_ the event (effectively, pushing the multi-runtime state directly into this closure rather than having it in the `SharedState` for all users?

vavrusa:
Ok let me try that! This works with multiple current-thread runtimes, I think what you're getting at is multiple current-thread runtimes switching on the same thread? Let me think about how that would work, the issue is caching on the TLS obviously, it should be possible to invalidate it with a hook though or move the thread-runtime association somewhere else.

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
vavrusa: That could work, I think it was the tracking of blocking threads was the issue, but good point I'll try again.

dial9-tokio-telemetry/src/telemetry/recorder/shared_state.rs
vavrusa: Slightly related - I would also be interested in this to support recording inside WASM containers (therefore not importing specific tokio bits)

vavrusa: Ok so I went through a few versions of this, I will extract the tokio and runtime bits to a separate structure to avoid overloading SharedState but leave general refactoring to split all tokio related bits to another PR (as you said you're already working on this). There's a few more leaky abstractions like `create_wake_event`.

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
vavrusa: Ok - I separated out the runtime bits to RuntimeContext, captured in the hook registration closure, and moved the registration to on_thread_unpark (effectively). The on_thread_start doesn't have runtime metrics yet at that point so it's still only for cpu profiling.

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -66,105 +67,64 @@ impl RuntimeContext {
...
-    /// Local queue depth for a resolved worker in this runtime.
+    /// Local queue depth for a worker in this runtime.
     fn local_queue_depth(&self, worker_index: usize) -> usize {
         self.metrics_and_base
             .get()
             .map(|(m, _)| m.worker_local_queue_depth(worker_index))
             .unwrap_or(0)
     }
+
+    /// Resolve the current thread's global worker ID using `tokio::runtime::worker_index()`.
```
vavrusa, line 78: Ah good catch, I tried to make the global state a single variable but this is much nicer.

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -66,105 +67,64 @@ impl RuntimeContext {
...
+    fn resolve_worker(&self, shared: &SharedState) -> Option<(WorkerId, usize)> {
+        let local_index = tokio::runtime::worker_index()?;
+        let (_, base) = self.metrics_and_base.get()?;
+        let global_id = base + local_index as u64;
+
+        register_worker_if_needed(self, local_index, global_id);
+        #[cfg(feature = "cpu-profiling")]
+        register_tid_if_needed(local_index, shared);
```
vavrusa, line 86: So before index was scoped to ResolvedWorker which has a runtime defined, this makes it a local index again but there's no context anymore so I don't think that will work.

vavrusa: Ah nvm I see that's fixed in the next

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
rcoh: I added this API a few weeks ago and it was released today, just in time :-)

rcoh:
I think the refactor isn't that big after I landed the PR so that TL
buffers contain serialized events.

Not sure I haven't looked deeply.

The main result (which will clean up this API before I cut 0.3 is that you
can spawn the dial9 recorder then attach individual runtimes to it.)

Right now you kinda need one to be the "host"

PR #149, opened 2026-04-01
Fix cpu_profile_workload symbolization and relative trace path handling

destinyFvcker:
## Summary

This PR fixes a background worker bug for bare relative trace paths like `cpu_profile_trace.bin` and improves the `cpu_profile_workload` example

## What changed

- Fix `BackgroundTaskConfig::trace_dir()` so bare relative filenames do not resolve to an empty parent path.
- Add a test for relative `trace_path` handling.
- Make the CPU-burning workload more robust in release builds so sampling at 99 Hz still captures useful frames.
- Clarify that `graceful_shutdown()` is required to seal the final segment and let worker post-processing complete.
- Enhance TraceReader's functionality to also read .bin files whose content is actually gzip.

## Why

Before this change, `cpu_profile_workload` could leave users with a `.bin` file that still contained only raw hex frame addresses. This is confusing because the example appears to succeed, but the viewer output looks broken.

There were two main issues:

- the worker could fail to discover sealed segments when `trace_path` was a bare relative filename(with out `./`).
- In fact, I found that symbolization only works correctly when using RotateWriter and simultaneously using `TracedRuntime::builder()` with the same `trace_base` specified in `with_trace_path`.

## User impact

After this change:

- using `cpu_profile_trace.bin` as a relative trace path works correctly
- users are less likely to open an unsymbolized file and see only `0x...` frames
- the example better explains when symbols should appear and why `graceful_shutdown()` matters

### DIFF
--- a/dial9-tokio-telemetry/examples/cpu_profile_workload.rs
+++ b/dial9-tokio-telemetry/examples/cpu_profile_workload.rs
@@ -4,7 +4,7 @@
 //! and prints any CpuSample events found.
 //!
 //! Run with:
-//!   RUSTFLAGS="-C force-frame-pointers=yes" cargo run --release --features cpu-profiling --example cpu_profile_workload
+//!   RUSTFLAGS="--cfg tokio_unstable -C force-frame-pointers=yes" cargo run --release --features cpu-profiling --example cpu_profile_workload
 //!
 //! You may need:
 //!   echo 2 | sudo tee /proc/sys/kernel/perf_event_paranoid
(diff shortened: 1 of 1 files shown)


dial9-tokio-telemetry/examples/cpu_profile_workload.rs
```diff
@@ -56,15 +67,28 @@ fn main() {
...
+
+    // Read back and report. TraceReader auto-detects gzip and parses
+    // SymbolTableEntry events into callframe_symbols.
+    eprintln!("\n=== Reading trace from {segment_path} ===");
+    let mut reader = dial9_tokio_telemetry::telemetry::TraceReader::new(segment_path).unwrap();
     let (magic, version) = reader.read_header().unwrap();
     eprintln!("Format: {magic} v{version}");
 
     let events = reader.read_all().unwrap();
+    // let syms = &reader.callframe_symbols;
```
rcoh: did you mean to leave these?

destinyFvcker: oh, I forgot to delete it😳.

PR #152, opened 2026-04-01
feat: Native axum executor integration

### LINKED ISSUE #62: bridge to hyper-util executor for the dial9 spawn hook
Currently our `axum_traced` example handrolls much of `axum::serve`. We do this because they hardcode `TokioExecutor` inline. We want to override it with an instrumented executor.

Ideally, we could contribute a knob on the axum API to allow specifying a custom executor. Worst case maybe dial9 can absorb some of the boilerplate into its own API (less preferred).

https://github.com/dial9-rs/dial9-tokio-telemetry/blob/main/examples/metrics-service/src/axum_traced.rs#L132

jv1i:
## Summary

Closes #62 

Implements the new `axum::serve::Executor` on `TelemetryHandle` behind `features = ["axum"]`
Deletes `axum_traced.rs` (previous hand-rolled serve loop) and simplifies metrics-service example to use `axum::serve` directly.

**Depends on this being released: https://github.com/tokio-rs/axum/pull/3704**

## Before

Users had to reimplement axum's serve loop to inject traced task spawning:

```rust
// axum_traced.rs: custom Serve, WithGracefulShutdown,
// TracedExecutor, IncomingStream, accept loop, graceful shutdown...
axum_traced::serve(listener, app, handle.clone())
    .with_graceful_shutdown(signal)
    .await;
```

## After

```rust
#[derive(Clone)]                                                                                                                                                                                      
struct TelemetryExecutor(TelemetryHandle);                       
                                            
impl axum::serve::Executor for TelemetryExecutor {
    fn execute<Fut>(&self, fut: Fut)
    where                                                                                                                                                                                             
        Fut: Future<Output = ()> + Send + 'static,
    {                                                                                                                                                                                                 
        self.0.spawn(fut);                                       
    }
}                                                                                                                                                                                                     
 
axum::serve(listener, app)                                                                                                                                                                            
    .with_executor(TelemetryExecutor(handle))                    
    .with_graceful_shutdown(signal)
    .await;
```


examples/metrics-service/src/main.rs
```diff
@@ -253,10 +252,11 @@ fn main() -> std::io::Result<()> {
             }
         });
 
-        axum_traced::serve(listener, app.into_make_service(), handle.clone())
+        axum::serve(listener, app.into_make_service())
+            .with_executor(handle.clone())
             .with_graceful_shutdown(async move { shutdown.cancelled().await })
-            .await
```
jv1i, line 258: this unwrap diff is unrelated, unwrap no longer needed due to API change upstream

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -216,6 +216,18 @@ impl TelemetryHandle {
     }
 }
 
+/// Implement `axum::serve::Executor` for `TelemetryHandle` to allow it to be
+/// used as an executor for axum's serve.
+#[cfg(feature = "axum")]
+impl axum::serve::Executor for TelemetryHandle {
```
jv1i: Alternatively we could newtype this or have a struct that has the handle as field and impl `axum::serve::Executor` for it. This seemed the simplest for users.

rcoh: I'd probably rather have this as the 10 lines of code we need in application crates rather than putting it in core for now.

e.g. if we have multiple versions of axum etc. etc. this just becomes kind a headache.

jv1i: Good idea, done

jlizen:
Is this still needed given https://github.com/tokio-rs/axum/pull/3704 ? And should we update our example to use that new feature?

I guess we are still waiting on the axum release?

jv1i:
> Is this still needed given https://github.com/tokio-rs/axum/pull/3704 ? And should we update our example to use that new feature?
> 
> I guess we are still waiting on the axum release?

@jlizen yup this PR already updates the example to pass in an executor with our telemetry into axum, I haven't opened the PR for merge until axum lands a release with this. (sorry I forgot to share update with a comment here)

PR #155, opened 2026-04-02
Fix WriteBackProcessor busy-loop re-processing already gzipped segments (#154)

### LINKED ISSUE #154: WriteBackProcessor busy-loops re-processing already gzipped segments
## Symptom

When using the `WriteBackProcessor` pipeline (non-S3 path), the background worker busy-loops re-processing already-gzipped segments, causing:

1. **"invalid trace header" warnings** every poll cycle for every processed segment:
   ```
   WARN dial9_tokio_telemetry::background_task::sealed
     failed to parse segment timestamp, falling back to mtime
       path: /data/dial9-traces/pp-16/trace.0.bin
       error: invalid trace header
   ```

2. **Busy loop with continuous disk I/O** — the worker never sleeps because processed segments are always rediscovered.

## Root cause

`WriteBackProcessor` writes gzip-compressed bytes back to the **original `.bin` path** (`mod.rs:399`). On the next poll:

1. `find_sealed_segments` (`sealed.rs:113`) scans for `.bin` files and rediscovers all processed segments.
2. `creation_epoch_secs` (`sealed.rs:18`) calls `Decoder::new`, which checks for `TRC\0` magic (`codec.rs:153`). The file now starts with gzip magic `0x1f 0x8b` → `InvalidHeader` → **warning emitted**.
3. `SymbolizeProcessor` detects gzip, skips (`mod.rs:343`). `GzipCompressor` detects gzip, skips (`mod.rs:283`). `WriteBackProcessor` rewrites the identical bytes to the same path (`mod.rs:399`).
4. `process_open_segments` returns `found = true` (`mod.rs:475`) because segments were discovered, so the `run()` loop at `mod.rs:457` **skips the `poll_interval` sleep** and immediately re-polls — creating a busy loop.

### Why only the non-S3 path

The S3 uploader **deletes** segments after upload (`s3.rs:236`), so they are never re-discovered. `WriteBackProcessor` has no equivalent cleanup.

## Impact

- **CPU/IO waste**: Continuous busy-loop reading and rewriting identical gzipped bytes with zero useful work.
- **Log noise**: WARN-level log per segment per loop iteration (unbounded rate).
- **Functional**: None — pipeline gracefully handles gzipped input, and mtime fallback provides a usable timestamp.

## Proposed fix: rename to `.bin.gz` after processing

After `WriteBackProcessor` writes the gzipped bytes, rename the file from `trace.{index}.bin` to `trace.{index}.bin.gz`. This is the cleanest solution because:

1. **Extension scanning is the existing boundary** — `find_sealed_segments` already filters by extension (`sealed.rs:118`), so `.bin.gz` files are naturally excluded from reprocessing.
2. **Matches S3 convention** — the S3 uploader already uses `.bin.gz` keys for gzipped data (`s3.rs:131`).
3. **No trace format changes** — the header is fixed at 5 bytes (magic + version, `codec.rs:15-17`) with no reserved space for flags. Expanding it would be a breaking format change affecting all decoders.
4. **TraceReader is extension-agnostic** — `analysis.rs:134` uses gzip magic-byte detection, not file extension, so it can read `.bin.gz` files without changes.

### Alternative considered: header flag

Storing a "processed" flag in the trace file header was considered but rejected because:
- The header (`codec.rs:15-17`) is exactly 5 bytes (`TRC\0` + version) with zero reserved space.
- Expanding it requires a format version bump and updating all decoders.
- A flag in the uncompressed header would not survive gzip compression (the header is inside the compressed payload), so it would need to be a wrapper — adding complexity for no benefit over a simple rename.

### Changes needed

1. **`WriteBackProcessor`** (`mod.rs:387-415`): after `fs::write`, rename the file from `.bin` to `.bin.gz`.
2. **`find_sealed_segments`** (`sealed.rs:113-133`): optionally extend to also discover `.bin.gz` files if downstream consumers need to find processed segments (e.g., for cleanup).
3. **`parse_segment_index`** (`sealed.rs:135-141`): handle `.bin.gz` suffix.

tillrohrmann:
The background worker re-discovered already-processed segments on every
poll cycle because WriteBackProcessor wrote gzip bytes back to the same
`.bin` path.  Since `find_sealed_segments` always found them again and
the processing loop skipped the sleep when segments were present, this
created a busy loop of continuous disk reads/writes with zero useful work.

Fix: GzipCompressor now sets a `write_back_extension` metadata key and
WriteBackProcessor writes to the extended path (e.g. `.bin.gz`) then
deletes the original `.bin`.  The writer's eviction also scans for
renamed variants when the original `.bin` is already gone, so processed
files respect the user's disk budget.

This fixes #154.

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/writer.rs
+++ b/dial9-tokio-telemetry/src/telemetry/writer.rs
@@ -1129,10 +1129,10 @@ mod tests {
         let max_total_size = max_file_size * 100;
         let mut writer = RotatingWriter::new(&base, max_file_size, max_total_size).unwrap();
 
-        // Write two events: the first fills segment 0, the second triggers
+        // Write two batches: the first fills segment 0, the second triggers
         // rotation (sealing segment 0 as trace.0.bin) and starts segment 1.
-        writer.write_event(&park_event()).unwrap();
-        writer.write_event(&park_event()).unwrap();
+        writer.write_encoded_batch(&test_batch()).unwrap();
+        writer.write_encoded_batch(&test_batch()).unwrap();
         // Segment 0 is now sealed as trace.0.bin.
 
         // Simulate the background worker renaming trace.0.bin → trace.0.bin.gz.
@@ -1145,7 +1145,7 @@ mod tests {
         // segment 0 (which has been renamed to .bin.gz on disk).
         writer.max_total_size = max_file_size;
         for _ in 0..3 {
-            writer.write_event(&park_event()).unwrap();
+            writer.write_encoded_batch(&test_batch()).unwrap();
         }
         writer.finalize().unwrap();
 


rcoh:
Nice! good catch and I like this change. Can you also add an end-to-end integration style test that validates that we are not leaking gziped files? (code works fine as far as I can tell, but I'd like a solid long term test since leaking gzipped files is obviously a major problem in prod)

tillrohrmann:
> Nice! good catch and I like this change. Can you also add an end-to-end integration style test that validates that we are not leaking gziped files? (code works fine as far as I can tell, but I'd like a solid long term test since leaking gzipped files is obviously a major problem in prod)

Yes, will look into adding it.

tillrohrmann:
@rcoh I've added the e2e test and pushed a small fix for handling concurrent evictions while the `WriteBackProcessor` creates the `bin.gz` file.

PR #174, opened 2026-04-09
fix: make perf-self-profile compile on macOS

### LINKED ISSUE #14: CPU profiling should still compile on macOS
Right now, the CPU profiling crate will only compile on linux

yulnr:
Closes #14

### Summary

Makes `dial9-perf-self-profile` compile on platforms other than Linux by re-organizing some of the implementation through splitting between platform specific and platform-agnostic code (and gating the linux-specific).

**Note:** I think that this approach could be a good idea long-term, it prevents lots of individual cfg gates everywhere, this might make it easier to maintain/extend without having to manually add cfgs frequently, and it should make adding support for other platforms later easier. That said it comes at the cost of some bigger diffs now, so I'm happy to switch to a smaller set of changes and more individual gating if that's preferred. 
 
###  Breakdown
Moves Linux-specific code behind `sys/linux/`. On non-Linux, `PerfSampler` constructors return `io::ErrorKind::Unsupported` and everything else is a no-op.

Shared types (`Sample`, `SymbolInfo`, `SymbolTableEntry`, etc.) stay in their original files. The Linux impl. (`PerfSampler`, ringbuffer, `resolve_symbol`, offline symbolization) moves into`sys/linux/`.

In case it makes it easier to review, it now looks like this:
```
perf-self-profile/src/
  lib.rs
  sampler.rs                     # shared types: EventSource, SamplerConfig, Sample
  symbolize.rs                   # shared types: MapsEntry, SymbolInfo, CodeInfo
  offline_symbolize.rs           # SymbolTableEntry + cross-platform entry point
  sys.rs                         # cfg dispatch
  sys/
    linux.rs                     # mod declarations, USER_ADDR_LIMIT
    linux/
      sampler.rs                 # PerfSampler, perf_event_open, ring buffer consumer
      ring_buffer.rs            # moved from src/
      symbolize.rs
      offline_symbolize.rs
    unsupported.rs               # stubs returning Unsupported
```

While at it, fixed a few things that caused warnings on macOS: the `blocking_pool_tracking` example was missing `required-features`, some items in events.rs/event_writer.rs needed cpu-profiling feature gating, and a couple test files were pulling in `mod common` unconditionally but they only use it behind feature gates.


dial9-tokio-telemetry/tests/cpu_sample_clock_alignment.rs
```diff
@@ -12,6 +12,7 @@
 //! by finding the `PollStart` event whose time interval (PollStart →
 //! corresponding PollEnd) contains the burn window.
 
+#[cfg(feature = "cpu-profiling")]
```
rcoh: entire test could be `![cfg(...)]` gated

perf-self-profile/src/sys/unsupported.rs
```diff
@@ -0,0 +1,82 @@
...
+        Vec::new()
+    }
+
+    pub fn fds(&self) -> Vec<i32> {
+        Vec::new()
+    }
+
+    pub fn active_rings(&self) -> usize {
+        0
+    }
```
rcoh: maybe these shouldn't be functions to begin with? But these methods seem unlikely to ever be useful off of linux?

perf-self-profile/src/sys.rs
```diff
@@ -0,0 +1,9 @@
+#[cfg(target_os = "linux")]
+mod linux;
+#[cfg(target_os = "linux")]
+pub use linux::*;
```
rcoh: lets avoid glob-re-exports

perf-self-profile/src/sampler.rs
rcoh, line 7: we can decide how we want to handle this but these event sources are linux only. don't need to do it now probably just make a todo. we could cfg gate individual variants potentially

rcoh (approved):
LGTM. Do you mind adding something to the github actions matrix that verifies these changes to make sure we don't regress?

perf-self-profile/src/sampler.rs
jv1i: ok, added a comment for now

perf-self-profile/src/sys/unsupported.rs
jv1i: agree, actually some of these were leftovers from my moving around of things, and two of these were already dead code before my changes. removed now.

rcoh (approved):

PR #179, opened 2026-04-10
feat: add time-based rotation to RotatingWriter (#136)

### LINKED ISSUE #136: rotating writer should also rotate based on time
right now, rotating writer will rotate on space but we should also have it rotate based on time

rcoh:
RotatingWriter now rotates segments based on wall-clock time in addition to file size. Rotation triggers when *either* condition is met:

- File exceeds `max_file_size` bytes (existing behavior)
- A wall-clock-aligned time boundary is crossed (new)

The default rotation period is 60 seconds, aligned to round minute boundaries (e.g. if the writer starts at 14:03:22, the first rotation fires at 14:04:00). This is configurable via the builder:

```rust
RotatingWriter::builder()
    .rotation_period(Duration::from_secs(300))  // every 5 min
```

Time-aligned segments produce clean S3 key paths when using the `worker-s3` feature (keys already use `{YYYY-MM-DD}/{HHMM}` bucketing).

Edge cases:
- Time rotation is skipped when no real events have been written (avoids empty sealed segments)
- `single_file()` disables time rotation (Duration::MAX)
- Eviction budget is respected with frequent time rotations

Closes #136


dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -76,17 +79,30 @@ impl TraceWriter for NullWriter {
...
 ///
 /// Files are named `{base_path}.0.bin`, `{base_path}.1.bin`, etc.
 /// Each file is a self-contained trace with its own header.
 pub struct RotatingWriter {
     base_path: PathBuf,
     max_file_size: u64,
     max_total_size: u64,
+    /// How often to rotate based on wall-clock time. `Duration::MAX` disables
+    /// time-based rotation (used by `single_file()`).
+    rotation_period: Duration,
```
jlizen, line 103: We should reject 0 seconds. Currently it rotates every batch.

jlizen (approved):

dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -442,6 +512,13 @@ impl TraceWriter for RotatingWriter {
             return Ok(());
         };
         if batch.event_count > 0 {
+            // If the time boundary expired while the segment was empty,
+            // advance it so the incoming event starts a fresh window rather
+            // than being immediately rotated out as a single-event segment.
+            if !self.has_real_events && time_source().system_time() >= self.next_rotation_time {
+                self.next_rotation_time =
+                    Self::next_boundary(time_source().system_time().as_std(), self.rotation_period);
+            }
```
jv1i: very nit: these two system_time calls should logically be the same one, could capture once:
```suggestion
            let now = time_source().system_time();
            if !self.has_real_events && now >= self.next_rotation_time {
                self.next_rotation_time =
                    Self::next_boundary(now.as_std(), self.rotation_period);
            }
```
don't think it matters much in practice in this case though, just how it reads

yulnr (approved):
LGTM

PR #183, opened 2026-04-10
Enable perf_event_open & kallsyms for E2E CI

Fluzko:
Fixes malformed demo trace generation, without CPU samples due to `kernel.perf_event_paranoid` & `kernel.kptr_restrict` not being set


rcoh (approved):
lgtm. at some point we should figure out if there is a way to centralize our ci config

PR #186, opened 2026-04-10
Intrusively Drain ThreadLocal Buffers on a Schedule

rcoh:
The existing design where the TL buffers are flushed when they reach a fixed size has two main issues:
1. Threads that are fully hung will not publish events
2. Threads that are not publishing events quickly will end up with their events in another file that gets published much later when the buffer finally fills up.

This implements a fairly minimal version where the flush thread maintains weak handles to each TL buffer. On a schedule (~30 seconds) it locks and drains the buffer. There are two optimizations to ensure that this Mutex is extremely low contention:

1. The epoch counter is bumped 5ms before so if any events are actually flushed during that 5ms window, _that_ will trigger the flush.
2. The flush thread does not drain any threads that having matching epoch counters.

Performance overhead seems to be minimal:
```
=== Comparison ===
Baseline:   289811 req/s, p50=172.5µs, p99=278.5µs
Telemetry:  274387 req/s, p50=180.2µs, p99=320.5µs
Overhead:   5.3%

=== Validation ===
✅ PASS: Telemetry overhead acceptable (5.3%)
```


dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -782,6 +786,19 @@ impl TracedRuntimeBuilder<HasTracePath> {
...
                         cycle_count += 1;
                         let drain_self = exit || cycle_count.is_multiple_of(SELF_DRAIN_INTERVAL);
+                        // Periodically drain all thread-local buffers so idle/silent
+                        // threads don't hold events across trace file rotations.
+                        // We bump the epoch one tick early (~5 ms grace period) so
+                        // busy threads self-flush on their next record_event,
+                        // and the intrusive drain only locks truly idle buffers.
+                        // On exit we bump + drain in the same tick since there is
+                        // no next tick for the grace period.
+                        if exit || cycle_count.is_multiple_of(TL_DRAIN_INTERVAL.saturating_sub(1)) {
```
jlizen:
```suggestion
                       if exit || (cycle_count + 1).is_multiple_of(TL_DRAIN_INTERVAL) {
                            shared.bump_drain_epoch();
                        }
```

current implementation slowly moves out of sync between the bump and the flush

dial9-tokio-telemetry/src/telemetry/recorder/shared_state.rs
```diff
@@ -83,6 +94,175 @@ impl SharedState {
...
+                            tracing::error!(
+                                "dial9: thread-local buffer mutex poisoned in drain_all_tl_buffers; skipping flush"
+                            );
+                        });
+                        continue;
+                    }
+                };
+                if buf.has_pending_events() {
+                    self.collector.accept_flush(buf.flush());
+                }
```
jlizen, line 156:
```suggestion
                }
                // Stamp so we skip this buffer next cycle if it stays idle.
                handle.flush_epoch.store(epoch);
```

dial9-tokio-telemetry/src/telemetry/buffer.rs
```diff
@@ -296,4 +376,57 @@ mod tests {
...
+        let epoch_clone = epoch.clone();
+        let handle = std::thread::spawn(move || {
+            epoch_clone.store(7);
+        });
+        handle.join().unwrap();
+        assert_eq!(epoch.load(), 7);
+    }
+
+    #[test]
+    fn test_flush_epoch_stamped_on_self_flush() {
```
jlizen, line 400: minor: would be also good to test something like `record event, bump epoch, record another
event, verify collector has a batch`

jlizen (changes requested):

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
rcoh: Oh good catch. I stared at this code but clearly not long enough

jlizen (approved):

PR #195, opened 2026-04-13
Round of viewer improvements

rcoh:
1. `index.html` is now the browse page. It includes a "load demo trace button" and `?trace` will redirect to the viewer so all the existing flows work.
2. The browse page now starts from a time picker then lets you select by time which then shows you available hosts

 Screenshot 2026-04-13 at 7 54 14 PM
Screenshot 2026-04-13 at 7 55 04 PM


jlizen (approved):
LGTM.

From offline, the length limit is being removed because we have users with 100MB + trace files

PR #196, opened 2026-04-14
draft / RFC: Add support for application defined events

rcoh:
The main challenge is access to string interning which I solved by introducing `Encodable`. A side benefit is that `Encodeable` is dyn Safe which is convenient.

### DIFF
--- a/dial9-tokio-telemetry/README.md
+++ b/dial9-tokio-telemetry/README.md
@@ -107,6 +107,39 @@ runtime.block_on(async {
 
 For frameworks like Axum where you don't control the spawn call, you need to wrap the accept loop. See [`examples/metrics-service/src/axum_traced.rs`](/examples/metrics-service/src/axum_traced.rs) for a working example that wraps both the accept loop and per-connection futures.
 
+## Custom events
+
+You can emit your own application-level events into the trace alongside the built-in runtime events. Define a struct with `#[derive(TraceEvent)]` and call `record_event`:
+
+```rust,no_run
+# fn main() {
+use dial9_trace_format::TraceEvent;
+use dial9_tokio_telemetry::telemetry::{record_event, clock_monotonic_ns, TelemetryHandle};
+
+#[derive(TraceEvent)]
+struct RequestCompleted {
+    #[traceevent(timestamp)]
+    timestamp_ns: u64,
+    status_code: u32,
+    latency_us: u64,
+}
+
+# let handle: TelemetryHandle = todo!();
+record_event(
+    RequestCompleted {
+        timestamp_ns: clock_monotonic_ns(),
+        status_code: 200,
+        latency_us: 1500,
+    },
+    &handle,
+);
+# }
+```
+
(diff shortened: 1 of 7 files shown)


jlizen (approved):
This seems reasonable to me on first read-through. I say, go ahead and implement it, and we can iterate on first usage if it isn't quite right.

dial9-tokio-telemetry/tests/end_to_end.rs
```diff
@@ -143,3 +143,56 @@ fn task_terminate_events_are_captured() {
         "expected at least {N} TaskTerminate events, got {terminate_count}"
     );
 }
+
```
jlizen, line 146: We probably should have a test for the advanced Encodable path (manual impl with string interning) in the test suite.


Also no test for the edge case where encode is called multiple times within a single Encodable::encode impl

dial9-tokio-telemetry/src/telemetry/buffer.rs
```diff
@@ -17,89 +17,129 @@ use std::sync::atomic::{AtomicU64, Ordering};
...
+/// In the example above, the trace will contain events named `"HttpRequestWire"`,
+/// not `"HttpRequest"`.
+pub trait Encodable {
+    /// Encode this event into the thread-local trace buffer.
+    ///
+    /// Implementations should call [`ThreadLocalEncoder::encode`] exactly once.
+    /// Each `encode` call is counted as one event for buffer flush decisions;
+    /// calling `encode` multiple times will produce multiple wire events but
+    /// only one event will be counted.
+    fn encode(&self, encoder: &mut ThreadLocalEncoder<'_>);
```
jlizen, line 122: This means that `event_count` will no longer match what is sent on the wire, do we care?

dial9-tokio-telemetry/src/telemetry/buffer.rs
```diff
@@ -17,89 +17,129 @@ use std::sync::atomic::{AtomicU64, Ordering};
...
+///             method,
+///             status: self.status,
+///         });
+///     }
+/// }
+/// ```
+///
+/// # Wire event naming
+///
+/// The event name in the trace comes from the struct passed to
```
jlizen, line 111: This might be a bit surprising, would it be worth supporting custom name method? I'm fine to leave it as-is for now though and see if anybody asks.

dial9-tokio-telemetry/README.md
```diff
@@ -107,6 +107,39 @@ runtime.block_on(async {
...
+        latency_us: 1500,
+    },
+    &handle,
+);
+# }
+```
+
+For events with repeated string values (HTTP methods, endpoint paths, etc.), implement `Encodable` manually to use string interning — see [`examples/custom_events.rs`](/dial9-tokio-telemetry/examples/custom_events.rs) for a complete example showing both patterns.
+
+Custom events are encoded into the same thread-local buffer as built-in events (~100–200 ns per call) and appear in the trace viewer alongside poll/park/wake events.
```
jlizen, line 141: If we are going to claim this, should we have benches for the custom events?

jlizen (approved):

PR #197, opened 2026-04-14
fix overhead_bench telemetry and baseline p99

fmzbl:
`overhead::telemetry::p99_lat_ns` was being faster than  `overhead::baseline::p99_lat_ns`, which shouldnt really happen.

implements:
doubled the default measurement time to tighten the estimate. 
added warmup before measurement. 
fixed underflow error.

tested on fork

### DIFF
--- a/dial9-tokio-telemetry/benches/overhead_bench.rs
+++ b/dial9-tokio-telemetry/benches/overhead_bench.rs
@@ -253,6 +253,9 @@ fn main() {
     if is_bmf {
         let mut report = bmf::Report::new();
         let mut results = std::collections::HashMap::new();
+        // Discarded warmup run so the first measured mode doesn't pay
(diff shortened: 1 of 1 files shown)


yulnr (approved):

yulnr:
After approving I realized this fixes the issue going forward, but I think it doesn't fix the poisoned/broken value from previous runs right? `telemetry_p99_added_latency_ns` in the report from bencher above is ~23 quadrillion µs
So we'd need to either delete the previous values on bencher (I think via API), or discard this metric (by not using the same name) and using a new metric name. We're doing this now on #199 tracking p90, so that should be fine, or perhaps it could keep tracking p99 but renamed.

Also even if these are separate issues feel free to combine these type of changes on a single PR next time, might make your life a bit easier for testing results :)

PR #199, opened 2026-04-14
Switch noisy p99 metrics in bencher to p90

fmzbl:
switches noisy p99 metrics to p90
fixes underflow issue.

### DIFF
--- a/dial9-tokio-telemetry/benches/overhead_bench.rs
+++ b/dial9-tokio-telemetry/benches/overhead_bench.rs
@@ -272,11 +272,11 @@ fn main() {
             );
             results.insert(mode, r);
         }
-        let baseline_p99 = results["baseline"].hist.value_at_percentile(99.0);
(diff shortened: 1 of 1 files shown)


yulnr:
I'm a bit out of the loop, so commenting from the side.

I looked at the bencher output for these last two PRs and it seems the p99 noise is mostly from baseline mode (~25% between runs vs <1.5% for telemetry/noop), that matches the cold start bias since baseline runs first. 
Since #197 should fix that with warmup, perhaps p99 behaves better after #197 lands? Or perhaps you tested both changes (this PR on top of 197) and it was still noisy?

The upside here is that the change to p90 is giving you a rename so we don't compare with the previous broken values of p99. If we were to keep p99 we'd have to rename it e.g. `telemetry_p99_overhead_ns` (or delete the previous run values from bencher).

fmzbl:
ready to merge after #197

rcoh (approved):

PR #201, opened 2026-04-14
Viewer Optimizations

rcoh:
See individual commits. With this, we can load traces with 8M events without UI issues. I also added a loading spinner

### DIFF
--- a/dial9-viewer/ui/trace_analysis.js
+++ b/dial9-viewer/ui/trace_analysis.js
@@ -90,7 +90,6 @@
           e.eventType === EVENT_TYPES.WorkerPark ||
           e.eventType === EVENT_TYPES.WorkerUnpark
         ) {
-          workerQueueSamples[w] ??= [];
           workerQueueSamples[w].push({ t: e.timestamp, local: e.localQueue });
           if (e.localQueue > maxLocalQueue) maxLocalQueue = e.localQueue;
         }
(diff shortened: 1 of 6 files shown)


rcoh:
benchmark is flaky

dial9-viewer/ui/trace_parser.js
```diff
@@ -150,13 +153,27 @@
...
     ]);
 
-    for (const frame of frames) {
+    let lastYieldPos = 0;
+    let frame;
+    while ((frame = dec.nextFrame()) !== null) {
+      // Yield to browser periodically so spinner can update
+      if (onProgress && dec._pos - lastYieldPos >= YIELD_BYTES) {
+        lastYieldPos = dec._pos;
+        onProgress({ bytesRead: dec._pos, totalBytes, eventCount: events.length });
```
jlizen: nit: this relies on an internal property of the trace decoder which might break, would be nice to have less coupling and keep it encapsulated

dial9-viewer/ui/trace_analysis.js
```diff
@@ -35,31 +39,62 @@
...
 
     for (const [w, wEvents] of Object.entries(perWorker)) {
       for (const e of wEvents) {
+        // Extract local queue samples inline
+        if (
+          e.eventType === EVENT_TYPES.PollStart ||
+          e.eventType === EVENT_TYPES.WorkerPark ||
+          e.eventType === EVENT_TYPES.WorkerUnpark
+        ) {
+          workerQueueSamples[w] ??= [];
```
jlizen: nit: this is dead code, it is already initialized above

dial9-viewer/ui/viewer.html
```diff
@@ -917,6 +978,41 @@ <h3>⌨ Keyboard</h3>
             // ── Rendering ──
             let mouseNs = null; // Track mouse position for crosshair
             let queueChartRafId = null; // Throttle queue chart redraws
+
+            /**
+             * Binary search for the span containing a given timestamp.
+             * Spans must be non-overlapping and sorted by start time.
+             */
+            function findSpanAt(spans, ns) {
```
jlizen, line 1037: minor: Lot of repetition of this binary search code, worth extracting `findNearest` and `findFirstGreaterEquals`?

dial9-viewer/ui/viewer.html
```diff
@@ -931,10 +1027,23 @@ <h3>⌨ Keyboard</h3>
...
-                        }
+                    // Binary search for visible range
+                    let lo = 0, hi = samples.length - 1;
+                    while (lo <= hi) {
+                        const mid = (lo + hi) >> 1;
+                        if (samples[mid].t < viewStart) lo = mid + 1;
+                        else hi = mid - 1;
+                    }
+                    const start = Math.max(0, lo - 1);
+                    lo = 0; hi = samples.length - 1;
```
jlizen: nit: this variable shadowing is a bit odd, maybe lo2 and hi2 here?

dial9-viewer/ui/viewer.html
```diff
@@ -664,20 +681,28 @@ <h3>⌨ Keyboard</h3>
                 });
 
             function loadFile(file) {
+                showLoadingState(`Loading ${file.name}…`);
                 const reader = new FileReader();
                 reader.onload = async (e) => {
                     try {
-                        trace = await parseTrace(e.target.result);
+                        const opts = getParseOptions();
```
jlizen, line 688: minor: it seems fragile to only parse options on reload... today it works, but people might optimize to remove reloads in future. Why not just parse url params at call time, it is cheap?

```
 function getParseOptions() {
    const params = new URLSearchParams(window.location.search);
    const opts = {};
    const s = params.get('start');
    const e = params.get('end');
    if (s != null) opts.startTime = Number(s);
    if (e != null) opts.endTime = Number(e);
    return opts;
  }
```

jlizen (approved):
Good improvements, some minor things, up to you.

non-blocking: rather than maintaining `extractLocalQueueSamples` and `indexWakeEvents` for backwards compat with tests, can they internally call `buildWorkerSpans`, or just have tests call `buildWorkerSpans` directly? Seems like a maintenance future footgun.

PR #209, opened 2026-04-14
Rate-limit all tracing/logging to prevent log spam

### LINKED ISSUE #139: change all logging to be rate limited
there's never a good reason for dial9 to spam tracing events

example impl: https://github.com/awslabs/metrique/blob/b7201ba7f4eeb9abe09ce78ac8f55995a4b1f37f/metrique-writer-format-emf/src/rate_limit.rs

rcoh:
## Summary

Wraps all `warn!`/`error!`/`info!` tracing calls that could fire repeatedly in the `rate_limited!` macro (at most once per 60s per call site). This prevents dial9 from spamming application logs when error conditions persist.

### What changed

- **background_task/mod.rs**: Rate-limited warnings for segment read failures, processor failures, upload failures, symbolization failures, eviction failures, and the `uploaded` info log (10s interval)
- **telemetry/recorder/mod.rs**: Rate-limited warnings for dropped batches, transcode failures, flush failures, finalize failures, and CPU profiler startup failures
- **telemetry/writer.rs**: Rate-limited eviction warnings, "writer already closed" warning, and dropped-events info in Drop
- **telemetry/buffer.rs**: Rate-limited the "dropping unflushed events" warning in ThreadLocalBuffer Drop
- **background_task/sealed.rs**: Rate-limited the "failed to parse segment timestamp" warning
- **rate_limit.rs**: Fixed `Duration::MAX` to use fully qualified `std::time::Duration::MAX` so the macro works at call sites without `Duration` in scope

### What was NOT rate-limited (one-time lifecycle events)

- Worker started/stopped
- Stop signal received, draining / drain complete / drain timed out
- Resolved bucket region
- Graceful shutdown debug logs
- Worker thread panicked during shutdown

### Testing

- `cargo fmt --check` ✅
- `cargo clippy --all-targets --all-features` ✅
- `cargo nextest run` — 356 tests pass ✅
- `cargo nextest run --stress-duration 20s` — 4 iterations, all pass ✅

Closes #139

### DIFF
--- a/dial9-tokio-telemetry/src/background_task/mod.rs
+++ b/dial9-tokio-telemetry/src/background_task/mod.rs
@@ -7,6 +7,7 @@ pub mod s3;
 pub(crate) mod sealed;
 
 use crate::metrics::{Operation, SegmentProcessMetrics, SegmentProcessMetricsGuard};
+use crate::rate_limit::rate_limited;
 use metrique::timers::Timer;
 use metrique_writer::BoxEntrySink;
 use pipeline_metrics::{MetriqueResult, PipelineMetrics, StageMetrics};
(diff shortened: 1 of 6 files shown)


jlizen (approved):
Has the time come to expose this in a dial9-util crate? We've reimplemented it so many times...

PR #210, opened 2026-04-15
Fix: add explicit monotonic <-> realtime clock sync to trace format

### LINKED ISSUE #202: timestamps stored are CLOCK_MONOTONIC instead of CLOCK_REALTIME so you cannot recover the true event time
CLOCK_MONOTONIC is aligned to boot so it doesn't provide a clean way to re-align to wall clock

yulnr:
closes #202

## Summary

Event timestamps are `CLOCK_MONOTONIC` nanoseconds since boot, but the viewer was feeding them directly to `new Date()`, so absolute-time mode rendered 1970-ish timestamps (e.g. ~00:00 in the viewer).

This change records explicit monotonic-to-realtime anchors in the trace and uses them in the viewer to recover wall clock correctly. New traces now show accurate absolute time without changing event timestamp semantics. Older traces work through a compatibility fallback. 

I went for this approach rather than directly switching event timestamps to `CLOCK_REALTIME`, because as I understand it clock jumps/slew (NTP), suspend/resume, or manual clock edits could break duration math.

## Details

Adds `ClockSyncEvent { timestamp_ns, realtime_ns }` wire type. Both values come from `clock_pair()`, which reads monotonic -> realtime -> monotonic back-to-back and pairs realtime with the midpoint monotonic sample, so correlation error is roughly half a `clock_gettime` interval. The writer emits anchors at file open and on rotation. The viewer picks the nearest anchor for each timestamp, computes `realtime - monotonic` at that anchor, and applies that offset in `fmtWallClock`.

I kept this as a new event type instead of extending SegmentMetadataEvent for compatibility, AFAIK in our typed decode path, missing fields on older traces can make that event decode to None and get skipped.

This also changes something I found confusing: `SegmentMetadataEvent.timestamp_ns` used wall clock while being a `#[traceevent(timestamp)]` field, and other timestamp fields are monotic. This now matches the rest of events, and wall clock is explicit in `ClockSyncEvent.realtime_ns`.
For new files, `sealed.rs` reads segment creation time from `ClockSyncEvent`. For older files, both `sealed.rs` and the viewer fall back to legacy `SegmentMetadata.timestamp_ns` when it looks epoch-scale, and reconstruct wall clock as an approximation.

The behavioral break is only for any external consumers that treated `SegmentMetadataEvent.timestamp_ns` as wall clock. If any those consumers should migrate to `ClockSyncEvent.realtime_ns` (or apply the same `LEGACY_EPOCH_NS_FLOOR` fallback for old traces).

If we'd rather not change `SegmentMetadataEvent.timestamp_ns`, I'm happy to revert but perhaps semantics should change in some other way.

## Future improvements

If needed, we could add interpolation between neighboring anchors (instead of nearest-anchor offset). Perhaps that could help on very long traces, or something like during gradual clock drift?

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/events.rs
+++ b/dial9-tokio-telemetry/src/telemetry/events.rs
@@ -188,6 +188,16 @@ pub enum TelemetryEvent {
         /// Key-value metadata pairs.
         entries: Vec<(String, String)>,
     },
+    /// Clock-correlation anchor pairing a monotonic timestamp with the
+    /// wall-clock value captured at the same instant.
+    ClockSync {
+        /// Monotonic nanoseconds.
+        #[serde(rename = "timestamp_ns")]
+        timestamp_nanos: u64,
+        /// Nanoseconds since the Unix epoch.
+        #[serde(rename = "realtime_ns")]
+        realtime_nanos: u64,
+    },
 }
 
 impl TelemetryEvent {
@@ -224,6 +234,9 @@ impl TelemetryEvent {
             TelemetryEvent::ThreadNameDef { .. } => None,
             TelemetryEvent::SegmentMetadata {
                 timestamp_nanos, ..
+            }
+            | TelemetryEvent::ClockSync {
+                timestamp_nanos, ..
             } => Some(*timestamp_nanos),
         }
     }
@@ -241,7 +254,8 @@ impl TelemetryEvent {
             | TelemetryEvent::TaskTerminate { .. }
             | TelemetryEvent::ThreadNameDef { .. }
             | TelemetryEvent::WakeEvent { .. }
-            | TelemetryEvent::SegmentMetadata { .. } => None,
+            | TelemetryEvent::SegmentMetadata { .. }
+            | TelemetryEvent::ClockSync { .. } => None,
         }
     }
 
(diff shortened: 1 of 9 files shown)


dial9-tokio-telemetry/src/telemetry/events.rs
```diff
@@ -367,15 +381,49 @@ pub fn clock_monotonic_ns() -> u64 {
...
+    ts.tv_sec as u64 * 1_000_000_000 + ts.tv_nsec as u64
+}
+
+#[cfg(not(target_os = "linux"))]
+pub(crate) fn clock_realtime_ns() -> u64 {
+    use std::time::{SystemTime, UNIX_EPOCH};
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map(|d| d.as_nanos() as u64)
+        .unwrap_or(0)
```
rcoh, line 413: what would cause this? now being earlier than the unix epoch? Seems like an expect here would be safe.

dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -267,44 +268,44 @@ impl RotatingWriter {
...
         segment_metadata: &[(String, String)],
     ) -> std::io::Result<RawEncoder<BufWriter<File>>> {
         let mut encoder = Encoder::new_to(writer)?;
         let entries = segment_metadata.to_vec();
-        let timestamp_ns = time_source()
-            .system_time()
-            .duration_since(std::time::UNIX_EPOCH)
-            .unwrap_or_default()
-            .as_nanos() as u64;
         encoder.write(&SegmentMetadataEvent {
```
rcoh, line 280: should we use the same mono in SegmentMetadataEvent?

rcoh (approved):

PR #212, opened 2026-04-15
Bootstrap Dial9 with a macro

### LINKED ISSUE #158: Document that code within `block_on` does not produce events
The `block_on` task within tokio _does not_ fire poll start/stop callbacks meaning that any work running in the root of `#[tokio::main]` is effectively invisible to dial9.

Fluzko:
closes #158 

## Summary
This PR aims to do both:
- Document how to get your async code instrumented and tracked by explaining how Dial9 works & interacts with Tokio.
- Add new macro `#[dial9_tokio_telemetry::main()]` that mainly solves the user putting async code in places where it won't be tracked, and also provides a helpful interface to simplify the setup


dial9-tokio-telemetry/src/config.rs
```diff
@@ -0,0 +1,182 @@
...
+    ///
+    /// The closure receives the staged builder by value and must return it.
+    /// Use this to reach every knob on `TracedRuntimeBuilder` —
+    /// `with_runtime_name`, `with_task_tracking`, `install`,
+    /// `with_worker_poll_interval`, `with_worker_metrics_sink`, and
+    /// (feature-gated) `with_cpu_profiling`, `with_sched_events`,
+    /// `with_s3_uploader`, `with_s3_client`.
+    ///
+    /// Can be called multiple times; each call composes onto the prior state.
+    pub fn with_runtime<F>(mut self, f: F) -> Self
```
jv1i, line 138: I would perhaps not list the config of `TracedRuntimeBuilder` entirely here, to avoid maintaining that. 
We could rephrase to something like "Use this to access runtime configuration methods like ..., see [`TracedRuntimeBuilder`] for details` (linking to the type rather than re-documenting here

dial9-tokio-telemetry/README.md
```diff
@@ -24,24 +24,61 @@ Without this flag, compilation will fail with errors about missing methods on `t
 
 ## Quick start
 
+There are two ways to set up dial9: the `#[main]` macro (recommended for most apps) or manual `TracedRuntime` setup. The macro handles the boilerplate of building the runtime and spawning your code as an instrumented task. Inside the body, call `TelemetryHandle::current()`to reach a handle for wake-event tracking. Use manual setup when you need control over the runtime builder (e.g. custom thread counts, `current_thread` runtime) or need to integrate into an existing runtime setup.
```
jv1i: The last part about when to use each is probably a bit stale
> Use manual setup when you need control over the runtime builder.. 

we can control the runtime builder using the macro setup as well (incl. custom thread counts, etc). Perhaps we can phrase this highlighting common use cases, e.g. "Use manual setup when you have multiple tokio runtimes, don't own main (e.g. library code, embedded services), or need to integrate into existing runtime-building code..."

dial9-macro/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+/// async fn main() {
+///     let handle = TelemetryHandle::current();
+///     handle
+///         .spawn(async { /* instrumented sub-task */ })
+///         .await
+///         .unwrap();
+/// }
+/// ```
+#[proc_macro_attribute]
+pub fn main(attr: TokenStream, item: TokenStream) -> TokenStream {
```
jv1i, line 132: Perhaps either here or/and in the readme we should mention that this macro is a replacement for 
#[tokio::main], instead of complement. It might be obvious for some users but some might try to stack them.

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -257,6 +291,28 @@ impl std::fmt::Debug for TelemetryHandle {
...
+    /// Return the [`TelemetryHandle`] for the current thread.
+    ///
+    /// The handle is installed on every thread owned by a traced runtime
+    /// (workers and blocking threads) via the runtime's `on_thread_start`
+    /// hook, and cleared on `on_thread_stop`.
+    ///
+    /// Panics if called from a thread that is not owned by a dial9 runtime
+    /// (e.g. the thread that called `runtime.block_on(...)` on a
+    /// `current_thread` runtime, or any non-runtime thread). Use
+    /// [`TelemetryHandle::try_current`] if you need to handle that case.
```
jv1i, line 328: nit: we could explicitly say that this is fine to do inside a ` #[dial9_tokio_telemetry::main]` body

dial9-tokio-telemetry/src/lib.rs
```diff
@@ -18,3 +18,37 @@ pub(crate) mod rate_limit;
...
+/// #[dial9_tokio_telemetry::main(config = my_config)]
+/// async fn main() {
+///     let handle = TelemetryHandle::current();
+///     handle
+///         .spawn(async { /* instrumented */ })
+///         .await
+///         .unwrap();
+/// }
+/// ```
+pub use dial9_macro::main;
```
jv1i: I think we might not need to duplicate this doc here since this is just a re-export.

dial9-macro/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+            "the `async` keyword is missing from the function declaration",
+        ));
+    }
+
+    let config_fn = &args.config;
+    let attrs = &input.attrs;
+    let vis = &input.vis;
+    let name = &input.sig.ident;
+    let ret = &input.sig.output;
+    let body_stmts = &input.block.stmts;
```
jv1i, line 68: If I'm not mistaken looks like expand_main would silently drop args, where clauses and generics. We might want to either support them (e.g. someone using dial9 on a CLI that uses args) or warn/err about what we don't support, e.g.
```rust
if !input.sig.inputs.is_empty() {
    return Err(syn::Error::new_spanned(
        &input.sig.inputs,
        "#[dial9::main] does not support function arguments",
    ));
}
```

yulnr (commented):
Good work! 
As discussed earlier overall approach LGTM. Did a quick pass now mostly looking at docs/usage

yulnr:
I think we should add an example or update some of the examples in `dial9-tokio-telemetry/examples` with usage of the macro

dial9-macro/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+                .expect("failed to initialize dial9 runtime");
+            let __dial9_handle = __dial9_guard.handle();
+            __dial9_runtime.block_on(async move {
+                __dial9_handle
+                    .spawn(async move {
+                        #(#body_stmts)*
+                    })
+                    .await
+                    .expect("dial9 root task panicked")
+            })
```
jv1i: Note that here `.expect()` on the awaited `JoinHandle` double-panics, the worker already fired the real panic, then main re-panics with a `String` wrapping the `JoinError` debug. also means `catch_unwind` loses the original payload type.

So panics in the main body will output something like this:
```
thread 'tokio-rt-worker' panicked at src/main.rs:133:22:
called `Option::unwrap()` on a `None` value

thread 'main' panicked at src/main.rs:110:1:
dial9 root task panicked: JoinError::Panic(Id(12), "called `Option::unwrap()` on a `None` value", ...)
```

wdyt about matching on the result and doing `resume_unwind` on the panic arm? (to propagate the error instead)
```rust
match __dial9_handle.spawn(async move { #(#body_stmts)* }).await {
    ::std::result::Result::Ok(output) => output,
    ::std::result::Result::Err(err) if err.is_panic() => {
        ::std::panic::resume_unwind(err.into_panic())
    }
    ::std::result::Result::Err(err) => {
        ::std::panic!("dial9 root task cancelled: {err}")
    }
}
```
(or some better version of that)

dial9-macro/src/lib.rs
jlizen: I'd be fine to just error on it for now, but yeah not silently drop.

dial9-macro/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+                .build()
+                .expect("failed to initialize dial9 runtime");
+            let __dial9_handle = __dial9_guard.handle();
+            __dial9_runtime.block_on(async move {
+                __dial9_handle
+                    .spawn(async move {
+                        #(#body_stmts)*
+                    })
+                    .await
+                    .expect("dial9 root task panicked")
```
jlizen: We should have some tests probing behavior when user's body returns `Result<_, E>` the error type propagates correctly, and also that any panic inside the async task propagates context properly.

dial9-macro/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+
+struct MainArgs {
+    config: Path,
+}
+
+impl Parse for MainArgs {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let ident: syn::Ident = input.parse()?;
+        if ident != "config" {
+            return Err(syn::Error::new(ident.span(), "expected `config`"));
```
jlizen: nit: could be a bit user friendlier
```
              "missing required `config = <fn>` argument, e.g. #[dial9_tokio_telemetry::main(config = my_config)]",
```

dial9-macro/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+use syn::{ItemFn, Path, Token, parse_macro_input};
+
+struct MainArgs {
+    config: Path,
+}
+
+impl Parse for MainArgs {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let ident: syn::Ident = input.parse()?;
+        if ident != "config" {
```
jlizen, line 22: We should also validate that the config is 0-argument, otherwise you'll get a compile error at the call site that is much less clearer.

dial9-macro/src/lib.rs
jlizen: Also, if there is no argument at all (`#[dial9_tokio_telemetry::main]`) you would just get something like: `"unexpected end of input, expected identifier"`, we should add a check for empty input as well.

dial9-tokio-telemetry/tests/macro_main.rs
```diff
@@ -0,0 +1,51 @@
+use std::path::PathBuf;
+
+use dial9_tokio_telemetry::config::Dial9Config;
+
+fn tmp_base_path() -> PathBuf {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let path = dir.path().join("trace.bin");
+    std::mem::forget(dir);
```
jlizen, line 9: minor: this (and some other places) leak the `TempDir` which gets annoying over time. I tend to instead prefer returning both the handle along with the tempdir in the test. And then, you can newtype it to leak and output its path on panic so that you still have access to the tempdir for debugging.

dial9-tokio-telemetry/src/lib.rs
```diff
@@ -18,3 +18,37 @@ pub(crate) mod rate_limit;
...
+/// # Usage
+///
+/// ```rust,ignore
+/// use dial9_tokio_telemetry::{main, config::Dial9Config, telemetry::TelemetryHandle};
+///
+/// fn my_config() -> Dial9Config {
+///     Dial9Config::new("/tmp/trace.bin", 1024 * 1024, 16 * 1024 * 1024)
+/// }
+///
+/// #[dial9_tokio_telemetry::main(config = my_config)]
```
jlizen: it would be great to support `config = || {}` inline, but it's fine to be later scope. Probably a good backlog though.

dial9-macro/src/lib.rs
jlizen: It would be great to have `trybuild` compile_fail tests, but I don't think we're using it already, so it's fine to wait.

jlizen (commented):
Core impl seems solid. Agree with @yulnr 's comments. Some other minor things.

rcoh (commented):
Have not done a close look yet. I want to make sure there is a very easy way to have dial9 fully disabled by runtime configuration when dial9 main is used.

dial9-macro/src/lib.rs
Fluzko: Oh, good catch, I completely missed it

dial9-macro/src/lib.rs
Fluzko: @jlizen I've created an issue to add `trybuild`, sounds great to have it. #228 feel free to refine the description, which is pretty vague

dial9-tokio-telemetry/src/lib.rs
Fluzko: Love it, creating a follow-up for it: #229

dial9-macro/src/lib.rs
Fluzko: Thanks for the explanation, it wasn't clear on my end. Take a look at how it's implemented now. It's pretty much the same, except for the await external panic, which  I replaced with `unreachable`.  Since we're `.await`ing the task immediately, no one else has a reference to it, and `block_on` holds the task until it finishes, there's no way someone can drop or abort the handle before the `await` resolves. Correct me if I'm wrong

Fluzko:
> I think we should add an example or update some of the examples in `dial9-tokio-telemetry/examples` with usage of the macro

Added a quick example [here]([a4017fb](https://github.com/dial9-rs/dial9-tokio-telemetry/pull/212/commits/a4017fbb8fb02dde3bb1cadf92848851a662af85)), @rcoh do we want to change the rest of the examples to default the macro instead?

rcoh:
One option here is having the config function return `Option`. when it returns `None` we _fully_ do not start dial9 at all.

So you could do:
```rust
let v = std::env::var("ENABLE_DIAL9");
if !v { return None }
```

Fluzko:
> One option here is having the config function return `Option`. when it returns `None` we _fully_ do not start dial9 at all.
> 
> So you could do:
> 
> ```rust
> let v = std::env::var("ENABLE_DIAL9");
> if !v { return None }
> ```

I like the idea, but this approach introduces another problem: you are not able to configure `tokio` anymore, which is weird, since we wrap it.
So what about if we still return a `Dial9Config`, but we do something like this:
```rust
Dial9ConfigBuilder::disabled().with_tokio(|t| {...})
```
Check [this commit]([8ba8c3f](https://github.com/dial9-rs/dial9-tokio-telemetry/pull/212/commits/8ba8c3ff2d10949eacdb6ccb42f0c325e50374bb)) with the new proposal. It basically reintroduces the concept of `Dial9ConfigBuilder` and allows you to choose your path (enabled/disabled) by doing `::new()` or `::disabled()`, then you configure based on the options, and both of them converge by returning a `Dial9Config`.

CC: @yulnr @jlizen

jlizen (approved):
Changes seem good to me. I don't mind the builder API. The rationale for it seems reasonable. Does that work for you too @rcoh ?

Re: other examples:
- `simple_workload.rs` probably is worth cutting over since the macro is the "simple thing" (and also it needs fixing to use the runtime handle to spawn with anyway)
- `macro_workload.rs`, I would probably re-name to something like `conditionally_enable.rs` since that's the more interesting part of it (assuming the macro is documented in the README and the simpler example)

For the rest, I don't have a strong feeling about switching over to the macro... might as well, but then we would probably want to leave behind `multiple_runtimes.rs` or something else that shows the manual setup. Even just `manual_setup.rs` :)

rcoh:
Totally agree with all of Jess' suggestions

rcoh (commented):
The one other place I want to see cutover is the example service using the TL handle for spawning

PR #213, opened 2026-04-15
Taskdump pending

rcoh:


rcoh:
I'm going to hold on this for a bit. It means we require tokio 1.52 and it seems like 1.52 may have a regression so I want to wait until everything stabilizes.

PR #216, opened 2026-04-16
 feat!: add optional field type modifier, named field decode, and cut thread_name off sentinel value to that

jlizen:
### Summary

Two changes: a wire format extension for optional fields in `dial9-trace-format`, and converting `CpuSampleEvent.thread_name` to use it.

There is a supporting Rust decoder change to use named field lookup instead of positional indexing, similar to how the javascript viewer works. This was initially required because i wanted to add a `span_name` field to a poll event behind conditional compilation, which I scoped out of this PR. But, it seemed sensible enough in general that I left it in. Glad to rip it out if reviewers don't like it.

#### Optional field type modifier (dial9-trace-format)

With the addition of custom application events (#196), users will increasingly define event structs with optional fields (optional span context, optional request metadata, optional error details). The wire format needs first-class support for `Option<T>` rather than relying on sentinel values or conventions that each consumer must know about.

Adds a high-bit modifier (0x80) to the field type tag in the schema frame. Any field type can be made optional by setting this bit (e.g. `PooledString` tag 7 becomes `OptionalPooledString` tag 0x87). On the wire, optional fields are preceded by a 1-byte presence prefix: `0x00` = absent, `0x01` = present (followed by the inner type's normal encoding). This gives clean `Option<T>` semantics with 1 byte overhead when absent, and saves wire size compared to always encoding a sentinel value (e.g. 1 byte vs 4 bytes for an absent PooledString).

#### Positional decoding -> field name-based decoding

The derive macro's `decode` is also changed from positional indexing (`fields.get(idx)?`) to named field lookup. The decoder zips field names from the schema with decoded values, and the derive matches by name. This means:

- Fields can be `cfg`-gated on the wire struct without positional mismatches
- A reader compiled with different features than the writer handles missing/extra fields gracefully
- `Option<T>` fields that are absent from the schema (not just absent on the wire) default to `None`

Old decoders that don't recognize the 0x80-modified tag will reject the schema, which is the correct behavior since they can't interpret the data.

#### thread_name cutover (dial9-tokio-telemetry)

Converts `CpuSampleEvent.thread_name` from `InternedString` with a `"<no thread name>"` sentinel to `Option<InternedString>`. This removes the sentinel convention and uses the new optional field encoding. The viewer's sentinel check is removed accordingly.

This exercises the new code path and also saves a bit of wire space. But, it does make it a breaking change (I assume? Since viewers expecting the old format will reject the newly optional field?)

I can leave that field alone if preferred, though. #218 will set us up to better poke at the optional value usage in custom events instead.

#### Alternatives rejected

1. **Sentinel values (e.g. `""` or `"<no thread name>"`)**: Always encode optional fields as their inner type, using a sentinel for absent. Wastes wire bytes (e.g. 4 bytes for an absent PooledString vs 1 byte with the optional prefix). Consumers must check for the sentinel, which is a convention rather than a type-level guarantee. Breaks down for user-defined custom events where `Option<T>` should just work.

2. **Positional-only decode**: Field indices shift depending on which features are enabled, causing mismatches when writer and reader have different features. Named field decode eliminates this.

3. **Presence bitmask in event header**: More space-efficient for many optional fields, but couples all optional fields together and changes the event frame layout. Overkill for 1-2 optional fields where the per-field prefix byte is simpler.

4. **Trailing-only `Option<T>` fields**: Can't have None followed by Some, and inserting new optional fields between existing ones is impossible.

5. **High-bit optional without named decode**: Requires the field to always be present on the wire (1 byte when absent). Combined with named decode, the field can be `cfg`-gated out entirely for zero overhead.

### Testing

- dial9-trace-format: round-trip tests for optional field encoding/decoding (present, absent, mixed with required fields), named decode with missing fields
- Demo trace regenerated with optional thread_name encoding and CPU profile samples. Validated that the thread_name comes through properly.
- Manual end-to-end overhead measurement via `overhead_bench` noop mode: 30-second runs showed throughput and latency differences within noise (~1-2%, flipping direction between runs)


rcoh (commented):
Overall this seems good, I wonder if we should move optional into FieldType

dial9-trace-format/src/decoder.rs
```diff
@@ -59,6 +59,7 @@ pub struct RawEvent<'a, 'f> {
     pub name: &'f str,
     pub timestamp_ns: Option<u64>,
     pub fields: &'f [FieldValueRef<'a>],
+    pub field_names: &'f [String],
```
rcoh: I wonder if we could actually put the schema in this field instead of field_names + fields (ref to the schema)

rcoh (commented):
I wonder if the best long term approach is to have `RawEvent` function as the serde _format_ then allow deserializing from raw event into a Rust struct

I like this because then it allows open ended deserialization of trace files into predefined types.

Then we can have `enum TokioEvent { ... }`, `enum ProfilingEvent { ... }`, then during `for_each_event`, try from your list of known frame types and handle it

dial9-trace-format/src/decoder.rs
jlizen: It's possible but fairly complex since it runs into borrow checker issues. Both the schema and the pool are mutated during the read loop, we would need a two-pass processing to do schema/pool frames first, or some sort of refcell based solution. Better to leave til later I think, the whole decoder is due for a refactor

jlizen:
> I wonder if the best long term approach is to have RawEvent function as the serde format then allow deserializing from raw event into a Rust struct

Yeah, this seems like a nicer direction, though I don't have time to look at it now. Probably worth cutting a backlog issue to dump thoughts into.

dial9-trace-format/src/decoder.rs
```diff
@@ -391,6 +394,13 @@ impl<'a> Decoder<'a> {
         mut f: impl for<'f> FnMut(RawEvent<'a, 'f>) -> Result<(), E>,
     ) -> Result<(), TryForEachError<E>> {
         let mut values_buf: Vec<FieldValueRef<'a>> = Vec::new();
+        // Cloned schema metadata, reused across iterations to avoid
```
jlizen: As mentioned in the other comment, switching over to storing the schema directly is attractive, but fairly complex given how much woven-together immutable and mutable state we have. We can't simply split into a "mutable struct" and "immutable" struct, it would require two-pass decoding.

dial9-trace-format/src/encoder.rs
```diff
@@ -332,8 +332,18 @@ impl<W: Write> Encoder<W> {
...
+            if schema.entry.fields[i].field_type.is_optional() {
+                match v {
+                    FieldValue::None => enc.state.writer.write_all(&[0x00])?,
+                    other => {
+                        enc.state.writer.write_all(&[0x01])?;
+                        enc.write_field_value(other)?;
+                    }
+                }
+            } else {
+                enc.write_field_value(v)?;
```
rcoh: field value _is_ none right? so do we still need this branch here? or can this be  in `write_field_value`?

rcoh: ah I see, this can't be in write_field_value because the value itself doesn't have a type.

rcoh (approved):

Issue #217, opened 2026-04-16
`TelemetryEvent` has a lossy `Serialize` impl due to interned strings


jlizen:
## Problem

`TelemetryEvent` derives `Serialize`, but several fields are `InternedString` values that serialize as raw `u32` pool IDs (e.g. `PollStart.spawn_loc`). The string pool lives on `TraceReader`, not on the event, so serialized events alone cannot reconstruct the original strings.

This means `serde_json::to_string(&event)` produces lossy output that looks like `{"spawn_loc": 3}` instead of `{"spawn_loc": "src/main.rs:42"}`.

Today, consumers must resolve interned strings manually before serializing:
- Built-in events: look up `spawn_loc` in `reader.spawn_locations`
- Custom events: call `reader.resolve_custom_fields()` or `reader.resolve_interned` (added in #218 )
- `trace_to_fat_jsonl` demonstrates both patterns

This is tedious and unpleasant. If we are going to impl `Serialize`, we should have a simple way to take an event and represent it on the wire without lossiness.

## Some options

1. **Eagerly resolve during decoding**: change `to_owned_event` to resolve all `InternedString` fields to `String` so `TelemetryEvent` never contains opaque IDs. `PollStart.spawn_loc` becomes `Option<String>`, custom event fields become `FieldValue::String`. Simple, but loses the information that a value was interned (if anyone cares) and changes every variant's field types. It also forces all CLI consumers to pay this cost even if they don't want to resolve all these fields/allocate for them.

2. **Resolved wrapper type**: a `ResolvedEvent<'a>` that borrows the event and the `TraceReader`, implementing `Serialize` to produce self-contained output. No changes to `TelemetryEvent` itself, but adds API surface and requires the caller to wrap events before serializing.

3. **Include the pool in serialization**: serialize a container that includes both the events and the string pool, so consumers can reconstruct values. Preserves fidelity but makes the output format much more complex.

Option 2 is likely the best fit. It avoids forcing allocation costs on CLI analysis consumers who don't need resolved strings, keeps `TelemetryEvent` lean for the common case, and provides a clean opt-in path for serialization.

PR #218, opened 2026-04-16
feat!: surface custom events in analysis pipeline and CLI

jlizen:
### Summary

PR #196 added custom event support on the write path. However, the read/analysis side silently drops them: `decode_ref` returns `None` for unrecognized event names, so custom events never appear in `TraceReader` or CLI tools.

This change closes that gap with three pieces:

**`Custom` variant on `TelemetryEvent`.** When `TraceReader` encounters an unknown event name, it constructs `Custom { name, timestamp_nanos, fields }` pairing each field value with its schema-defined name. Custom events flow through the entire analysis pipeline without per-event-type code. To support this, `RawEvent` in `dial9-trace-format` gains `field_names` (from the schema) and `segment_index` (incremented on mid-stream header resets).

**Per-segment string pool tracking.** Custom events may contain `PooledString` (interned) fields. The existing code resolves interned strings eagerly for known fields using hand-rolled helpers, which doesn't scale to arbitrary user-defined fields. `TraceReader` now snapshots the string pool at each segment boundary and exposes two resolution APIs:

- `resolve_interned(event_index, id)`: zero-cost `&str` lookup for individual fields. Demonstrated in the `custom_events` example.
- `resolve_custom_fields(event_index, fields)`: bulk resolution replacing all `PooledString` values with `String`. Allocates a new `Vec`. Demonstrated in `trace_to_fat_jsonl`.

**Supporting changes in `dial9-trace-format`.** `FieldValueRef::to_owned()` for converting zero-copy field values to owned. `FieldValue` gains a `Serialize` impl behind the `serde` feature gate. `FieldValue` is re-exported from the crate root.

The existing `populate_spawn_loc` helper is left in place since it serves the hot path for spawn location lookups, but new code should prefer `resolve_interned`.

### Testing

- Extended the `custom_events` example to demonstrate `TraceReader` round-trip with `resolve_interned` (and double checked the output).
- Updated `trace_to_fat_jsonl` to emit custom events using `resolve_custom_fields`.
- New unit tests covering: custom events with primitive-only fields, inline `String` fields, interned string resolution across multiple segment resets (including colliding pool IDs and same pool sizes), and `resolve_custom_fields` correctness.

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/analysis.rs
+++ b/dial9-tokio-telemetry/src/telemetry/analysis.rs
@@ -1,6 +1,7 @@
 use crate::telemetry::events::{CpuSampleSource, TelemetryEvent};
 use crate::telemetry::format::{self, TelemetryEventRef, WorkerId};
 use crate::telemetry::task_metadata::TaskId;
+use dial9_trace_format::FieldValue;
 use dial9_trace_format::InternedString;
 use dial9_trace_format::decoder::{Decoder, StringPool};
 use std::cmp::Reverse;
@@ -75,6 +76,33 @@ impl TraceReader {
                     thread_names.insert(tid, name.clone());
                 }
                 events.push(owned);
+            } else {
+                // Unknown event name: capture as Custom, resolving any
+                // PooledString fields to String while the pool is available.
+                let fields = ev
+                    .field_names
+                    .iter()
+                    .zip(ev.fields.iter())
+                    .map(|(name, val)| {
+                        let owned = match val {
+                            dial9_trace_format::types::FieldValueRef::PooledString(id) => {
+                                let s = ev
+                                    .string_pool
+                                    .get(*id)
+                                    .unwrap_or("<unresolved>")
+                                    .to_string();
+                                FieldValue::String(s)
+                            }
+                            other => other.to_owned(),
+                        };
+                        (name.clone(), owned)
+                    })
+                    .collect();
+                events.push(TelemetryEvent::Custom {
+                    timestamp_nanos: ev.timestamp_ns,
+                    name: ev.name.to_string(),
+                    fields,
+                });
             }
         })
         .map_err(|e| std::io::Error::new(std::io::ErrorKind::InvalidData, e.to_string()))?;
@@ -300,7 +328,8 @@ pub fn analyze_trace(events: &[TelemetryEvent]) -> TraceAnalysis {
             | TelemetryEvent::ThreadNameDef { .. }
             | TelemetryEvent::WakeEvent { .. }
             | TelemetryEvent::SegmentMetadata { .. }
-            | TelemetryEvent::ClockSync { .. } => {}
+            | TelemetryEvent::ClockSync { .. }
+            | TelemetryEvent::Custom { .. } => {}
         }
     }
 
@@ -816,7 +845,8 @@ mod tests {
     use super::*;
     use crate::telemetry::format::WorkerId;
     use crate::telemetry::task_metadata::UNKNOWN_TASK_ID;
-    use dial9_trace_format::InternedString;
+    use dial9_trace_format::encoder::Encoder;
+    use dial9_trace_format::{FieldValue, InternedString};
     const UNKNOWN_SPAWN_LOC: InternedString = InternedString::from_raw(0);
 
     #[test]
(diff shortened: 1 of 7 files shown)


dial9-tokio-telemetry/src/telemetry/analysis.rs
```diff
@@ -101,8 +145,57 @@ impl TraceReader {
...
+    }
+
+    /// Return a copy of `fields` with every `FieldValue::PooledString`
+    /// replaced by `FieldValue::String` using the string pool that was
+    /// active at `event_index` (an index into [`all_events`](Self::all_events)).
+    ///
+    /// This allocates a new `Vec` and clones every field. For zero-cost
+    /// resolution of individual interned string fields, use
+    /// [`get_interned_string`](Self::get_interned_string) directly.
+    pub fn resolve_custom_fields(
```
jlizen: I could see an argument for making this even broader, something like, `resolve_events` that also handles all the spawn location resolution, etc. Feel free to tell me to get rid of this in favor of #217 ...

dial9-tokio-telemetry/src/telemetry/analysis.rs
```diff
@@ -101,8 +145,57 @@ impl TraceReader {
...
     }
+
+    /// Resolve an `InternedString` from the string pool that was active when
+    /// the event at `event_index` (an index into [`all_events`](Self::all_events))
+    /// was decoded. Returns a borrowed `&str` with no allocation.
+    ///
+    /// This is the zero-cost option for resolving individual interned fields.
+    /// For bulk resolution of all fields in a `Custom` event (with allocation),
+    /// see [`resolve_custom_fields`](Self::resolve_custom_fields).
+    pub fn get_interned_string(&self, event_index: usize, id: InternedString) -> Option<&str> {
```
jlizen: This API didn't feel QUITE right - ideally we would have some sort of borrowed equivalent of `resolve_custom_fields` that doesn't force you to match down into finding your intern string ID. But, we need to convert from `FieldValue::PooledString` to `FieldValue::String`, so we can just return a `&FieldValue`.

Probably we could do something with a Cow-like iterator that mostly returns borrowed values, and then returns an owned String, but then we are allocating for the resolved case.

Or we could avoid it with something like,
```
 enum ResolvedField<'a> { 
      Borrowed(&'a  FieldValue),
      Resolved(&'a str)
} 
```

Except, then you have to match on it anyway...

My conclusion was, this is premature optimization, just do something that makes the CLI analysis of custom events not mostly impossible, and wait until we have use cases that call for more.

dial9-trace-format/src/types.rs
```diff
@@ -80,6 +80,33 @@ pub enum FieldValue {
     StringMap(Vec<(Vec<u8>, Vec<u8>)>),
 }
 
+#[cfg(feature = "serde")]
+impl serde::Serialize for FieldValue {
```
jlizen, line 84: This is needed due to including the `Custom` variant in our `TelemetryEvent`, which as `Serialize` implemented on it.

dial9-tokio-telemetry/src/telemetry/analysis.rs
```diff
@@ -25,6 +26,13 @@ pub struct TraceReader {
     pub thread_names: HashMap<u32, String>,
     /// Key-value metadata from the most recent SegmentMetadata event.
     pub segment_metadata: Vec<(String, String)>,
+    /// Per-segment string pools for resolving `InternedString` fields.
+    /// Each entry is a snapshot of the string pool for one segment (between
+    /// mid-stream header resets).
+    segment_pools: Vec<HashMap<InternedString, String>>,
```
rcoh: Kinda seems like we should just resolve into strings at parse time for custom events

dial9-tokio-telemetry/src/telemetry/analysis.rs
rcoh: I think custom events should just resolve strings always instead of leaving them interned for simplicity

rcoh (approved):
Seems fine. I've been thinking that we might want to basically have the in memory representation on the parse side basically be exclusively "custom event" with nice accesors like ev["field_name"] then you can deserialize into specific shapes if you want

It's lower performance but I think the code will be simpler

dial9-tokio-telemetry/src/telemetry/analysis.rs
jlizen: Yeah, I guess I'll just do this.

PR #224, opened 2026-04-16
fix: Flush rotation coordination

rcoh:
When files were rotated you would frequently end up with event in one trace and some event in another for the same window of wall clock time.

This coordinates rotation with the flush buffers so it triggers the pre flush when the file is ready to rotate then the force flush on rotation.

The doesn't yet gracefully handle when the size trigger is hit. The recommendation is to set the time based rotation trigger sufficiently shorter than your size based trigger

I ran a test for a few minutes and verified that this produces good data


rcoh:
ran the benchmark locally just to be sure: 
```
=== Baseline (no telemetry) ===
[baseline] warmup done, measuring 30s...
{
  "requests": 8728073,
  "wall_time_secs": 30.001,
  "throughput_rps": 290926,
  "mean_lat_ns": 171731,
  "min_lat_ns": 22016,
  "p50_lat_ns": 171519,
  "p90_lat_ns": 229375,
  "p99_lat_ns": 276479,
  "p99_9_lat_ns": 316415,
  "max_lat_ns": 1254399
}

=== Telemetry enabled ===
[telemetry] warmup done, measuring 30s...
{
  "requests": 8140312,
  "wall_time_secs": 30.001,
  "throughput_rps": 271335,
  "mean_lat_ns": 184148,
  "min_lat_ns": 19968,
  "p50_lat_ns": 182271,
  "p90_lat_ns": 251903,
  "p99_lat_ns": 323071,
  "p99_9_lat_ns": 428543,
  "max_lat_ns": 1374207
}

=== Comparison ===
Baseline:   290926 req/s, p50=171.5µs, p99=276.5µs
Telemetry:  271335 req/s, p50=182.3µs, p99=323.1µs
Overhead:   6.7%

=== Trace Efficiency ===
Trace bytes/request:    N/A
Bytes/trace event:      N/A
Client requests/sec:    N/A

=== Validation ===
✅ PASS: Telemetry overhead acceptable (6.7%)
```

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -1059,6 +1101,17 @@ fn run_flush_loop(
...
         flush_timer.stop();
+
+        // Notify the writer that TL buffers have been drained and flushed.
+        // The writer may rotate the segment or just advance its drain timer.
+        // Skip on exit — finalize() below will seal the final segment.
+        if do_drain
+            && !exit
+            && let Err(e) = event_writer.drained()
+        {
+            tracing::warn!("failed to complete post-drain action: {e}");
```
jlizen, line 1112: do we want this to be rate limited?

dial9-tokio-telemetry/tests/rotation_time_alignment.rs
```diff
@@ -0,0 +1,206 @@
...
+        .build()
+        .unwrap();
+    guard.enable();
+
+    let mut builder = tokio::runtime::Builder::new_multi_thread();
+    builder.worker_threads(num_workers).enable_all();
+
+    let (runtime, _handle) = guard.trace_runtime("main").build(builder).unwrap();
+
+    // Generate continuous events across multiple rotation boundaries.
```
jlizen, line 55: I guess we can keep an eye on this for brittleness if our test process is not getting enough cycles? I don't see a great way to move off of wall clock time, but we can always make it run longer.

jlizen (commented):
Looks great

dial9-tokio-telemetry/tests/rotation_time_alignment.rs
jlizen: Hah, as I say that, I see CI just failed on this test on macs (presumably even more oversubscribed than average workers):
```
thread 'rotated_segments_have_bounded_time_overlap' (18024) panicked at dial9-tokio-telemetry/tests/rotation_time_alignment.rs:186:9:
segment 1 → 2 overlap is 3.806s, exceeds 2s tolerance. Segment 1 max=7767647666, segment 2 min=3961303541
```

jlizen:
Looking at the [CI failure](https://github.com/dial9-rs/dial9-tokio-telemetry/actions/runs/24540614933/job/71745391903?pr=224), it looks like we might have an actual issue @rcoh  - we have some events left behind

dial9-tokio-telemetry/tests/rotation_time_alignment.rs
jlizen: Actually per https://github.com/dial9-rs/dial9-tokio-telemetry/pull/224#issuecomment-4264418381 looks like there is more going on than that

jlizen (approved):
The TOCTOU fix in https://github.com/dial9-rs/dial9-tokio-telemetry/pull/224/changes/8752f54ef2d99094558ef897e1b6923fa12db241 looks plausible.

Anyway I see you also improved failure diagnostics, we shall see.

PR #231, opened 2026-04-16
Add `dial9-viewer agents` with a skill to analyze dial9 trace files

rcoh:
The main point of this PR is to build up a toolkit so that agents can analyze trace files. I did that by bundling a skill inside of the dial9-viewer binary – it includes both written text, example strategies and code + an easy way to grab a "dial9 toolkit" with all the files.

Here's the conclusion from it analyzing the demo trace:
```
Here's what's going on in demo-trace.bin:

  This is a Tokio runtime trace from a service called `metrics-service` (an Axum-based HTTP service
  listening on 0.0.0.0:3001), captured over ~4.4 seconds with 2 worker threads.

  Key findings

  Overall health: The runtime is mostly idle — workers are only ~13% active, spending most time
  parked in epoll. 268 tasks were spawned and all completed by trace end.

  The big problem — two ~179ms blocking polls: Both workers got blocked simultaneously at the 3.37s
  mark by a socket() syscall during HTTP connection establishment (hyper-util's
  HttpConnector::call). The kernel stacks show the block was in synchronize_rcu / expand_fdtable /
  expand_files — the kernel was expanding the file descriptor table, which triggered an RCU
  synchronization. This is a known issue where the kernel's FD table expansion can briefly stall all
  threads.

  Scheduling delays: 170 tasks experienced wake-to-poll latency >1ms. The worst was 178.77ms (caused
  by the blocking poll above). The p50 is 1.62ms and p99 is 3.84ms — generally acceptable except for
  that outlier.

  Task spawns: Most tasks (179) come from hyper-util's Tokio runtime adapter, and 86 from the Axum
  handler at axum_traced.rs:146. This looks like an HTTP load test or benchmark scenario.

  Off-CPU profile: 59% of off-CPU time is in epoll_wait (normal idle), 38% in futex_wait (mostly
  parking/synchronization). Small amounts of futex contention appear in AWS SDK types (ConfigBag,
  SdkBody) and Axum handler futures.

  CPU hotspots: Very few on-CPU samples (52 total), spread across malloc_consolidate, AWS-LC crypto
  (Keccak), and AWS SDK interceptors. No single CPU bottleneck.

  The actionable takeaway: The main issue is the ~179ms stall from the socket() syscall triggering
  kernel FD table expansion. This blocked both workers simultaneously. Possible mitigations:
  pre-warm connections (connection pooling), or increase the initial FD table size to avoid runtime
```


dial9-viewer/src/main.rs
```diff
@@ -1,67 +1,148 @@
...
-    port: u16,
+    #[command(subcommand)]
+    command: Commands,
+}
 
-    /// S3 bucket name (if omitted, bucket must be specified per-request)
-    #[arg(long)]
-    bucket: Option<String>,
+#[derive(Subcommand, Debug)]
+enum Commands {
```
jlizen, line 45: probably should call this out as a breaking change to the CLI, not that anybody is using it directly, probably. i don't think clap has great derive support for a default subcommand, though are ways you could hack around it (and also use the default port) as discussed in https://github.com/clap-rs/clap/issues/3857

dial9-viewer/build.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+fn generate_skills(manifest_dir: &str, out_dir: &str) {
+    let skills_dir = Path::new(manifest_dir).join("skills");
+    let dest = Path::new(out_dir).join("skill_files.rs");
+
+    let mut code = String::new();
+    let mut segments: Vec<(String, String, String)> = Vec::new(); // (name, title, path)
+
+    for entry in fs::read_dir(&skills_dir).unwrap() {
```
jlizen: `if skills_dir.is_dir()`?

dial9-viewer/toolkit/analyze.js
```diff
@@ -0,0 +1 @@
+../skills/analyze.js
```
jlizen, line 1: re: symlinks, I guess we anyway don't claim windows support yet?

dial9-viewer/skills/analysis.md
```diff
@@ -0,0 +1,103 @@
+# Analysis Pipeline
+
+After parsing, run the analysis pipeline to derive higher-level structures. All functions are in `trace_analysis.js`.
+
+## Standard pipeline
```
jlizen, line 5: This same pipeline appears in a lot of places, and a typical session might span multiple... is there a more context window friendly way of handling it?

dial9-viewer/skills/recipes.md
```diff
@@ -0,0 +1,233 @@
...
+  const wakes = wakesByTask[taskId];
+  if (!wakes || wakes.length === 0) return;
+  const lastWake = wakes[wakes.length - 1];
+  const loc = trace.taskSpawnLocs.get(taskId) || '(unknown)';
+  console.log(`${'  '.repeat(depth)}Task ${taskId} (${loc}) woken by task ${lastWake.wakerTaskId}`);
+  if (depth < 5) traceWakeChain(lastWake.wakerTaskId, wakesByTask, depth + 1, seen);
+}
+
+// Example: trace wake chain for the task with the longest poll
+traceWakeChain(worst.poll.taskId, spans.wakesByTask);
```
jlizen: I'd be careful about depending on variables from other recipes, it seems like it might trip up agents a bit. Compared to something self contained like:
```
  // Example: pick a task ID of interest and trace its wake chain
  const taskId = 42; // replace with a task ID from your trace
```

I also would suggest avoiding `traceWakeChain` relying on external state (`trace`) and instead take `taskSpawnlocs` as an explicit parameter.

Same feedback applies in various other places.

dial9-viewer/skills/loading.md
```diff
@@ -0,0 +1,128 @@
...
+# Decompress and concatenate multiple segments
+gunzip -k segment-001.bin.gz segment-002.bin.gz segment-003.bin.gz
+cat segment-001.bin segment-002.bin segment-003.bin > combined.bin
+```
+
+Pass the combined file to `parseTrace` as usual. The parser handles multiple concatenated segments transparently — headers, string pools, and schemas are re-read at each segment boundary.
+
+## Handling gzip
+
+`parseTrace` automatically decompresses gzip input. You can pass `.bin.gz` files directly.
```
jlizen: This is a bit confusing since we earlier say to decompress any gzipped segments first. Maybe move this to be earlier than merging multiple?

dial9-viewer/skills/analysis.md
```diff
@@ -0,0 +1,103 @@
...
+
+## computeSchedulingDelays(workerSpans, workerIds, wakesByTask)
+
+For each poll, finds the most recent wake event for that task before the poll started. The delay is `pollStart - wakeTime`. Returns:
+```
+[{wakeTime, pollTime, delay, taskId, wakerTaskId, worker, poll}]
+```
+Sorted by wakeTime. Large delays mean a task was woken but had to wait before being polled (workers were busy).
+
+## filterPointsOfInterest(filterType, workerSpans, workerIds, schedDelays, hasSchedWait, opts)
```
jlizen: nit: probably we should refactor the referenced API to use an object rather than a naked bool: ` {
  hasSchedWait: true, sortByWorst: true }`

dial9-viewer/skills/recipes.md
```diff
@@ -0,0 +1,233 @@
...
+A task leak means tasks are spawned but never terminate, causing the active count to grow monotonically.
+
+```javascript
+const samples = taskTimeline.activeTaskSamples;
+if (samples.length > 0) {
+  const first = samples[0].count;
+  const last = samples[samples.length - 1].count;
+  const peak = Math.max(...samples.map(s => s.count));
+  console.log(`Active tasks: start=${first}, end=${last}, peak=${peak}`);
+  if (last > first * 2 && last === peak) {
```
jlizen: This seems noisy, lots of traces start with fewer tasks and then have more as load ramps up. Maybe tweak to check if they are monotonically increasing at each tick, or just note that short traces might have ramp-up?

dial9-viewer/skills/analyze.js
```diff
@@ -0,0 +1,306 @@
...
+const { parseTrace, EVENT_TYPES, formatFrame, symbolizeChain, deduplicateSamples } = require(path.resolve(__dirname, 'trace_parser.js'));
+const { buildWorkerSpans, attachCpuSamples, buildActiveTaskTimeline,
+        computeSchedulingDelays, filterPointsOfInterest } = require(path.resolve(__dirname, 'trace_analysis.js'));
+
+const TRACE_PATH = process.argv[2];
+if (!TRACE_PATH) { console.error('Usage: node analyze.js <trace.bin>'); process.exit(1); }
+
+// ── Helpers ──
+
+/** Classify a symbol as kernel, libc, tokio-runtime, or application code */
```
jlizen: might be worth throwing in jemalloc and mimalloc symbols as well?

dial9-viewer/skills/recipes.md
```diff
@@ -0,0 +1,233 @@
...
+  console.log(`Worker ${w}: ${polls.length} polls in window`);
+  for (const p of polls) {
+    const dur = (p.end - p.start) / 1e6;
+    const rel = (p.start - minTs) / 1e6;
+    console.log(`  ${rel.toFixed(1)}ms +${dur.toFixed(2)}ms task=${p.taskId} spawn=${p.spawnLoc}`);
+  }
+}
+
+// Check queue depth at that time
+const nearestQueue = spans.queueSamples.reduce((best, s) =>
```
jlizen: this crashes on an empty array, should have a guard if `length > 0`

jlizen (approved):
The server infrastructure seems reasonable (and I assume this goes in `0.3` where we don't get about adding a mandatory CLI arg - not that CLIs REALLY follow semver anyway).

The content seems fine. The general format seems good, without having pointed an agent to it directly. I'm assuming you have already iterated a fair bit.

We should probably think about some sort of test harness that has traces with patterns and sees how well the agent identifies them. Even just something that we use manually to a/b test for now.

PR #232, opened 2026-04-16
Attempt to fix ci space issues

rcoh:


.github/actions/rust-build/action.yml
```diff
@@ -12,28 +12,22 @@ runs:
...
           run: |
+              if [ "${{ inputs.toolchain }}" = nightly ]; then
+                  rm -fv Cargo.lock
+              fi
               if [ "$RUNNER_OS" = "Linux" ]; then
                   # Raise locked-memory limit for large perf ring buffers
                   sudo prlimit --pid $$ --memlock=unlimited:unlimited
               fi
-              cargo test --all-features
+              cargo test --all-targets --all-features
```
jlizen, line 33: This seems fine, it will build in debug mode but who cares. But indeed it will sweep up binaries. A comment would be nice though.

.github/workflows/ci.yml
```diff
@@ -4,6 +4,9 @@ on: [pull_request, merge_group]
 permissions:
   contents: read
 
+env:
+  STABLE_TOOLCHAIN: "1.94.1"
```
jlizen, line 8: This seems annoying to maintain, why do we need this? It's not like stable is changing frequently enough that we really want to force ourselves onto something old

.github/workflows/ci.yml
```diff
@@ -34,7 +40,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        toolchain: [stable, nightly]
+        toolchain: ["1.94.1", nightly]
```
jlizen, line 43: If did want to keep the env, why hardcode it here?

jlizen (commented):
The change to consolidate build and test seems fine to me. Are we sure we don't have eg trace files leaking though? Or stale cargo cache entries? Did you find a way into the host to validate the disk space footprint?

The pinning the stable tool chain seems weird though, why do we need that?

.github/workflows/ci.yml
rcoh: Yeah the clippy failures happening in main are annoying and also I wasn't sure if the problem was a 1.95 regression

rcoh:
It is failing before we even run tests and the cache is only 500mb

jlizen (approved):
I'm fine to try it out, but can we have a backlog to unpin stable?

PR #233, opened 2026-04-16
feat: support sampling Schedule events

### LINKED ISSUE #79: support sampling Schedule events
Currently, we are recording all kernel scheduling event, however, in prod, sampling is probably sufficient to catch issues. This is controlled by the settings we pass to `PERF` when we open the perf event buffer.

yulnr:
closes #79

## Summary

Adds `SchedEventConfig::sampling_period` so callers can record 1-in-N context switches instead of every one. 
`None` (the default) preserves current behavior, `Some(10)` records every 10th context switch. 

Uses perf's period-based sampling rather than frequency-based, because I couldn't get frequency-based sampling to behave in a predictable or intuitive way. It seems like the kernel's frequency tuner undersamples SW events like context switches (perhaps because they're bursty?), so I was ending up with much lower sample ratio than what you would expect (and it was inconsistent).

Internally, this replaces `SamplerConfig::frequency_hz` with a new `SamplingMode` enum (`FrequencyHz`/`Period`) so `build_attr` can configure the perf attr depending on frequency being `1`.

Fixes `kernel_sched_events` example reading file path.


perf-self-profile/src/sampler.rs
```diff
@@ -24,13 +24,22 @@ pub enum EventSource {
...
 }
 
+/// The sampling mode to use.
+#[derive(Debug, Clone, Copy)]
+pub enum SamplingMode {
+    /// Sample at this frequency in Hz (e.g., 999 or 4000).
+    FrequencyHz(u64),
+    /// Record one sample per this many events. `1` = every event.
+    Period(u64),
+}
```
jv1i, line 34: Part of the reason for this change is that previously `frequency_hz` was used for both frequency-based and event-based (sched) sources, with 1 meaning "every event" for sched. 
Rather than extending that with more logic tied to frequency being or not being `1`, this should makes the intent more explicit.

perf-self-profile/src/sys/linux/sampler.rs
```diff
@@ -253,7 +273,14 @@ impl PerfSampler {
...
             | PERF_SAMPLE_CPU as u64
-            | PERF_SAMPLE_PERIOD as u64
+            // Omit PERF_SAMPLE_PERIOD in Period mode. When set, perf_swevent_event()
+            // emits a sample on every SW event, skipping the countdown 
+            // that implements 1-in-N sampling.
+            | if matches!(config.sampling, SamplingMode::FrequencyHz(_)) {
+                PERF_SAMPLE_PERIOD as u64
+            } else {
+                0
+            }
```
jv1i, line 283: Can't say for certain atm but I think this is where the skipping happens: https://github.com/torvalds/linux/blob/master/kernel/events/core.c#L10853

more precisely due to this: https://github.com/torvalds/linux/commit/5d81e5cfb37a174e8ddc0413e2e70cdf05807ace

dial9-tokio-telemetry/src/telemetry/cpu_profile.rs
```diff
@@ -45,6 +45,8 @@ impl Default for CpuProfilingConfig {
 /// so each worker thread gets its own perf fd via `on_thread_start`.
 #[derive(Debug, Clone, Default)]
 pub struct SchedEventConfig {
+    /// Record every Nth context switch. `None` records every event.
+    pub sampling_period: Option<u64>,
```
rcoh: Maybe interval? Period sounds like a unit of time.

Might as well make this a builder while we're here

rcoh (commented):
Nice! Overall looks great

rcoh (approved):

PR #236, opened 2026-04-17
feat(viewer): add local directory mode

### LINKED ISSUE #234: add a mode to the dial9-viewer where it is pointed at a local directory
right now it can be pointed at an s3 bucket but it would be useful for local dev to point it at your dial9 traces directory

rcoh:
## Summary

Add `--local-dir` flag to serve traces from a local directory instead of S3. Useful for local development — point the viewer at your dial9 traces directory without needing AWS credentials or S3.

**Usage:**
```bash
dial9-viewer --local-dir /tmp/my_traces
# or
dial9-viewer serve --local-dir /tmp/my_traces
```

## Changes

- **`LocalBackend`** in `storage.rs`: implements `StorageBackend` using filesystem operations — recursive walk for `list_objects`, `read_dir` for `list_prefixes`, `fs::read` for `get_object`. The `bucket` parameter is ignored.
- **`--local-dir` CLI flag** on both `Serve` subcommand and top-level args. When set, uses `LocalBackend` with a sentinel bucket so existing routes work unchanged.
- **8 new tests**: search (all files, prefix filter, empty dir, file sizes), trace fetch/decompress, not-found, prefix listing, and full e2e search→view flow.

## Testing

- All 22 tests pass (14 existing + 8 new)
- `cargo fmt --check` clean
- `cargo clippy --all-targets --all-features` clean
- `cargo nextest run --stress-duration 20s`: 18 iterations × 22 tests, zero failures

Closes #234


dial9-viewer/src/storage.rs
```diff
@@ -188,3 +189,120 @@ impl StorageBackend for S3Backend {
         })
     }
 }
+
+/// Local filesystem storage backend. Serves trace files from a directory.
+///
+/// The `bucket` parameter is ignored — all operations are relative to `root`.
+/// Keys are relative paths from `root`.
+pub struct LocalBackend {
```
jlizen, line 197: It's not every day that you see code to make a local filesystem look like s3 instead of the other way around ^^

dial9-viewer/src/storage.rs
```diff
@@ -188,3 +189,120 @@ impl StorageBackend for S3Backend {
...
+    }
+
+    fn get_object(
+        &self,
+        _bucket: &str,
+        key: &str,
+    ) -> Pin<Box<dyn Future<Output = Result<Vec<u8>, StorageError>> + Send + '_>> {
+        let path = self.root.join(key);
+        Box::pin(async move {
+            std::fs::read(&path).map_err(|e| match e.kind() {
```
jlizen: probably better to use tokio fs / spawn_blocking for these, even if it is just a dev server

dial9-viewer/src/storage.rs
```diff
@@ -188,3 +189,120 @@ impl StorageBackend for S3Backend {
...
+            Ok(prefixes)
+        })
+    }
+
+    fn get_object(
+        &self,
+        _bucket: &str,
+        key: &str,
+    ) -> Pin<Box<dyn Future<Output = Result<Vec<u8>, StorageError>> + Send + '_>> {
+        let path = self.root.join(key);
```
jlizen, line 288: This is a path traversal risk. I know this is just a local devtool, but in the age of agents, we probably do want to force things to stay within the root directory.

IE, we can canonicalize and check that the path starts with `self.root`.

Same applies to `list_prefixes`

dial9-viewer/src/main.rs
```diff
@@ -36,6 +36,10 @@ pub struct Cli {
     #[arg(long, global = true)]
     prefix: Option<String>,
 
+    /// Serve traces from a local directory instead of S3
+    #[arg(long, global = true)]
+    local_dir: Option<PathBuf>,
```
jlizen, line 41: we should make this conflict with bucket at both levels.

dial9-viewer/src/main.rs
jlizen: more than just s3 now right?

dial9-viewer/src/storage.rs
```diff
@@ -188,3 +189,120 @@ impl StorageBackend for S3Backend {
...
+            let mut prefixes = Vec::new();
+            for entry in entries {
+                let entry = entry.map_err(|e| StorageError::Other(e.to_string()))?;
+                if entry
+                    .file_type()
+                    .map_err(|e| StorageError::Other(e.to_string()))?
+                    .is_dir()
+                {
+                    let name = entry.file_name().to_string_lossy().into_owned();
+                    prefixes.push(format!("{prefix}{name}/"));
```
jlizen: Another thing that breaks on windows, not that I think we care? Might be worth leaving a breadcrumb though for when we someday do a cleanup to support non-unix

dial9-viewer/src/storage.rs
```diff
@@ -188,3 +189,120 @@ impl StorageBackend for S3Backend {
...
+        Box::pin(async move {
+            std::fs::read(&path).map_err(|e| match e.kind() {
+                std::io::ErrorKind::NotFound => StorageError::NotFound(path.display().to_string()),
+                _ => StorageError::Other(e.to_string()),
+            })
+        })
+    }
+}
+
+fn collect_files(
```
jlizen, line 316: This doesn't currently handle symlinks.

For directories, that would also lead to path traversal, we need to canonicalize them and check if they escaped root, like the other entrypoints.

For file, we silently drop them currently, but I'd probably also allow them but similarly guard against the traversal.

dial9-viewer/src/main.rs
jlizen, line 32: We don't need the arguments inside `Serve` if `global = true`, they are anyway allowed before OR after subcommands and then accessible under `cli.foo`

Though, we might want to add an extra guard against them being used with the `agents` subcommand... I don't think clap natively supports that.

Another option is to remove the `global` and keep them duplicated in both places, but set `#[command(args_conflicts_with_subcommands = true)]`, though that's also a bit confusing since then `dial9-viewer --local-dir /tmp serve` fails. Easier to maintain than runtime guards against N+1 arguments though.

jlizen (commented):
Probably is worth a simple README mention as well (in the `dial9-viewer` readme?

Though, the root README is a bit stale and just talks about opening up a HTML file, probably could use a refresh.

jlizen (approved):

PR #237, opened 2026-04-17
feat: include boot_id in default S3 key layout (#225)

### LINKED ISSUE #225: s3 keys should include a boot id
right now when the server restarts, you will see duplicated trace indexes which is somewhat confusing. the default trace format layout should include a boot id (randomly generated when server starts, 4 alpha characters or the base 64 encoded epoch start time)

Ensure that the browse interface in the viewer also supports this

rcoh:
Closes #225.

## Problem

When a service restarts, segment indices reset to 0 — producing duplicate
`{epoch_secs}-{index}` keys across successive runs. In the trace browser this
shows up as overlapping segments on the same host.

## Change

Inject a boot identifier into the default S3 key layout, before the filename:

\`\`\`
{prefix}/{YYYY-MM-DD}/{HHMM}/{service}/{instance}/{boot_id}/{epoch_secs}-{index}.bin.gz
\`\`\`

- Default \`boot_id\` is 4 lowercase alpha chars derived from the current system
  time nanoseconds, regenerated per process start.
- Existing \`S3ConfigBuilder::boot_id()\` callers see their value appear in the
  path *and* in the \`x-amz-meta-boot-id\` header (previously metadata only).

## Viewer

- \`parseKey()\` locates the \`YYYY-MM-DD\` segment and counts components to
  recognise both the new layout and the pre-#225 layout.
- Browse mode groups host cards by \`(service, host, boot_id)\` so distinct
  process runs are no longer blended together — segment-index duplicates
  are gone.
- New \`boot\` badge on host cards; new \`Boot\` column in raw search.
- \`test_parse_key.js\` covers both layouts.

## Tests

- New \`default_boot_id_is_four_lowercase_alpha_chars\`
- Updated key-format assertions in \`background_task::s3::tests\`
- Updated \`upload_and_delete_writes_to_s3_and_removes_local_file\` integration test
- Node \`test_parse_key.js\` for the viewer parser

\`cargo nextest run --workspace --all-features\`: 381 passed.
\`cargo nextest run --workspace --all-features --stress-duration 20s\`: passed.
\`cargo clippy --all-targets --all-features\`: clean.
\`cargo fmt --check\`: clean.

### DIFF
--- a/dial9-tokio-telemetry/src/background_task/s3.rs
+++ b/dial9-tokio-telemetry/src/background_task/s3.rs
@@ -45,7 +45,7 @@ pub struct SegmentInfo {
 /// Trait for custom S3 object key generation.
 ///
 /// Implement this to control the S3 key layout. The default key layout is
-/// `{prefix}/{date}/{HHMM}/{service}/{instance}/{epoch}-{index}.bin.gz`.
+/// `{prefix}/{date}/{HHMM}/{service}/{instance}/{boot_id}/{epoch}-{index}.bin.gz`.
 pub trait S3KeyFn: Send + Sync {
     /// Generate the S3 object key for the given segment.
     fn object_key(&self, segment: &SegmentInfo) -> String;
--- a/dial9-tokio-telemetry/src/background_task/s3.rs
+++ b/dial9-tokio-telemetry/src/background_task/s3.rs
@@ -67,7 +67,7 @@ where
 /// sensible defaults:
 ///
 /// - `instance_path`: system hostname
-/// - `boot_id`: `{pid}-{timestamp_nanos}` (unique per process start)
+/// - `boot_id`: `{4 alpha derived from timestamp}-{pid}` (unique per process start)
 /// - `prefix`: none (keys start at the time bucket)
 /// - `region`: auto-detected via `HeadBucket`
 /// - `key_fn`: built-in time-first layout
--- a/dial9-tokio-telemetry/src/background_task/s3.rs
+++ b/dial9-tokio-telemetry/src/background_task/s3.rs
@@ -10,13 +10,23 @@ use aws_sdk_s3_transfer_manager::Client;
 use std::collections::HashMap;
 use std::sync::Arc;
 
-/// Generate a per-process identifier from PID and current timestamp.
+/// Generate a short boot identifier (4 lowercase alpha chars).
+///
+/// Derived from the current system time nanoseconds. Different process
+/// restarts pick different values, letting you disambiguate segments
+/// from successive runs of the same service on the same host.
 fn default_boot_id() -> String {
     let nanos = std::time::SystemTime::now()
         .duration_since(std::time::UNIX_EPOCH)
         .unwrap_or_default()
         .as_nanos();
-    format!("{}-{}", std::process::id(), nanos)
+    let mut v = nanos as u64;
+    let mut s = String::with_capacity(4);
+    for _ in 0..4 {
+        s.push((b'a' + (v % 26) as u8) as char);
+        v /= 26;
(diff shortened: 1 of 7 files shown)


dial9-viewer/ui/index.html
```diff
@@ -349,23 +350,58 @@ <h1>dial9 Trace Browser</h1>
     }
 
     function parseKey(key) {
+        // Default layout (as of issue #225):
+        //   {prefix}/{YYYY-MM-DD}/{HHMM}/{service}/{instance}/{boot_id}/{epoch}-{index}.bin[.gz]
+        // Legacy layout (pre-#225):
+        //   {prefix}/{YYYY-MM-DD}/{HHMM}/{service}/{instance}/{epoch}-{index}.bin[.gz]
```
jlizen, line 356: do we have a plan for setting the old legacy parsing? worth a backlog?

dial9-tokio-telemetry/src/background_task/s3.rs
```diff
@@ -73,11 +90,12 @@ pub struct S3Config {
...
-    /// Identifies this process lifetime. Stored as S3 object metadata to
-    /// correlate segments from the same application run. A new value each
-    /// time the application restarts lets you group or filter by run.
+    /// Identifies this process lifetime. Included as both S3 object
+    /// metadata and in the default key path. A new value each application
+    /// start disambiguates segments (and segment indices) from different
+    /// runs of the same service on the same host.
     ///
-    /// Defaults to `{pid}-{timestamp_nanos}`.
+    /// Defaults to 4 random lowercase alpha characters.
```
jlizen, line 99: It's not actually random, it's time-based?

dial9-tokio-telemetry/src/background_task/s3.rs
```diff
@@ -10,13 +10,23 @@ use aws_sdk_s3_transfer_manager::Client;
...
+/// Derived from the current system time nanoseconds. Different process
+/// restarts pick different values, letting you disambiguate segments
+/// from successive runs of the same service on the same host.
 fn default_boot_id() -> String {
     let nanos = std::time::SystemTime::now()
         .duration_since(std::time::UNIX_EPOCH)
         .unwrap_or_default()
         .as_nanos();
-    format!("{}-{}", std::process::id(), nanos)
+    let mut v = nanos as u64;
```
jlizen, line 23: this collides within `0.5ms` restart due to the truncation, which is unlikely, but. why not also mix in the process id before converting to alphanumeric?

dial9-tokio-telemetry/src/background_task/s3.rs
```diff
@@ -61,9 +74,13 @@ where
 /// # Default key layout
 ///
 /// ```text
-/// {prefix}/{YYYY-MM-DD}/{HHMM}/{service_name}/{instance_path}/{epoch_secs}-{index}.bin.gz
+/// {prefix}/{YYYY-MM-DD}/{HHMM}/{service_name}/{instance_path}/{boot_id}/{epoch_secs}-{index}.bin.gz
 /// ```
 ///
+/// The `boot_id` segment disambiguates segment indices across process
+/// restarts — without it, a service that restarts will produce colliding
```
jlizen, line 82: nit: emdash

dial9-tokio-telemetry/src/background_task/s3.rs
jlizen: this is stale

jlizen (approved):

rcoh:
I think we should also probably do this in the writeback processor. It's actually even more useful for local dev when you might start the process multiple times. As it is, it overwrites which is somewhat annoying

PR #246, opened 2026-04-17
Bundle assets into binary & improve title bar

rcoh:


dial9-viewer/src/main.rs
```diff
@@ -119,42 +136,89 @@ async fn serve(
...
-        let backend = dial9_viewer::storage::S3Backend::from_env().await;
-        dial9_viewer::server::AppState::new(
-            std::sync::Arc::new(backend),
-            bucket.clone(),
-            prefix.clone(),
-        )
+        // Detect bucket region if a bucket is provided
+        if let Some(bucket_name) = &bucket {
+            if let Some(region) = detect_bucket_region(bucket_name).await {
+                tracing::info!(%region, bucket = %bucket_name, "detected bucket region");
```
jlizen, line 175: maybe make this debug?

jlizen (approved):
Looks like CI is failing due to unmaintained/unsound, do we want to tweak our cargo audit config to filter those? `metrique` has similar.

Fine by me to override and deal with separately though.

Issue #221, opened 2026-04-16
add ECS simulation test to CI


rcoh:
Add a test job to CI that simulates the permissions restrictions of ECS to validate the Ctimer based stack walking

Issue #220, opened 2026-04-16
add ASAN test to CI


rcoh:
Add test to CI that runs existing tests under address sanitizer

PR #250, opened 2026-04-18
feat: add CPU profiling fallback for perf-restricted environments

### LINKED ISSUE #219: flamegraphs do not work in ECS / fargate
ECS and fargate do not work with perf_event_open because of kernel restrictions on those platforms.

1. Implement detection that perf_event_open is actually working and log a warning on startup
2. Port the safe access pattern from async-profiler to install a SegSev handler to make fp unwinding safe
3. Implement ctimer based profiling in perf-self-profile that utilizes fp unwinding guarded by the safe access pattern

yulnr:
closes #219 #220 #221

## Summary

To provide CPU profiling on perf-restricted systems, this adds an automatic fallback using per-thread CPU timers + userspace frame-pointer unwinding, ported from async-profiler's approach.
When `perf_event_open` fails `PerfSampler` switches to ctimer mode.

The core of the fallback lives in a new `fp_profiler` module inside `perf-self-profile`, a short breakdown:
- safe load: asm trampoline + SIGSEGV handler for fault-tolerant pointer reads (so a corrupt FP chain aborts the walk instead of crashing the process)
- FP unwinder: walks the rbp/x29 chain from the signal handler's ucontext
- ctimer engine: `timer_create(CLOCK_THREAD_CPUTIME_ID)` with `SIGEV_THREAD_ID` per thread, fires SIGPROF when CPU time is consumed.
- lock-free sample buffer

Some re-organization of the sampler internals: extracted a `SamplerBackend` trait that both perf and ctimer backends implement. There's a big diff for the perf backend code being moved, but it's mostly moved as-is.

**Pairity:** Sched event profiling has no userspace fallback, this does not attempt to cover that. When perf is blocked it returns an error.

Thread registration for ctimer is exposed as `register_current_thread()` / `unregister_current_thread()` from the profiler crate. The recorder hooks call these directly, no-ops when perf is the active backend. 

For testing, we can use `DIAL9_FORCE_CTIMER=1` env var to force ctimer even when perf works.

### Testing 

Tested both backends side by side in Docker. Results are very close for the existing examples: sample count,
flamegraphs and traces have the same shape, etc.
The only difference I notice: ctimer stacks are 1 frame shallower (perf's kernel unwinder includes the vDSO entry frame, the FP walker stops at the userspace boundary).

I ran an extender version of the multithreaded example with different configurations, the sample count for both backends is similar:

| Config | perf | ctimer | Δ |
|--------|------|--------|---|
| 4t 99 Hz 5 s drain/100 ms | 2,015 | 1,996 | −0.9% |
| 16t 99 Hz 5 s drain/100 ms | 5,423 | 5,436 | +0.2% |
| 16t 999 Hz 5 s drain/100 ms | 52,600 | 50,409 | −4.2% |
| 64t 99 Hz 5 s drain/100 ms | 5,409 | 5,364 | −0.8% |
| 64t 999 Hz 5 s drain/100 ms | 52,477 | 50,405 | −3.9% |

These numbers might look different on a real machine, this was running on docker on my mac. 
(up to ~ 4-5% variance is usually noise between runs)

Two new CI jobs: ASAN on `perf-self-profile` (both perf and ctimer), and an ECS simulation that locks `perf_event_paranoid=3`.


dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -200,61 +190,44 @@ fn register_hooks(
...
+                    s_stop.thread_roles.lock().unwrap().remove(&tid);
+                }
                 if let Ok(mut prof) = s_stop.sched_profiler.lock()
                     && let Some(ref mut p) = *prof
                 {
                     p.stop_tracking_current_thread();
                 }
-            }
-        });
+                dial9_perf_self_profile::unregister_current_thread();
```
rcoh, line 261: worth a comment explaining how this differs from the scheduling profiling calls above.

perf-self-profile/src/sys/linux/fp_profiler/ctimer.rs
```diff
@@ -0,0 +1,253 @@
...
+use std::cell::Cell;
+use std::io;
+use std::mem;
+use std::ptr;
+use std::sync::atomic::{AtomicBool, AtomicI64, Ordering};
+
+const SIGEV_THREAD_ID: libc::c_int = 4;
+
+#[repr(C)]
+union SigvalUnion {
```
rcoh: there may a be a sys crate somewhere with these in it

perf-self-profile/src/sys/linux/fp_profiler/ctimer.rs
```diff
@@ -0,0 +1,253 @@
...
+static RUNNING: AtomicBool = AtomicBool::new(false);
+
+thread_local! {
+    static THREAD_TIMER: Cell<Option<libc::timer_t>> = const { Cell::new(None) };
+}
+
+/// Install SIGPROF handler and remember the sampling interval.
+///
+/// # Safety
+/// Modifies process-global signal state. Call once.
```
rcoh: safety comment should be clear about the guarantees the caller needs to uphold. "This function can be called only once for the lifetime of the process"? (or whatever the precise restriction is)

perf-self-profile/src/sys/linux/fp_profiler/ctimer.rs
```diff
@@ -0,0 +1,253 @@
...
+    Ok(())
+}
+
+pub fn stop() {
+    RUNNING.store(false, Ordering::Release);
+}
+
+/// Re-enable sampling after `stop()`. Per-thread timers keep firing;
+/// this just tells the signal handler to record samples again.
+pub fn resume() {
```
rcoh: these functions are per process or per thread? should just be clear either way

perf-self-profile/src/sys/linux/fp_profiler/ctimer.rs
```diff
@@ -0,0 +1,253 @@
...
+            if unsafe { timer_delete(t) } != 0 {
+                let err = io::Error::last_os_error();
+                tracing::warn!("ctimer: timer_delete failed in unregister_thread: {err}");
+            }
+        }
+    });
+}
+
+#[inline]
+fn gettid() -> libc::c_int {
```
rcoh: pretty sure we have this function in perf-self-profile somewhere already

perf-self-profile/src/sys/linux/fp_profiler/mod.rs
```diff
@@ -0,0 +1,201 @@
+//! Userspace frame-pointer profiling infrastructure.
+//!
+//! Three components:
+//! - **safe_load**: asm trampoline + SIGSEGV handler for fault-tolerant pointer
+//!   reads (so a corrupt FP chain aborts the walk instead of crashing).
+//! - **unwind**: frame-pointer stack walker using safe_load.
+//! - **ctimer**: per-thread CPU timer engine (`CLOCK_THREAD_CPUTIME_ID` +
+//!   `SIGEV_THREAD_ID`) that fires SIGPROF when CPU time is consumed.
+//!
+//! Linux x86_64 and aarch64 only.
```
rcoh: this says linux but the cfg gate below is only on arch. I _think_ the arch is actually correct here? I don't think this is linux specific

perf-self-profile/src/sys/linux/ctimer_sampler.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    fn for_each_sample(&mut self, f: &mut dyn FnMut(&Sample)) {
+        let dropped = sample_buffer::take_dropped_count();
+        if dropped > 0 {
+            tracing::debug!("[ctimer] dropped {dropped} samples (buffer full)");
+        }
+
+        sample_buffer::drain(|raw| {
+            let num_frames = raw.num_frames as usize;
+            let callchain: Vec<u64> = raw.frames[..num_frames.min(MAX_FRAMES)].to_vec();
```
rcoh: Because of the callback style API we can actually pass a slice here and not allocate a vec inside Sample.. not sure it's worth the squeeze here though this loop isn't that hot.

We may be able to eventually do the same optimization on the perf-event path

perf-self-profile/src/sys/linux/ctimer_sampler.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    unsafe {
+        let Some(mut slot) = sample_buffer::claim_slot() else {
+            return; // buffer full, sample dropped (counted internally)
+        };
+
+        let pid = libc::getpid() as u32;
+        let tid = libc::syscall(libc::SYS_gettid) as u32;
+        let cpu = sched_getcpu() as u32;
```
rcoh: get_cpu is not async signal safe so we can't call this but knowing this info isn't critical

rcoh (commented):
There may be a race condition in the ring buffer still trying to prove it.

Overall looking great

perf-self-profile/src/sys/linux/fp_profiler/sample_buffer.rs
```diff
@@ -0,0 +1,421 @@
...
+/// Claim a slot for writing from a signal handler. Returns a mutable
+/// reference to the slot's data fields (excluding `state`).
+///
+/// # Safety
+/// - Must be called from a signal handler context (async-signal-safe).
+/// - The returned pointer is valid only until the caller calls [`commit_slot`].
+/// - Only one signal handler invocation writes to a given slot (guaranteed
+///   by the atomic `fetch_add`).
+pub(crate) unsafe fn claim_slot() -> Option<SlotWriter> {
+    let idx = BUFFER.write_idx.fetch_add(1, Ordering::Relaxed);
```
rcoh: This has the effect of leaving these tombstone holes that lead to samples getting dropped.
This does `write_idx.fetch_add(1)` before the CAS on the slot state. 

When the CAS fails (slot still WRITING or READY from a previous cycle), the sample is correctly dropped — but write_idx has already advanced. 

The drain path processes strictly in order and stops at the first non-READY slot, so the "hole" blocks drain until a new producer eventually wraps around and writes to that slot, which can take up to BUFFER_CAP (16384) sample-claims.

This has the visible result of dropping something like 95% of samples under load because we hit these hole states.

perf-self-profile/src/sys/linux/fp_profiler/sample_buffer.rs
```diff
@@ -0,0 +1,421 @@
...
+#[repr(C)]
+struct SampleSlot {
+    state: AtomicU8,
+    pid: u32,
+    tid: u32,
+    time: u64,
+    cpu: u32,
+    period: u64,
+    num_frames: u32,
+    frames: [u64; MAX_FRAMES],
```
rcoh, line 33: I _think_ these need to technically be wrapped in `UnsafeCell` to correctly model them in Rust's memory model. They are being written to from the signal handler and read by the flush thread. They are synchronized by the atomics but we need to "tell" Rust that we're doing synchronization primitives by wrapping these in unsafecell.

rcoh (changes requested):

perf-self-profile/src/sys/linux/fp_profiler/ctimer.rs
jv1i: added a shared fn for this.

perf-self-profile/src/sys/linux/fp_profiler/mod.rs
jv1i: Updated it, nothing linux specific, I think the libc calls should work on macOS/freebsd also (the module itself is gated because it's under linux for now but we could move it later if we add macOS)

perf-self-profile/src/sys/linux/ctimer_sampler.rs
jv1i: replaced with a libc syscall

perf-self-profile/src/sys/linux/fp_profiler/sample_buffer.rs
jv1i: Right, wrapped them now.

perf-self-profile/src/sys/linux/fp_profiler/sample_buffer.rs
jv1i: Thanks for catching this, I've now reworked most of the sample buffer to use a Vyukov-ish ring buffer that doesn't have this problem. I tried a few smaller patch for this specifically (e.g. retries + CAS on writer_idx instead of fetch_add) but they felt hacky. 

I added some tests but I'm separately running some stress testing to see if I can replicate anything similar. I'll drop off for tonight but I'll check on results tomorrow.

.github/workflows/ci.yml
```diff
@@ -104,6 +114,59 @@ jobs:
...
+      - name: Enable perf_event_open and kallsyms
+        run: |
+          sudo sysctl kernel.perf_event_paranoid=1
+          sudo sysctl kernel.kptr_restrict=0
+      - name: Run perf-self-profile tests under ASAN (perf path)
+        run: |
+          sudo prlimit --pid $$ --memlock=unlimited:unlimited
+          cargo test -p dial9-perf-self-profile --lib --all-features \
+            -Z build-std --target x86_64-unknown-linux-gnu \
+            -- --skip sys::linux::symbolize::tests::resolve_kernel_symbol_returns_name
```
jv1i: I'm temporarily skipping this to check rest of steps. 
I haven't managed to get this to pass under this config, not sure if it's due to ASAN or some other config that differs with the other existing steps. 

I'd like to know if this is actually highlighting an existing problem we have where we can't resolve symbols under these conditions

jv1i: Update: `perf_event_paranoid=3` seems to be the cause of `/proc/kallsyms` showing zero kernel addresses.
Skipping now only for the fallback paths that force paranoid=3 (and not really something relevant for the frame-pointer unwinding path, since that doesn't include kernel addresses).

We might want to follow up this with proper detection and error/warning on symbolize.

perf-self-profile/src/sys/linux/ctimer_sampler.rs
```diff
@@ -0,0 +1,223 @@
...
+        let mut cpu = 0u32;
+        // SAFETY: `getcpu` writes one `u32` through `&mut cpu`, node/cache pointers are null (allowed).
+        let ok = libc::syscall(
+            libc::SYS_getcpu,
+            &mut cpu,
+            ptr::null_mut::<libc::c_void>(),
+            ptr::null_mut::<libc::c_void>(),
+        ) == 0;
+        if !ok {
+            cpu = 0; // fallback if getcpu fails
```
rcoh, line 178: Consider a todo to use a sentinel here that is not easily confusable with an actual value

perf-self-profile/src/sys/linux/fp_profiler/ctimer.rs
```diff
@@ -0,0 +1,237 @@
...
+
+/// Re-enable sampling after `disable`.
+pub fn enable() {
+    RUNNING.store(true, Ordering::Release);
+}
+
+/// Permanently disable sampling. Sets both flags so the handler self-disarms
+/// each thread's timer on its next tick. Not reversible; a subsequent `start`
+/// is required to sample again.
+pub fn disable_permanent() {
```
rcoh, line 99: So not exactly permanent but I guess permanent in practice

perf-self-profile/src/sys/linux/fp_profiler/unwind.rs
```diff
@@ -0,0 +1,248 @@
...
+        }
+
+        let saved_fp = unsafe { load(fp as *const usize) };
+        if saved_fp == SAFE_LOAD_FAULT {
+            break;
+        }
+        let ret_addr_slot = (fp + core::mem::size_of::<usize>()) as *const usize;
+        let ret_addr = strip_pac(unsafe { load(ret_addr_slot) });
+        if ret_addr == SAFE_LOAD_FAULT {
+            break;
```
rcoh, line 83: I wonder if we should increment a TL here or other atomic to track how often we're actually defaulting. 

Can be follow up scope. I'm mostly just curious if we manage to hit this, event in tests.

Ideally we would have at least tests on all platforms that exercise the sigsev trap

perf-self-profile/src/sys/linux/fp_profiler/unwind.rs
```diff
@@ -0,0 +1,248 @@
...
+        let n = unsafe { unwind(0x40_0000, base, base, &mut out) };
+
+        assert_eq!(n, 3);
+        assert_eq!(out[0], 0x40_0000); // interrupted PC
+        assert_eq!(out[1], 0x40_1000);
+        assert_eq!(out[2], 0x40_2000);
+    }
+
+    #[test]
+    fn frame_zero_is_always_the_interrupted_pc() {
```
rcoh, line 173: Do any of these actually test the sigsev handler? It looks like most are caught by heuristics

perf-self-profile/src/sys/linux/perf_sampler.rs
```diff
@@ -0,0 +1,472 @@
...
+            PERF_FLAG_FD_CLOEXEC as libc::c_ulong,
+        )
+    };
+    if fd < 0 {
+        let err = io::Error::last_os_error();
+        if err.kind() == io::ErrorKind::PermissionDenied {
+            return Err(io::Error::new(
+                io::ErrorKind::PermissionDenied,
+                format!(
+                    "perf_event_open denied ({}); for context-switch sampling \
```
rcoh, line 64: Follow up: an API to do a single shot probe to probe all the APIs available and give a true/false for each would be handy.

perf-self-profile/src/sys/linux/perf_sampler.rs
```diff
@@ -0,0 +1,472 @@
...
+            // values) in each sample. Only tracepoints produce meaningful raw
+            // data; CPU and context-switch sources have nothing to attach.
+            | if matches!(config.event_source, EventSource::Tracepoint(_)) {
+                PERF_SAMPLE_RAW as u64
+            } else {
+                0
+            };
+
+        // Use CLOCK_MONOTONIC so perf timestamps are in the same clock domain
+        // as Rust's `Instant::now()`. Without this, perf defaults to
```
rcoh, line 259: I don't think we ever actually compare to instant to my knowledge? Not sure off hand

rcoh (approved):
I want to do one more ai assisted pass before merging by but looks good to my human eyes

Issue #251, opened 2026-04-18
Improve dynamic event encoding API


jlizen:
# Improve dynamic event encoding API

## Today

Dynamic schema events require a `Vec<FieldValue>` per event, with the timestamp convention enforced at runtime:

```rust
impl Encodable for MyEvent {
    fn encode(&self, enc: &mut ThreadLocalEncoder<'_>) {
        let name = enc.intern_string(&self.name);
        enc.write_event(&self.schema, &[
            FieldValue::Varint(self.timestamp_ns), // must be first, runtime error if wrong
            FieldValue::Varint(self.counter),
            FieldValue::PooledString(name),
        ]);
    }
}
```

Static `#[derive(TraceEvent)]` events avoid the Vec by writing fields directly to the byte buffer via `EventEncoder`, which is public but has no ergonomic entry point for dynamic schemas.

Separately, `record_event` requires an `Encodable` struct, which forces cloning any borrowed data. Internally, the tracing layer uses `handle.with_encoder(|enc| { ... })` to avoid this, but it's `pub(crate)`.

## Desired

(All three APIs below (`write_event` with separate timestamp, `with_encoder`, and field-by-field `EventWriter`) are currently either missing or `pub(crate)`.)

```rust
// 1. Timestamp as a separate parameter (type-safe)
enc.write_event(&schema, timestamp_ns, &[
    FieldValue::Varint(counter),
    FieldValue::PooledString(name),
]);

// 2. Zero-copy encoding from borrowed data (no Encodable struct)
handle.with_encoder(|enc| {
    let name = enc.intern_string(&borrowed_data.name);
    enc.write_event(&schema, timestamp_ns, &[...]);
});

// 3. Zero-allocation field-by-field writing (no Vec<FieldValue>)
handle.with_encoder(|enc| {
    let mut ev = enc.event(&schema, timestamp_ns);
    ev.varint(counter);
    ev.intern(&borrowed_data.name);
    // validates field count on drop
});
```

Issue #247, opened 2026-04-17
Implicit worker attribution for custom events


jlizen:
## Problem

Custom events emitted via `record_event` are written into a thread-local buffer on the calling worker thread, but the wire format doesn't capture which worker wrote them. The worker association is implicit (determined by which thread-local buffer the event lands in) but lost after encoding.

This means the viewer can only render custom events on worker lanes if the event schema explicitly includes a `worker_id` field. Events without one (e.g., a simple `#[derive(TraceEvent)]` struct with just a timestamp and application fields) can't be placed on the timeline.

## Possible approach

In `record_event` (the public API for custom events), unconditionally prepend a `worker_id` field during encoding. The `record_event` path already has access to the `TelemetryHandle` and can call `current_worker_id()`. The implementation would wrap the user's `Encodable` in a thin adapter that encodes a `WorkerIdPrefix { worker_id: WorkerId }` event field before delegating to the inner event's `encode()`.

If the user's event already has a `worker_id` field, the schema would have two (the auto-prepended one and the user's). The viewer/decoder would use the first one. This is slightly redundant but avoids needing to inspect the schema at encode time.

Alternatively, the encoder could check whether the schema already contains a `worker_id` field and skip the prepend. This requires the `Encoder` to expose schema introspection, which it doesn't today.

The simplest v1: always prepend, accept the rare duplication.

## Impact

All custom events automatically appear on the correct worker lane in the viewer without requiring users to add boilerplate `worker_id` fields to their event structs.

jlizen:
One concern with accepting collisions is, some custom events might not actually be emitted on the worker that they relate to. So user-specified worker_id should be preferred, probably?

rcoh:
yeah I considered this approach but since the buffers are already linear, we could have the setting of worker id be considered "span like" and apply to all future events in the span. You would also need to have emitted a different schema, but I guess that is actually workable by wrapping the schema provided by the user.

Ultimately, I think we don't want worker id to be "special" there is really a set of fields that are _associated_ with the thread local buffer itself. You could imagine the TL buffer tracking these fields and emitting them the first time they appear.

```rust
#[derive(TraceEvent)]
struct Event {
   #[traceevent(tl_scoped)]
  worker_id: WorkerId
}

// when we emit this event, we actually only emit the worker_id if it is different (or has never been set) in the TL buffer).
```

That way _any_ event can attach a thread association even if its not coming from Tokio.

[...]

PR #252, opened 2026-04-18
feat: tracing layer

jlizen:
### Summary

Adds a `tracing-subscriber` Layer (`Dial9TokioLayer`) that emits span enter/exit events into dial9 traces, letting you see what happened inside each poll. This lets you identify interesting polls (by request id, duration compared to peer spans), see what is happening with subsegments of hte poll, and more. 

The layer hooks into `tracing`'s span lifecycle and emits per-callsite dynamic schemas, containing parent, field/values, worker id timestamp, etc. The viewer and analysis scripts correlates these with PollStart/PollEnd events on the same worker.

Key design decisions:

- **Feature-gated**: behind `tracing-layer` (not default) to avoid pulling `tracing-subscriber` into builds that don't need it.
- **Per-callsite schemas**: each `#[instrument]` or `span!` callsite gets its own wire schema with typed, interned fields. No StringMap overhead.
- **Interned field values**: all field values are interned as pooled strings. For most cases this is more compact than plain strings, though for high-cardinality values like request IDs it costs an extra byte. Could be more configurable someday.
- **Zero-copy encoding**: the layer uses an internal `with_encoder` API to write directly from span extensions without cloning field data.
- **Explicit parents only**: contextual parenting (`ctx.current_span()`) is unreliable across tasks on the same worker thread. Only `span!(parent: &x, ..)` parents are captured. The viewer infers nesting from timestamp containment.
- **No built-in filtering**: users compose with `tracing-subscriber`'s `Targets` filter. The metrics-service example demonstrates filtering to application spans only.

The viewer adds a dedicated span lane between worker lanes and task detail. You can filter and search by span name or field values, see percentile rank on hover, and clicking a span highlights the corresponding polls and parent chain across the timeline.

The analysis toolkit (`dial9-viewer agents`) now includes span data in the quick start template, span correlation recipes, and an unmatched span check in the red-flags scan.

#### Overhead

Criterion micro-benchmark results (NullWriter, current_thread runtime, 10s measurement):

| Scenario | Tracing only | With dial9 | dial9 overhead |
|---|---|---|---|
| 1 span enter+exit | ~190ns | ~300ns | ~100ns |
| 3 levels nested | ~630ns | ~760ns | ~45ns/span |
| 5 levels nested | ~1.0µs | ~1.24µs | ~50ns/span |
| 1 span with 3 fields | ~265ns | ~340ns | ~75ns |

The dial9 encoding adds ~50-100ns per span on top of tracing's own dispatch cost.

#### Schema mutex contention

The layer uses a `Mutex<HashMap>` for per-callsite schema lookup, shared across all worker threads. Benchmark: N threads each enter/exit 100 spans through a single shared `Dial9TokioLayer`:

| Threads | Wall time | Per-span |
|---|---|---|
| 1 | 72µs | 720ns |
| 4 | 269µs | 672ns |
| 8 | 481µs | 601ns |
| 32 | 1.9ms | 594ns |

Wall time scales linearly with thread count (threads run in parallel, not serialized). Per-span cost is flat or decreasing, confirming no mutex contention up to 32 threads.

### Testing

- End-to-end Rust tests covering the full pipeline (encode, flush, decode), explicit vs contextual parents, late field recording, no-handle safety, multi-worker concurrency, callsite schema dedup, and current_thread runtime regression.
- JS tests for span pairing, depth computation, cycle detection, and recycled span ID handling.
- Criterion micro-benchmark comparing tracing-only vs with-dial9 overhead (current_thread runtime, 10s measurement).
- [End-to-end validation](https://gist.github.com/jlizen/2a6e1a32adb1784ac6ab70640bc3d1ee): agent 0-shot instrumented an unfamiliar codebase (tokioconf mini-redis) using only the toolkit docs and correctly diagnosed the root cause.

### Follow-ups

- `write_event` API improvement: separate timestamp parameter, public `with_encoder`, zero-allocation encoding (#251).
- Implicit worker attribution for custom events (#247).


rcoh:
I did some prototypes and benchmarks. The TLDR I came to is "having 200ns per nested span is kinda scary but its basically inherent with how tracing layers work. Its faster than the JSON output of tracing subscriber by about 20% anyway.

I would really love to avoid to_stringing everything in the field visitor. But since we get the fields out of order, that seems really tricky.

I am a little worried about the global mutex on span context but we can see how it behaves in real applications before overly optimizing. I would be good to have some multithreaded benchmarks though.

jlizen:
> I would really love to avoid to_stringing everything in the field visitor. But since we get the fields out of order, that seems really tricky.

++ I initially was poking at this, but decided to punt...

> I am a little worried about the global mutex on span context but we can see how it behaves in real applications before overly optimizing. I would be good to have some multithreaded benchmarks though.

I added in some multithreaded benchmarks, not seeing any meaningful contention.

rcoh:
We may want a follow up ticket, in the output you sent the agent had a hard time actually finding the tracing events in the trace although it does find them eventually

dial9-viewer/skills/header.md
```diff
@@ -12,8 +12,65 @@ dial9 traces capture the internal behavior of a Tokio async runtime: task pollin
...
+// Wrap your runtime with TracedRuntime
+let writer = RotatingWriter::single_file("trace.bin")?;
+let mut builder = tokio::runtime::Builder::new_multi_thread();
+builder.worker_threads(4).enable_all();
+let (runtime, guard) = TracedRuntime::build_and_start(builder, writer)?;
+
+runtime.block_on(async { /* your app */ });
+
+drop(runtime);
+drop(guard); // flushes the trace
```
rcoh: This should suggest graceful shutdown. The agent was actually confused in the trace because we never closed the trace file

dial9-viewer/skills/red-flags.md
```diff
@@ -187,3 +233,12 @@ Workers are active (not parked) but spending less than 50% of wall time on CPU.
 
 ### kernel-sched-wait
 When a worker is woken (unparked), the kernel scheduling wait measures how long until the thread actually runs. High values indicate CPU contention at the OS level.
+
+### many-spans-per-poll
+When span events are present (`trace.customEvents`), check for polls containing many span intervals of the same name. This indicates a tight loop of synchronous operations inside a single poll (e.g., 50 redis commands where IO was immediately ready). The task should yield between iterations to let other tasks run. Use `buildSpanData(trace.customEvents)` and cross-reference with long polls to identify the culprit spans.
```
rcoh: Having this text verbatim seems a little too specific

dial9-viewer/skills/setup.md
```diff
@@ -0,0 +1,151 @@
...
+                .with_target("my_app", tracing::Level::TRACE)
+                .with_default(tracing::Level::ERROR),
+        ),
+    )
+    .init();
+```
+
+### Option B: Manual setup
+
+### Option B: Manual setup
```
rcoh: Header duplicate

dial9-viewer/skills/setup.md
```diff
@@ -0,0 +1,151 @@
...
+
+Replace your `tokio::runtime::Builder` with `TracedRuntime` and install `Dial9TokioLayer`:
+
+```rust,ignore
+use dial9_tokio_telemetry::telemetry::{RotatingWriter, TracedRuntime};
```
rcoh: I wonder if this file should just be the readme (e.g. symlink)

E.g. it is missing graceful shutdown

rcoh (approved):
Code looks good. May be worth reading through the agents guidance and doing small cleanups

dial9-viewer/skills/setup.md
jlizen: The README is way broader/verbose for an agent... I'll tweak our `build.rs` to pull out a list of relevant sections only.

jlizen:
> We may want a follow up ticket, in the output you sent the agent had a hard time actually finding the tracing events in the trace although it does find them eventually

I think that was before I added more steering explicitly around how to find the tracing events, but I'll make one more tweak in the header too.

PR #256, opened 2026-04-20
Support inline config for #[dial9_tokio_telemetry::main()]

### LINKED ISSUE #229: suport inline config definition for Dial9 macro
This is a follow-up for #212 
The idea is to support passing inline config to the macro, instead of just a function, would be like: 
```rust
#[dial9_tokio_telemetry::main(config = || {
  Dial9Config::new(tmp_base_path(), 1024 * 1024, 4 * 1024 * 1024)
})]
async fn main(){
  ...
}
```

Fluzko:
closes #229 

## Summary
This is a follow-up of #212 as per @jlizen's suggestion, to allow end users to pass an inline function as a shortcut for configuring Dial9.
With this PR, you're still allowed to do:
```rust
#[dial9_tokio_telemetry::main(config = my_config_function)]
async fn main() {
...
}
```
But you can also do:
```rust
#[dial9_tokio_telemetry::main(config = || {
  Dial9ConfigBuilder::new(...).build()
})]
async fn main() {
...
}
```

This aims to be a short and handy way for standard setups that don't need a bulky configuration.


dial9-macro/src/lib.rs
```diff
@@ -128,6 +152,17 @@ fn expand_main(args: MainArgs, input: ItemFn) -> Result<TokenStream2, syn::Error
 ///         .unwrap();
 /// }
 /// ```
+///
+/// Using an inline closure:
+///
+/// ```rust,ignore
```
jlizen, line 164: nit: any reason both of these can't be `no_run`?

dial9-macro/src/lib.rs
```diff
@@ -128,6 +152,17 @@ fn expand_main(args: MainArgs, input: ItemFn) -> Result<TokenStream2, syn::Error
...
+/// #[dial9_tokio_telemetry::main(config = || {
+///     Dial9ConfigBuilder::new("/tmp/trace.bin", 1024 * 1024, 16 * 1024 * 1024).build()
```
jlizen: Looks great! We could really use that builder API though... Anyway we have a trailing `build()` call :)

dial9-macro/src/lib.rs
```diff
@@ -2,17 +2,22 @@ use proc_macro::TokenStream;
...
+}
 
 struct MainArgs {
-    config: Path,
+    config: ConfigSource,
 }
 
-const MISSING_CONFIG_HELP: &str = "missing required `config = <fn>` argument, \
-                           e.g. #[dial9_tokio_telemetry::main(config = my_config)]";
+const MISSING_CONFIG_HELP: &str = "missing required `config` argument, e.g. \
```
jlizen: minor: Would it be worth showing both the inline and the external fn approach? I suspect external fn would be easier for newcomers

dial9-tokio-telemetry/examples/simple_workload.rs
```diff
@@ -12,22 +12,9 @@
 
 use std::time::Duration;
 
-use dial9_tokio_telemetry::config::{Dial9Config, Dial9ConfigBuilder};
+use dial9_tokio_telemetry::config::Dial9ConfigBuilder;
 use dial9_tokio_telemetry::telemetry::TelemetryHandle;
 
-fn my_config() -> Dial9Config {
```
jlizen, line 18: I'm fine to convert this to the inline, but is there anywhere left where we still have the longer-form? I guess the README, if nothing else?

dial9-tokio-telemetry/tests/macro_main.rs
```diff
@@ -32,6 +32,18 @@ fn macro_runs_async_body() {
     runs_async_body();
 }
 
+#[dial9_tokio_telemetry::main(config = || {
+    Dial9ConfigBuilder::new(tmp_base_path(), 1024 * 1024, 4 * 1024 * 1024).build()
+})]
+async fn runs_with_inline_closure() {
```
jlizen: Could we also add a test with a `move` closure?

jlizen (dismissed):
Looks great! Some non-blocking nits.

dial9-macro/src/lib.rs
Fluzko: Yes, the code below uses `dial9-tokio-telemetry` which is already importing `dial9-macro` crate, so would be a circular dep

dial9-tokio-telemetry/examples/simple_workload.rs
Fluzko: Matter in fact, I would say the opposite, I think we need to do more promotion to this one. The long `config = my_config` form is at:
- Main README
- Examples: `realistic_workload.rs`, `long_workload.rs`, `conditionally_enable.rs`
- Integration tests: `tests/macro_main.rs`

jlizen: Fair! Switching some more to this seems fine, up to you

dial9-macro/src/lib.rs
Fluzko: Makes sense, will implement it as part of this work

Fluzko:
I've spent some time reviewing the builder interface as per @jlizen's comment, since I wasn't that comfortable with the current `::new` & `::disabled` approach we introduced. The changes are [at this commit](https://github.com/dial9-rs/dial9-tokio-telemetry/pull/256/commits/62f4cac39bcb60a0334715960e2aebf13274bcdd)

Main changes are:
- I've put focus on having a cleaner API and usability + readability for end users when dealing with the config builder. 
- Remove `::new()` & `::disabled()`, now we have it default enabled, and we add a builder function `.enabled(bool)` which favors chaining instead of using `if` statements for configuration. If users decide to `.enable(false)` all dial9 configurations are still allowed to be passed as constructor methods, but at the end, ignored when built, and only take into account `.with_tokio()` calls.
- Introduce  a custom `Dial9Error` (should it be `Dial9ConfigError` instead?) that bubbles up tokio builder possible errors & also supports missing required parameters.
- Lean on `Bon` for wiring up the builder.
- Still have `Inner::Disabled` & `Inner::Enabled` even if we could do `if(self.enabled)`, since it's a no-cost and looks cleaner (open to suggestions)

dial9-tokio-telemetry/src/config.rs
```diff
@@ -161,56 +223,24 @@ impl Dial9ConfigBuilder {
...
-    pub fn with_tokio<F>(mut self, f: F) -> Self
+    /// The closure receives the staged builder by value and must return it.
+    /// Use this to access runtime configuration methods like
+    /// `with_runtime_name` and `with_task_tracking`; see
+    /// [`TracedRuntimeBuilder`] for the full list.
+    ///
+    /// Queued configurators are applied in call order during `build()`
+    /// once `base_path` is known. When `.enabled(false)` is set, queued
+    /// configurators are ignored.
+    pub fn with_runtime<F>(mut self, f: F) -> Self
```
jlizen, line 326: (non-blocking) - currently `with_runtime` discards all config if `enable(false)` - this is existing behavior, but it seems surprising. Is there a way to respect those settings?

dial9-tokio-telemetry/src/config.rs
```diff
@@ -18,14 +19,56 @@ use crate::telemetry::recorder::{
...
+// ---------------------------------------------------------------------------
+// Dial9Error — unified error for builder validation and runtime construction
+// ---------------------------------------------------------------------------
+
+/// Errors produced while building a [`Dial9Config`] or its tokio runtime.
+#[derive(Debug)]
+pub enum Dial9Error {
+    /// Telemetry is enabled (the default) but one or more required writer
+    /// fields were never set on the builder.
+    MissingFields(Vec<&'static str>),
```
jlizen: This is nice, but is there a way to avoid exposing its innards? Some opaque struct inner that has the debug/display? Current state is a semver risk.

dial9-tokio-telemetry/src/config.rs
```diff
@@ -18,14 +19,56 @@ use crate::telemetry::recorder::{
 };
 use crate::telemetry::writer::RotatingWriter;
 
+// ---------------------------------------------------------------------------
+// Dial9Error — unified error for builder validation and runtime construction
+// ---------------------------------------------------------------------------
+
+/// Errors produced while building a [`Dial9Config`] or its tokio runtime.
+#[derive(Debug)]
+pub enum Dial9Error {
```
jlizen: Should be `#[non_exhaustive]`

dial9-tokio-telemetry/src/config.rs
```diff
@@ -77,82 +120,101 @@ impl Dial9Config {
...
+        rotation_period: Option<Duration>,
+    ) -> Result<Dial9Config, Dial9Error> {
+        let enabled = enabled.unwrap_or(true);
+        if !enabled {
+            return Ok(Dial9Config(Inner::Disabled { tokio_builder }));
+        }
+
+        let (base_path, max_file_size, max_total_size) =
+            match (base_path, max_file_size, max_total_size) {
+                (Some(bp), Some(mfs), Some(mts)) => (bp, mfs, mts),
```
jlizen: nit: this is a bit ugly, is there a way to simplify?

dial9-tokio-telemetry/src/config.rs
```diff
@@ -77,82 +120,101 @@ impl Dial9Config {
...
-// Dial9ConfigBuilder — enabled path
+// Dial9ConfigBuilder — single bon-generated fluent entry point
 // ---------------------------------------------------------------------------
 
-/// Builder for a [`Dial9Config`] with telemetry enabled.
-///
-/// Created via [`Dial9ConfigBuilder::new`]. Exposes both tokio and dial9
-/// runtime knobs. Call [`.build()`](Self::build) to produce a [`Dial9Config`].
-#[derive(Debug)]
-pub struct Dial9ConfigBuilder {
```
jlizen, line 88: So it looks like we're opting to go the breaking change route, since we are changing `Dial9ConfigBuilder` and ripping out `DisabledDial9ConfigBuilder`? Releasing `0.4.0` isn't a big deal, just want to make sure that is your intent.

dial9-tokio-telemetry/src/config.rs
```diff
@@ -77,82 +120,101 @@ impl Dial9Config {
...
+        state_mod = dial9_config_builder,
+    )]
+    pub fn builder(
+        #[builder(field = default_tokio_builder())] tokio_builder: tokio::runtime::Builder,
+
+        #[builder(field)] runtime_configurators: Vec<RuntimeConfigurator>,
+
+        /// Defaults to `true`. When `false`, required writer fields are
+        /// ignored and the runtime is built without telemetry.
+        enabled: Option<bool>,
```
jlizen: nit:
```
#[builder(default = true)]
enabled: bool,
```

dial9-tokio-telemetry/src/config.rs
jlizen: (And either way we should have a test guarding whichever behavior)

jlizen (commented):
Great overall! Some nits and questions

dial9-tokio-telemetry/src/config.rs
Fluzko: Yes, I believe it's the best given the previous API, I feel this is a good step forward in terms of devx. That said, open to discuss if we should maintain the old API and deprecate it slowly.

dial9-tokio-telemetry/src/config.rs
Fluzko: I'm not getting the idea, all you can do with `with_runtime` is dial9 related, I don't see the need to not drop it when we want telemetry disabled, what do you find surprising about this behaviour?

dial9-tokio-telemetry/src/config.rs
Fluzko: I've did some rework around this, but honestly, I don't think there's any other safe approach. I was thinking something like 
```rust
if base_path.is_none() {
 missing.push("base_path");
}
// ...
let base_path = base_path.unwrap();
```
But you _can't guarantee_ that the unwrap is safe.

Check the current rework and let me know if you've got something else in mind.

dial9-tokio-telemetry/examples/conditionally_enable.rs
```diff
@@ -17,28 +17,21 @@
...
+        .enabled(std::env::var("ENABLE_DIAL9").is_ok())
+        .base_path("conditionally_enable_trace.bin")
+        .max_file_size(64 * 1024 * 1024)
+        .max_total_size(256 * 1024 * 1024)
+        .with_tokio(|t| {
+            t.worker_threads(4);
+        })
+        .with_runtime(|r| r.with_task_tracking(true))
+        .build()
+        .expect("config build failed")
```
rcoh: maybe `unwrap_or(Disabled)`? I think that is what people will probably do in prod rather than risk their app panicing on startup

dial9-tokio-telemetry/src/config.rs
jlizen: Ah my mistake, got mixed up about which thing this was modifying (had assumed tokio runtime)

dial9-tokio-telemetry/src/config.rs
jlizen: New approach lgtm

dial9-tokio-telemetry/examples/conditionally_enable.rs
Fluzko: I like this idea, what if we add a side function next to `.build()` that's `build_or`:
```rust
Dial9Config::builder()
            .base_path(tmp_base_path())
            .max_file_size(1024 * 1024)
            .max_total_size(4 * 1024 * 1024)
            .build_or(|builder, error| { 
                     builder.enabled(false) // << tbd new builder or the old one to be tweaked
                            .with_tokio(|t|){...}
                            .build()
                            .unwrap("tokio failed") // << we can't do much more on our end
            })
```

CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Breaking changes
```
jlizen: This is stale

dial9-tokio-telemetry/src/legacy_config.rs
```diff
@@ -0,0 +1,329 @@
...
+
+// ---------------------------------------------------------------------------
+// Dial9Config — opaque value the macro consumes
+// ---------------------------------------------------------------------------
+
+/// Finalized configuration consumed by the `#[main]` macro.
+///
+/// Constructed via [`Dial9ConfigBuilder::build`] or
+/// [`DisabledDial9ConfigBuilder::build`].
+#[deprecated(
```
jlizen: Instead of deprecating it (which will annoy anybody using it with compiler warnings), can we just state inline in the docs that it is not recommended and to use the other one instead? And add `#[do_not_recommend]`.

We have no real reason to rip it out at this point, which is what deprecation would signal is coming.

dial9-tokio-telemetry/src/config.rs
```diff
@@ -18,14 +19,79 @@ use crate::telemetry::recorder::{
...
 // ---------------------------------------------------------------------------
 // Dial9Config — opaque value the macro consumes
 // ---------------------------------------------------------------------------
 
 /// Finalized configuration consumed by the `#[main]` macro.
 ///
-/// Constructed via [`Dial9ConfigBuilder::build`] or
-/// [`DisabledDial9ConfigBuilder::build`].
+/// Constructed via [`Dial9Config::builder()`].
 #[allow(missing_debug_implementations)]
```
jlizen: nit: is there any particular reason not to have a debug impl on this? Even just showing which fields have been set if their values are awkward to display?

dial9-tokio-telemetry/src/legacy_config.rs
```diff
@@ -0,0 +1,329 @@
...
+    note = "use `dial9_tokio_telemetry::Dial9Config` instead — see module docs for migration"
+)]
+#[allow(missing_debug_implementations)]
+pub struct Dial9Config(Inner);
```
jlizen, line 32: Why do we need to have two different `Dial9Config` structs? They anyway appear to be identical?

I'm actually kind of surprised if this doesn't break usage of the macro with the old config given type mismatch...?

dial9-tokio-telemetry/tests/macro_main.rs
```diff
@@ -1,7 +1,7 @@
 use std::panic::{AssertUnwindSafe, catch_unwind};
 use std::path::PathBuf;
 
-use dial9_tokio_telemetry::config::{Dial9Config, Dial9ConfigBuilder};
```
jlizen, line 4: Can we also keep tests using the legacy config?

dial9-tokio-telemetry/src/config.rs
```diff
@@ -64,95 +132,121 @@ impl Dial9Config {
...
-        let runtime_builder = TracedRuntime::builder().with_trace_path(base_path.clone());
-        Self {
+    /// When telemetry is enabled (the default), the following setters are
+    /// required: [`base_path`](Dial9ConfigBuilder::base_path),
+    /// [`max_file_size`](Dial9ConfigBuilder::max_file_size),
+    /// [`max_total_size`](Dial9ConfigBuilder::max_total_size). They may be
+    /// omitted if `.enabled(false)` is set, in which case a plain tokio
+    /// runtime is built without telemetry.
+    #[builder(
+        builder_type = Dial9ConfigBuilder,
```
jlizen, line 188: This means that there are two possible `Dial9ConfigBuilder` structs, depending on whether you take the root export (new one) or use the `config::Dial9ConfigBuilder`. That wouldn't cause type resolution issues like with `Dial9Config`, but it still seems somewhat confusing.

I'm not quite seeing why we can't use `Dial9ConfigBuilder` for legacy too, and just have a wrapper / the worse old entrypoint (that is not recommended), eg:
```
  impl Dial9ConfigBuilder {
      pub fn new(base_path: impl Into<PathBuf>, max_file_size: u64, max_total_size: u64) -> Self {
          Self {
              inner: crate::Dial9Config::builder()
                  .base_path(base_path)
                  .max_file_size(max_file_size)
                  .max_total_size(max_total_size),
          }
      }

      pub fn disabled() -> DisabledDial9ConfigBuilder {
          DisabledDial9ConfigBuilder {
              inner: crate::Dial9Config::builder().enabled(false),
          }
      }

      pub fn build(self) -> Dial9Config { /* delegate */ }
      // ...
  }
```

But I know you tried it and it wound up messier, so it's fine if we do need to split.

In that case, though, we would need to keep the old one named as such to avoid breaking, and I don't think we want to have the punned-on names. Maybe just `Dial9Builder` for the new one?

dial9-tokio-telemetry/src/legacy_config.rs
Fluzko: #[diagnostic::do_not_recommend] is only for traits implementations, so here it won't do anything. Since the noise comes from the aggressive deprecation, I'm removing it and suggesting moving to the new API

dial9-tokio-telemetry/src/legacy_config.rs
Fluzko: That's why I proposed a wrapper, which is the other proposal I've done. 
Since we're encouraging moving to the new API and we're not maintaining it, I want a clear & fast removal approach for when we decide to move away. Coupling with the new API will make us need to keep maintaining the old one and coexisting without much sense. So, I would say there are two paths. If we want to keep the old API alive for a long time, we should couple it in the most important parts (final return types, core attrs, etc). Otherwise, I'll leave it as is and dispose of it when we're ready.

>I'm actually kind of surprised if this doesn't break usage of the macro with the old config, given type mismatch.

The trick with this is the import path of each one and having the same final return type

dial9-tokio-telemetry/src/config.rs
Fluzko: I went through the cognitive process that you mention, but I felt that having something named `Dial9` and running a build method over it would mean that you have an instance of `Dial9` and you can run it somehow, which is not the case with this; it's just its configuration.
I think that using it the way you mention will be wasting an opportunity to provide a way to instantiate Dial9 without a macro and do something like:

```rust
async fn run(){ /* My async code here */ }

fn main() {
  let cfg = Dial9Config::builder().blah().build();
  Dial9::new(cfg).block_on(run); // << same thing as the macro
}
```

Open to keep this discussion, honestly, I don't really have a _that strong_ an opinion about this, just pointing it out.

jlizen: You're right, I don't think we should pin ourselves down to a bad name. The only alternative I can think of is `Config`/`ConfigBuilder`, but since we export from root, I think that is probably a mistake.

dial9-tokio-telemetry/examples/conditionally_enable.rs
jlizen: We should show the pattern that doesn't involve a panic in all of our examples.

Maybe we can simplify to:
```
    Dial9Config::builder()
        .enabled(std::env::var("ENABLE_DIAL9").is_ok())
        .base_path("conditionally_enable_trace.bin")
        .max_file_size(64 * 1024 * 1024)
        .max_total_size(256 * 1024 * 1024)
        .with_tokio(|t| t.worker_threads(4);)
        .with_runtime(|r| r.with_task_tracking(true))
        .build_or_disabled()
       .expect("tokio failed")
```

That internally warns and holds onto the tokio config in case of io failures or missing fields?

We should also probably panic at debug levels if build is called with missing required fields.

dial9-tokio-telemetry/src/config.rs
```diff
@@ -64,95 +134,121 @@ impl Dial9Config {
...
+
+        /// Defaults to `true`. When `false`, required writer fields are
+        /// ignored and the runtime is built without telemetry.
+        #[builder(default = true)]
+        enabled: bool,
+        /// Trace output path.
+        #[builder(into)]
+        base_path: Option<PathBuf>,
+        /// Per-file rotation threshold in bytes.
+        max_file_size: Option<u64>,
```
jlizen, line 205: We should set some reasonable defaults for max file size / total size / rotation period rather than treating them not being set as an error. For base path, probably it can stay an error for now, though it's tempting to default to eg `/tmp/dial9/`...

dial9-macro/src/snapshots/dial9_macro__tests__expand_basic.snap
```diff
@@ -3,30 +3,9 @@ source: dial9-macro/src/lib.rs
...
-    } else {
-        __tokio_runtime
-            .block_on(async move {
-                do_work().await;
-            })
-    }
+    let __dial9_rt = ::dial9_tokio_telemetry::TelemetryRuntime::from_config(my_config());
+    __dial9_rt
+        .block_on(async move {
+            do_work().await;
```
rcoh, line 9: Nice! because `TelemetryRuntime` spawns the task

dial9-macro/src/lib.rs
```diff
@@ -105,18 +115,35 @@ fn expand_main(args: MainArgs, input: ItemFn) -> Result<TokenStream2, syn::Error
...
+/// Using a named function:
 ///
 /// ```rust,ignore
-/// use dial9_tokio_telemetry::{main, config::{Dial9Config, Dial9ConfigBuilder}, telemetry::TelemetryHandle};
+/// use dial9_tokio_telemetry::{main, Dial9Config, telemetry::TelemetryHandle};
 ///
 /// fn my_config() -> Dial9Config {
-///     Dial9ConfigBuilder::new("/tmp/trace.bin", 1024 * 1024, 16 * 1024 * 1024)
+///     Dial9Config::builder()
+///         .base_path("/tmp/trace.bin")
```
rcoh, line 145: would be great to show an example of disabling here

dial9-tokio-telemetry/examples/conditionally_enable.rs
```diff
@@ -17,28 +17,32 @@
...
+
+fn my_config() -> Dial9Config {
+    Dial9Config::builder()
+        .enabled(std::env::var("ENABLE_DIAL9").is_ok())
+        .base_path("conditionally_enable_trace.bin")
+        .max_file_size(64 * 1024 * 1024)
+        .max_total_size(256 * 1024 * 1024)
+        .with_tokio(configure_tokio)
+        .with_runtime(|r| r.with_task_tracking(true))
+        .build()
```
rcoh: this is equivalent to `build_or_disabled` right?

dial9-tokio-telemetry/src/telemetry_runtime.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+/// A tokio runtime paired with its (optional) dial9 telemetry guard.
+///
+/// The guard, when present, must outlive the runtime so traces are flushed on
+/// drop - keeping both inside one struct enforces that ordering at the type
+/// level. Construct one via [`TelemetryRuntime::try_from`] from a
+/// [`Dial9Config`] (strict) or [`Dial9ConfigFallback`] (lenient).
+#[derive(Debug)]
+pub struct TelemetryRuntime {
```
rcoh: we have `TracedRuntime` in one place and `TelemetryRuntime` here. should this be called `TracedRuntime`?

dial9-tokio-telemetry/README.md
```diff
@@ -60,6 +60,61 @@ async fn main() {
...
+
+#[dial9_tokio_telemetry::main(config = my_config)]
+async fn main() {
+    // Telemetry may or may not be active; check with try_current().
+    use dial9_tokio_telemetry::telemetry::TelemetryHandle;
+    if TelemetryHandle::try_current().is_some() {
+        // wake-event-tracked spawn
+    } else {
+        // plain tokio fallback path
+    }
```
rcoh: IMO `telemetry_handle::current()` should always exist but sometimes it is "inert" and spawns just go right to Tokio. 

this code would be quite annoying in practice

dial9-tokio-telemetry/examples/conditionally_enable.rs
Fluzko: yes, forgot to change it, tks

dial9-tokio-telemetry/src/telemetry_runtime.rs
Fluzko: `TracedRuntime` is in charge of installing the hooks and returns the tuple, which is later used by `TelemetryRuntime`, which is a higher-level wrapper that owns both the tokio `Runtime` and an`Option<TelemetryGuard>` in one struct, so drop ordering is enforced at the type level

dial9-tokio-telemetry/README.md
jlizen: ++, I could live with this being a follow-up PR, but it seems simple enough to tweak that it might be nice to throw in here in the interest of clean examples

I had a comment along similar lines with `TelemetryRuntime::handle()` - [ref](https://github.com/dial9-rs/dial9-tokio-telemetry/pull/256#discussion_r3164903631)

dial9-tokio-telemetry/src/telemetry_runtime.rs
jlizen: I like this new API, but it is indeed a bit awkward that we have both names sticking around. It seems like we could under the hood cut over `TracedRuntime` to be this nicer struct instead of the tuple, without any REAL semver breakage? Assuming we preserve the legacy methods that return the two portions, and document them as not recommended?

Concretely, something like:
```
  // TracedRuntime becomes the real struct (what TelemetryRuntime is today)
  pub struct TracedRuntime {
      runtime: tokio::runtime::Runtime,
      guard: Option<TelemetryGuard>,
  }

  impl TracedRuntime {
      // New recommended API (currently on TelemetryRuntime)
      pub fn from_config<C>(config: C) -> Self
      where
          C: TryInto<TracedRuntime>,
          <C as TryInto<TracedRuntime>>::Error: std::fmt::Display,
      { ... }

      pub fn block_on<F>(&self, fut: F) -> F::Output { ... }
      pub fn runtime(&self) -> &tokio::runtime::Runtime { ... }

    // guard() method, or delegation methods that are no-ops if not initialized as discussed in my comment on that method

      // builder() stays exactly as today but has doc comments recommending against, maybe #[docs(hidden)] it to avoid confusion
      pub fn builder() -> TracedRuntimeBuilder<NoTracePath> { ... }
  }

  // Old builder methods keep their tuple return types unchanged
  impl TracedRuntimeBuilder<HasTracePath> {
      pub fn build_and_start(self, builder, writer)
          -> io::Result<(Runtime, TelemetryGuard)>
      {
          // works exactly as before
      }
  }
```

  Existing callers are unaffected:
```
  // Still works, no code changes needed
  let (runtime, guard) = TracedRuntime::builder()
      .with_trace_path(path)
      .build_and_start(tokio_builder, writer)?;
```

  New callers get one unified type:

```
  let rt = TracedRuntime::from_config(
      Dial9Config::builder()
          .base_path("/tmp/trace.bin")
          .max_file_size(64 * 1024 * 1024)
          .max_total_size(256 * 1024 * 1024)
          .build()
          .expect("config")
  );
  rt.block_on(async { /* ... */ });
```

The only actual breakage would be pretty esoteric, technically needs `0.4`, but shouldn't require end user code migration. Basically it breaks `let x = TracedRuntime`, which nobody has any reason to do since there are no methods on it. We might even be comfortable releasing this under 0.3?

dial9-tokio-telemetry/src/config.rs
```diff
@@ -1,220 +1,365 @@
...
+    /// `RotatingWriter` / telemetry-core I/O failure during
+    /// [`crate::TelemetryRuntime::try_from`] silently downgrades to a
+    /// plain tokio runtime built from the same configurators. Only
+    /// [`std::io::Error`] (from the tokio builder itself) can still
+    /// escape that conversion.
+    ///
+    /// Counterpart to [`build`](Self::build), which preserves today's
+    /// strict behavior — RotatingWriter / telemetry-core I/O failures
+    /// propagate as [`crate::TelemetryRuntimeError`] from the strict path.
+    pub fn build_or_disabled(self) -> Dial9ConfigFallback {
```
jlizen: Bikeshed: `Dial9ConfigFallback` implies we DID fallback, maybe `MaybeDial9Config`?

This becomes moot if we move i/o validation inside the config builder though, [ref](https://github.com/dial9-rs/dial9-tokio-telemetry/pull/256/changes#r3164838775)

dial9-tokio-telemetry/src/telemetry_runtime.rs
```diff
@@ -0,0 +1,326 @@
...
+    runtime: tokio::runtime::Runtime,
+    guard: Option<TelemetryGuard>,
+}
+
+impl TelemetryRuntime {
+    /// Build from a config, panicking with the underlying error if
+    /// construction fails. Used by the `#[dial9_tokio_telemetry::main]`
```
jlizen: not sure if it's worth mentioning the macro usage here, an end user would reach for this if they are instrumenting runtimes manually right? For instance due to multiple runtimes, wanting graceful shutdown, etc.

dial9-tokio-telemetry/src/config.rs
```diff
@@ -1,220 +1,365 @@
...
-    /// Create a [`DisabledDial9ConfigBuilder`] that builds a plain tokio
-    /// runtime with no telemetry. Only `.with_tokio()` is available.
-    pub fn disabled() -> DisabledDial9ConfigBuilder {
-        DisabledDial9ConfigBuilder::new()
+/// Shared finish-fn body: builder-staged fields > [`Inner`].
+///
+/// Used by both `build()` (which propagates the error) and
+/// `build_or_disabled()` (which substitutes [`Inner::Disabled`] on error).
+#[allow(clippy::too_many_arguments)]
+fn assemble(
```
jlizen: nit: struct param syntax would be better here

dial9-tokio-telemetry/src/config.rs
```diff
@@ -1,220 +1,365 @@
...
+    /// escape that conversion.
+    ///
+    /// Counterpart to [`build`](Self::build), which preserves today's
+    /// strict behavior — RotatingWriter / telemetry-core I/O failures
+    /// propagate as [`crate::TelemetryRuntimeError`] from the strict path.
+    pub fn build_or_disabled(self) -> Dial9ConfigFallback {
+        let cfgs_for_fallback = self.tokio_configurators.clone();
+        match self.build() {
+            Ok(cfg) => Dial9ConfigFallback(cfg.0),
+            Err(_) => Dial9ConfigFallback(Inner::Disabled {
```
jlizen: Can we have a debug_assert panic if required fields are missing? (unless we can figure out a nicer type-state way to make missing fields impossible, but I'm fine to punt on it, can add compile-time validation later... i would argue unconditional runtime -> compile time failure is not a breaking change)

Also probably a error log (including release mode)?

dial9-tokio-telemetry/src/telemetry_runtime.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// macro.
+    ///
+    /// Generic over any input that converts into a [`TelemetryRuntime`],
```
jlizen: This is getting fairly confusing since we have two different failure points (during config build, during from_config), and also two different failure handling modes (strict or lenient).

Ideally we would prefer to have the strict/lenient distinction stop mattering to users after the user build the config (even if we handle the generic internally). Since, that is the point where they are making the choice to be lenient or no while calling `build()`.

### Today:

Config build/build_or_disabled:
- missing fields -> err result on strict, whole thing becomes non-op on lenient

From_conf:
- transport i/o failures -> panic on strict, non-op on lenient
- tokio failure -> both strict and lenient panic

### My suggestion:

Config build/build_or_disabled:
- missing fields -> err on strict, non-op the whole thing on lenient
- transport i/o failure -> err on strict, non-op on lenient

From_conf:
- tokio failure -> both panic (or return result for the try version)

I think it's fairly intuitive to probe for i/o reachability inside the builder where you are setting the transport related fields.

That simplifies this function to only be fallible in the case of tokio runtime failure, which will be much easier to describe.

It also has the side benefit of no longer needing a distinct `Dial9ConfigFallback` type, since we are only branching locally in `build_or_disabled()`

jlizen: Either way, we should have a clearer doc block on panics:

```
## Panics

Panics if tokio runtime cannot be constructed. For fallible construction...  <doc example showing the fallible approach>
```

(or also mention I/O failures here, if we do keep I/O failures inside the second build phase)

dial9-tokio-telemetry/src/telemetry_runtime.rs
```diff
@@ -0,0 +1,326 @@
...
+        } => {
+            let runtime = materialize_tokio_builder(&tokio_configurators)
+                .build()
+                .map_err(TelemetryRuntimeError::TokioRuntimeBuilder)?;
+            Ok((runtime, None))
+        }
+    }
+}
+
+impl TryFrom<Dial9Config> for TelemetryRuntime {
```
jlizen: Consider adding `TelemetryRuntime::try_from_config` for more locally readable callsites:
```
  let rt = TelemetryRuntime::try_from_config(cfg)?;
```
(or `try_build_from`)


We could still keep this trait impl, though I might personally prefer to have our own crate-local bridge trait to call inside our `try_from_config` helper (or whatever name). `TryFrom` is generally used for fallible data transformations, whereas we are overloading it here to model I/O errors.

IE I might add:
```
  pub(crate) trait IntoRuntime {
      type Error: std::fmt::Display;
      fn into_runtime(self) -> Result<TelemetryRuntime, Self::Error>;
  }
```

Neither are blocking concerns though

dial9-tokio-telemetry/src/telemetry_runtime.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// construction fails. Used by the `#[dial9_tokio_telemetry::main]`
+    /// macro.
+    ///
+    /// Generic over any input that converts into a [`TelemetryRuntime`],
+    /// so it accepts strict ([`Dial9Config`]), lenient
+    /// ([`Dial9ConfigFallback`]), and the deprecated positional
+    /// [`crate::config::Dial9Config`] configs transparently. The
+    /// generic shape is what keeps the macro source-compatible across
+    /// these three input types.
+    pub fn from_config<C>(config: C) -> Self
```
jlizen: `from_config` naming doesn't feel right, since it has a lot of extra stuff happening beyond a conversion (i/o, runtime setup, etc).

Some alternatives:

On the runtime type:
- `new(config)` / `try_new(config)`
- `build(config)` / `try_build(config)`

On the config type:
 - `Dial9Config::build_runtime()` / `try_build_runtime()`

The config side approach is nice since it chains well:
```
  let rt = Dial9Config::builder()
      .base_path(...)
      .max_file_size(...)
      .max_total_size(...)
      .build()? // or build_or_disabled()
      .build_runtime();
```

dial9-tokio-telemetry/src/telemetry_runtime.rs
```diff
@@ -0,0 +1,326 @@
...
+            .unwrap_or_else(|e| panic!("failed to initialize runtime: {e}"))
+    }
+
+    /// Borrow the underlying tokio runtime.
+    pub fn runtime(&self) -> &tokio::runtime::Runtime {
+        &self.runtime
+    }
+
+    /// Borrow the telemetry guard, if telemetry was enabled.
+    pub fn guard(&self) -> Option<&TelemetryGuard> {
```
jlizen: Similar to @rcoh 's comment about `TelemetryHandle::current` returning something inert, I think we ideally would have this always return a `TelemetryGuard` as well, and have the guard support non-op versions of its methods? `graceful_shutdown()`, etc.

And we could add an `is_enabled()` method to the guard?

Ideally we want to avoid people having to locally handle the "might already be disabled cases" both when spawning tasks and when doing graceful shutdown and similar things.

WDYT?

dial9-tokio-telemetry/src/config.rs
```diff
@@ -1,220 +1,365 @@
...
+// Dial9ConfigBuilderError — unified error for builder validation and runtime construction
+// ---------------------------------------------------------------------------
+
+/// Errors produced while building a [`Dial9Config`].
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum Dial9ConfigBuilderError {
+    /// Telemetry is enabled (the default) but one or more required writer
+    /// fields were never set on the builder.
+    MissingFields(MissingFields),
```
jlizen: Minor: I might prefer to keep this even more opaque, since I suspect we can get rid of the runtime missing field validation at some point using an even more complicated typestate pattern (that accommodates the disabled method). And then we are stuck with a now-vestigial variant.

In practice, the end user will anyway just log the error message to see what is missing, which we can anyway produce without strong typing.

That said per [this comment](https://github.com/dial9-rs/dial9-tokio-telemetry/pull/256/changes#r3164838775) I was suggesting we move validating the transport layer I/O into this builder. For that one, surfacing `std::io::Result` is useful. So maybe instead an enum that has a `Transport` variant (or `IO`).

So if we do want an enum, maybe rename this variant to a more generic `Validation` that has a stringified error?

WDYT? Not a huge deal either way.

dial9-tokio-telemetry/src/config.rs
```diff
@@ -1,220 +1,365 @@
...
     }
+}
 
-    /// Finalize into a [`Dial9Config`] ready for the macro.
-    pub fn build(self) -> Dial9Config {
-        Dial9Config(Inner::Disabled {
-            tokio_builder: self.tokio_builder,
-        })
+impl<S: dial9_config_builder::IsComplete> Dial9ConfigBuilder<S> {
+    /// Finish into a [`Dial9ConfigFallback`] that never reports a build
```
jlizen: Discussed more [here](https://github.com/dial9-rs/dial9-tokio-telemetry/pull/256/changes#r3164838775), I think we could simplify this a good bit of we just probed for transport i/o failures inside this function. Then, this means, you always get a working dial9 config, or a non-op, and then after that, you still need to build the tokio runtime successfully.

dial9-tokio-telemetry/src/config.rs
```diff
@@ -1,220 +1,365 @@
...
+    /// emits a [`Dial9ConfigFallback`] in its disabled state with the
+    /// user's `with_tokio` configurators preserved. The resulting
+    /// config also opts into a runtime-time cascade: any
+    /// `RotatingWriter` / telemetry-core I/O failure during
+    /// [`crate::TelemetryRuntime::try_from`] silently downgrades to a
+    /// plain tokio runtime built from the same configurators. Only
+    /// [`std::io::Error`] (from the tokio builder itself) can still
+    /// escape that conversion.
+    ///
+    /// Counterpart to [`build`](Self::build), which preserves today's
```
jlizen: "today's strict behavior" won't age well.

Maybe rephrase as something like, use Self::build() to explicitly handle failures to build Dial9Config.

dial9-tokio-telemetry/src/config.rs
jlizen: If we moved I/O checks into here, I wouldn't have a debug_assert panic on failure, that seems annoying, but I would definitely have an error level log for that too

examples/metrics-service/src/axum_traced.rs
```diff
@@ -170,12 +170,8 @@ where
...
 {
-    match TelemetryHandle::try_current() {
-        Some(handle) => {
-            handle.spawn(fut);
-        }
-        None => {
-            tokio::spawn(fut);
-        }
-    }
+    // `TelemetryHandle::current()` always returns a usable handle —
```
jlizen: Is this comment relevant here? My intuition would be to just show the usage without drawing attention to it

dial9-tokio-telemetry/examples/conditionally_enable.rs
```diff
@@ -17,28 +17,20 @@
...
-        "conditionally_enable_trace.bin",
-        64 * 1024 * 1024,
-        256 * 1024 * 1024,
-    )
-    .with_tokio(|t| {
-        t.worker_threads(4);
-    })
-    .with_runtime(|r| r.with_task_tracking(true))
-    .build()
+    Dial9Config::builder()
```
jlizen, line 24: api looks nice

dial9-tokio-telemetry/examples/long_workload.rs
```diff
@@ -1,17 +1,21 @@
...
+    Dial9Config::builder()
+        .base_path("long_trace.bin")
+        .max_file_size(64 * 1024 * 1024)
+        .max_total_size(256 * 1024 * 1024)
         .with_tokio(|t| {
             t.worker_threads(4);
         })
         .with_runtime(|r| r.with_task_tracking(true))
         .build()
+        .expect("config build failed")
```
jlizen: maybe: `dial9 failed to start`?

CHANGELOG.md
```diff
@@ -7,22 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `TelemetryGuard::is_enabled()` and a no-op `Disabled` mode: a `TelemetryGuard` is now always present on a `TracedRuntime`, regardless of whether telemetry was installed. `TelemetryGuard::handle()` returns an inert `TelemetryHandle` on a disabled guard, and `graceful_shutdown()` is a successful no-op there.
 
-### Other
+### Changed
 
-- Make worker lanes scrollable in viewer ([#275](https://github.com/dial9-rs/dial9-tokio-telemetry/pull/275))
-- Fix docs.rs broken links by using absolute GitHub URLs ([#277](https://github.com/dial9-rs/dial9-tokio-telemetry/pull/277))
+- **Breaking:** `TelemetryHandle::current()` no longer panics off-runtime. It now returns an inert handle when called from a thread that is not owned by a dial9 runtime, mirroring the always-present-guard model. Use `TelemetryHandle::is_enabled()` to branch on whether telemetry is live on the current thread.
+- **Breaking:** `TelemetryHandle::try_current()` is deprecated in favor of `current()`; call sites that relied on the `Option` ceremony can now drop it.
```
jlizen: Is there a reason to deprecate it compared to just having a doc comment to suggest against it? I don't think we want to force userspace code changes by hard deprecating it, so the deprecation notices are just irritating.

CHANGELOG.md
```diff
@@ -7,22 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [0.3.4](https://github.com/dial9-rs/dial9-tokio-telemetry/compare/dial9-tokio-telemetry-v0.3.3...dial9-tokio-telemetry-v0.3.4) - 2026-04-23
-
 ### Added
```
jlizen, line 11: Can we trim this added block into just 1 or 2 points that are more user oriented (longer is fine)? IE, 
- "the macro now supports inline config" - code example
 - "build_or_disabled provides a convenient way to fail open in prod" - code example

The breaking changes or major behavior changes are worth itemizing explicitly too (IE the changed section)

It's also fine to just wait to fill this out until we cut the release, so that the commit hash can go in it, but nice to get a jump on it now while it's fresh.

CHANGELOG.md
```diff
@@ -7,22 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `TelemetryHandle::disabled()` — explicit constructor for an inert handle whose `spawn` falls through to `tokio::spawn` and whose control methods are no-ops. `TelemetryHandle::is_enabled()` distinguishes the live and inert modes.
+- `TelemetryGuard::is_enabled()` and a no-op `Disabled` mode: a `TelemetryGuard` is now always present on a `TracedRuntime`, regardless of whether telemetry was installed. `TelemetryGuard::handle()` returns an inert `TelemetryHandle` on a disabled guard, and `graceful_shutdown()` is a successful no-op there.
 
-### Other
+### Changed
 
-- Make worker lanes scrollable in viewer ([#275](https://github.com/dial9-rs/dial9-tokio-telemetry/pull/275))
-- Fix docs.rs broken links by using absolute GitHub URLs ([#277](https://github.com/dial9-rs/dial9-tokio-telemetry/pull/277))
+- **Breaking:** `TelemetryHandle::current()` no longer panics off-runtime. It now returns an inert handle when called from a thread that is not owned by a dial9 runtime, mirroring the always-present-guard model. Use `TelemetryHandle::is_enabled()` to branch on whether telemetry is live on the current thread.
```
jlizen: I would argue that this isn't REALLY a breaking change, or at least not one we should highlight as such. I don't think we would bump semver over it.

CHANGELOG.md
```diff
@@ -7,22 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **Breaking:** `TelemetryHandle::try_current()` is deprecated in favor of `current()`; call sites that relied on the `Option` ceremony can now drop it.
+- **Breaking:** `TracedRuntime::guard()` returns `&TelemetryGuard` (always-present) instead of `Option<&TelemetryGuard>`. Replace `rt.guard().is_some()` with `rt.guard().is_enabled()`.
```
jlizen: The ergonomics of the new API seem better but do we think it is worth breaking semver over @rcoh rather than just adding a newer method that returns the always-present handle, and maybe hiding (or recommending against the old)?

dial9-tokio-telemetry/examples/optional_telemetry.rs
```diff
@@ -0,0 +1,60 @@
+//! Optional / best-effort telemetry via `build_or_disabled`.
+//!
+//! `Dial9Config::builder().build_or_disabled()` returns a
```
jlizen: This seems somewhat redundant with `conditionally_enable.rs`. It feels like it might be better to just quietly change all the examples to build_or_disabled? Is there a reason to highlight this so heavily when it is basically always the right choice to use?

dial9-tokio-telemetry/README.md
```diff
@@ -60,6 +60,62 @@ async fn main() {
 
 The macro automatically spawns your function body as a task, so top-level code is visible in traces (unlike plain `#[tokio::main]` where `block_on` work is invisible — see [below](#the-root-future-is-not-instrumented)). dial9 installs a `TelemetryHandle` on every runtime-owned thread via `on_thread_start`. Call `TelemetryHandle::current()` to get it for spawning wake-tracked sub-tasks.
 
+### Optional telemetry on I/O failure
```
jlizen: This feels a bit too detailed for the README. More like module-level docs on the runtime

Maybe we just update the other examples in the README to use `build_or_disabled`, and somewhere mention that this will return a pass-through handle on failure, and to use regular `build()` for a fallible builder?

dial9-tokio-telemetry/README.md
```diff
@@ -60,6 +60,62 @@ async fn main() {
...
+use dial9_tokio_telemetry::{Dial9Config, TracedRuntime};
+
+let cfg = Dial9Config::builder()
+    .base_path("/tmp/my_traces/trace.bin")
+    .max_file_size(1024 * 1024)
+    .max_total_size(5 * 1024 * 1024)
+    .build()
+    .expect("config build failed");
+
+let rt = TracedRuntime::try_new(cfg).expect("runtime build failed");
```
jlizen: nit: `.expect("tokio runtime failed to start")`?

dial9-tokio-telemetry/README.md
```diff
@@ -60,6 +60,62 @@ async fn main() {
...
+        // any code paths specific to "telemetry on"
+    }
+}
+```
+
+See [`examples/optional_telemetry.rs`](/dial9-tokio-telemetry/examples/optional_telemetry.rs) for an end-to-end run including a `DIAL9_TRACE_PATH=/unwritable/...` mode that exercises the downgrade.
+
+### Without the macro
+
+The macro expands to `TracedRuntime::new(...).block_on(...)`. If you'd rather drive that yourself — for tests, libraries that build their own runtime, or any code that doesn't own `main` — `TracedRuntime` is a public type that accepts a `Dial9Config`:
```
jlizen: Or if you need to handle graceful shutdown, or if you want to instrument multiple tokio runtimes

jlizen (commented):
Changes look great. Remaining tweaks are just about docs / examples, and then I wanted to get a second opinion from @rcoh about the breakage of `TracedRuntime::handle()`.

I think it's sparsely enough used that it's probably ok, with `0.4` right around the corner anyway, but just checking in about it.

CHANGELOG.md
jlizen: Wait a sec, is `TracedRuntime::guard()` even a method in the existing codebase? It seems like this might have been a change from an earlier version of this PR. I don't htink we need to documment it as a change at all / let alone breaking.

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -316,93 +305,167 @@ fn attach_runtime(
...
 
-    /// Return the [`TelemetryHandle`] for the current thread, or `None` if
-    /// no dial9 runtime has claimed this thread.
+    /// Return the [`TelemetryHandle`] installed for the current thread,
+    /// or `None` if no dial9 runtime has claimed this thread.
+    ///
+    /// Prefer [`current`](Self::current) — it returns an inert handle
+    /// in the same situations and avoids the `Option` ceremony at
+    /// callsites.
+    #[deprecated(
```
jlizen: I don't think we should deprecate this, since anyway we probably don't want to break people calling it currently in a future release, at least for a while. Probably instead I would just just discourage use its usage. (I don't know if i would go as far as docs(hidden)-ing it)

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -316,93 +305,167 @@ fn attach_runtime(
...
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("TelemetryHandle").finish_non_exhaustive()
+        f.debug_struct("TelemetryHandle")
+            .field("enabled", &self.is_enabled())
+            .finish_non_exhaustive()
     }
 }
 
 impl TelemetryHandle {
+    pub(crate) fn enabled(
```
jlizen, line 362: Currently this will collide with #332 which uses `enabled()` as a getter for the bool (though I prefer `is_enabled()` and will suggest as such).

Either way though there will be some colliding code, #332 's goes a little deeper into shared state so I'd probably rebase and take that.

dial9-tokio-telemetry/src/config.rs
```diff
@@ -1,220 +1,377 @@
...
+    /// Lenient counterpart to [`build`](Self::build). Use
+    /// [`build`](Self::build) instead when you want validation and
+    /// writer-I/O failures to surface as
+    /// [`Dial9ConfigBuilderError`].
+    pub fn build_or_disabled(self) -> Dial9Config {
+        let cfgs_for_fallback = self.tokio_configurators.clone();
+        match self.build() {
+            Ok(cfg) => cfg,
+            Err(e) => {
+                tracing::error!(
```
jlizen: Oh, another concern - this typically will be called before the tracing registry is set up, right? Should we use `eprintln!` instead? Or perhaps both?

If we want something fancier, we could check `tracing::dispatcher::has_been_set` to decide.

Are you aware of any conventions here @rcoh ?

rcoh (approved):

PR #257, opened 2026-04-20
Add docker compose for Linux demo trace generation

Fluzko:
Since CPU-profiling and, probably, other future features are scoped to Linux only, we need a secure & standard way to re-generate them regardless of the running OS.
This PR aims to provide a `compose.yml` file with a side script `regenerate_demo_trace_docker.sh` to do this work.


jlizen (approved):
Looks great, good call!

jlizen:
A future improvement might be to spin up `s3s` as well?

PR #258, opened 2026-04-20
fix: Commit setup skill as a file instead of generating at build time

jlizen:
Fixes: https://github.com/dial9-rs/dial9-tokio-telemetry/actions/runs/24669703712/job/72136756576

### Summary

The `dial9-viewer` build script was extracting sections from `../dial9-tokio-telemetry/README.md` at build time to generate a setup skill. This broke `cargo package` (and `cargo publish`) because the tarball verification build runs from an isolated directory under `target/package/`, where the relative path to the sibling crate's README doesn't exist.

The fix commits `setup.md` directly into `dial9-viewer/skills/` so it's included in the crate tarball. The build script no longer reads the README or generates the setup skill; instead, `setup.md` is picked up automatically like every other skill file in the directory.

To keep the README as the single source of truth, `scripts/sync-setup-skill.sh` regenerates `setup.md` from the README sections. A `--check` flag (used in CI) diffs the committed file against a fresh generation and fails if they diverge.

### Testing

Verified `cargo package -p dial9-viewer` succeeds (previously panicked during tarball verification). Confirmed the staleness check script detects drift and passes when the file is current. All clippy, fmt, and check pass.


jlizen:
Superseded by #259

PR #261, opened 2026-04-20
feat: add cargo binstall support

### LINKED ISSUE #249: remove `dev-server` from binaries that are published by dial9-viewer crate

jlizen:
(also fixes #249)

### Summary

Adds pre-built release binaries and `cargo-binstall` support for `dial9-viewer`. Previously, installing the viewer required compiling from source (`cargo install dial9-viewer`), which is slow due to the large dependency tree (AWS SDK, axum, etc.).

A new `build-binaries.yml` workflow triggers on `dial9-viewer-v*` releases (created by release-plz) and cross-compiles for five targets: Linux x86_64/aarch64, macOS x86_64/aarch64, and Windows x86_64. The asset naming convention follows the `cargo-binstall` default discovery pattern, configured via `[package.metadata.binstall]` in Cargo.toml.

The workflow also supports `workflow_dispatch` for manual builds without a release (uploads as artifacts instead of release assets).

Also excludes `dev-server` from the published crate (#249). The `dev-server` binary is a local development tool (fake S3 server) that shouldn't ship to end users. Adding `exclude = ["src/bin/dev_server.rs"]` to `[package]` keeps it out of the tarball while preserving it for local development.

### Testing

Verified `cargo package -p dial9-viewer` succeeds and the tarball no longer contains `dev_server.rs`. The workflow itself can only be fully validated once a release tag is pushed; the `workflow_dispatch` path can be used to test the build matrix before that.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -1,0 +1,77 @@
+# Builds cross-platform binaries and uploads to GitHub Releases.
+# Binaries are discoverable by cargo-binstall (https://github.com/cargo-bins/cargo-binstall).
+# Triggered automatically when release-plz publishes a dial9-viewer release.
+name: Build Binaries
+
+on:
+  release:
+    types: [published]
+
+permissions:
+  contents: write
+
+jobs:
+  build:
+    name: Build ${{ matrix.target }}
+    runs-on: ${{ matrix.os }}
+    if: startsWith(github.event.release.tag_name, 'dial9-viewer-v')
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+            archive: tar.gz
+          - target: aarch64-unknown-linux-gnu
+            os: ubuntu-24.04-arm
+            archive: tar.gz
+          - target: x86_64-apple-darwin
+            os: macos-latest
+            archive: tar.gz
+          - target: aarch64-apple-darwin
+            os: macos-latest
+            archive: tar.gz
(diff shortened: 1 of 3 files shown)


dial9-viewer/README.md
```diff
@@ -2,6 +2,18 @@
 
 CLI tool that serves a web UI for browsing and viewing [dial9-tokio-telemetry](../dial9-tokio-telemetry) trace files stored in S3 or on the local filesystem.
 
+## Installation
+
+Pre-built binaries are available from [GitHub Releases](https://github.com/dial9-rs/dial9-tokio-telemetry/releases) for Linux (x86_64, aarch64), macOS (x86_64, aarch64), and Windows (x86_64).
+
+```bash
+# From source via crates.io
+cargo install dial9-viewer
```
rcoh: Should say --locked probably

.github/workflows/build-binaries.yml
```diff
@@ -0,0 +1,98 @@
+# Builds cross-platform binaries and uploads to GitHub Releases.
+# Binaries are discoverable by cargo-binstall (https://github.com/cargo-bins/cargo-binstall).
+# Triggered automatically when release-plz publishes a dial9-viewer release.
+# workflow_dispatch lets you build binaries manually without a release.
+name: Build Binaries
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
```
rcoh: Workflow dispatch is quite hard to secure let's remove it for now

rcoh (approved):

.github/workflows/build-binaries.yml
jlizen: Will do

PR #263, opened 2026-04-20
chore: release 0.3.2

jlizen:
Release, with a side of `CONTRIBUTING.md`. We'll see if the binstall works properly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,12 @@ jobs:
   check-docs:
     name: Check docs.rs
     runs-on: ubuntu-latest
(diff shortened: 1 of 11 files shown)


jlizen:
Ah `cargo package` step has the chicken and egg problem, let me fix it...

jlizen:
Ok, now we skip `cargo package` on release PRs (sort of, weak check for `release/` in branch name).

Since we have issues with chicken-and-egg if we name versions that are not yet upstream, and yet we also don't want to use a workspace dryrun since we want to catch issues where things are missing from the `dial9-viewer` tarball due to crates.io install.

I think there might be something nicer we can do with `release-plz release --dry-run` but it's annoying to test locally. It also only tests crates with version bumps.

rcoh (approved):

PR #264, opened 2026-04-20
fix(ci): build binaries in same workflow as release-plz to avoid token issues

jlizen:
### Summary

The `build-binaries.yml` workflow (added in #261) never triggered because `release.yml` creates GitHub releases using `GITHUB_TOKEN`. GitHub's security policy prevents events created by `GITHUB_TOKEN` from triggering other workflows, to avoid infinite loops.

The fix switches `release.yml` to use a PAT (`RELEASE_PLZ_TOKEN`) for both the checkout and the `GITHUB_TOKEN` env var passed to release-plz. This attributes the release event to a real user, allowing it to trigger `build-binaries.yml`. See the [release-plz docs on triggering downstream workflows](https://release-plz.dev/docs/github/trigger).

**Manual setup required before merging**: create a fine-grained PAT scoped to `dial9-rs/dial9-tokio-telemetry` with `contents: write` and `pull-requests: write` permissions, then add it as a repository secret named `RELEASE_PLZ_TOKEN` at https://github.com/dial9-rs/dial9-tokio-telemetry/settings/secrets/actions.

### Testing

No way to test this without a real release. The next release after the secret is configured will validate the full chain: release-plz creates the release, which triggers `build-binaries.yml` to upload platform binaries.


rcoh (approved):

jlizen:
Updating to do it all in the same workflow to avoid needing a PAT

PR #269, opened 2026-04-21
feat(viewer): Pop-out flamegraph with interactive features

### LINKED ISSUE #38: make an option to "pop out" the flamegraph
When you open a flame graph, it should be possible to easily pop it out into another window (by loading the data in as a URL param)

This window should support all the standard flamegraph stuff (searching zooming etc.)

Ideally we would then make it easy to diff as well

yulnr:
Closes #38

## Summary

Extracts the flamegraph renderer into a shared `flamegraph.js` module and adds a standalone `flamegraph.html` page. The viewer's flamegraph panel now has a "Pop Out" button that opens the current selection (with zoom state) in a dedicated tab via URL params. For file-dropped traces it uses a Blob URL, which shows a warning toast in case you attempt to share between regular and private browsing (the blob becomes inaccesible).

The shared module owns the full flamegraph lifecycle: spawn location filtering, worker/offworker splitting, tree building, rendering, zoom, frame search, etc.

Search shows result stats with matched frame count and sample percentage, scoped to the current zoom level.

The standalone page persists zoom state as the full frame path in the URL and restores it on load. 
If the zoom state range in the URL doesn't match the trace and can't be reproduced, it shows all samples with a warning.

### Demo
https://github.com/user-attachments/assets/81fc82f5-d706-432d-813d-c0658e1e2883

## Future Improvements

- Keyboard navigation (e.g. after search be able to navigate and zoom between results)
- Diffing between different traces
- Regex search
- Per-worker flamegraphs?
- Support for inverted/icicle graphs? (for finding hot leaf functions)

### DIFF
--- a/dial9-viewer/ui/flamegraph.css
+++ b/dial9-viewer/ui/flamegraph.css
@@ -39,15 +39,16 @@
     max-width: 350px;
 }
 .fg-help-btn {
-    color: #666;
+    color: #aaa;
     cursor: pointer;
-    font-size: 0.85em;
-    border: 1px solid #444;
-    border-radius: 3px;
-    padding: 0 5px;
-    line-height: 1.4;
+    font-size: 1.1em;
+    background: none;
+    border: 1px solid #555;
+    border-radius: 4px;
+    padding: 2px 6px;
+    line-height: 1;
 }
-.fg-help-btn:hover { color: #ccc; border-color: #666; }
+.fg-help-btn:hover { color: #fff; border-color: #888; }
 
 .fg-breadcrumb {
     display: none;
--- a/dial9-viewer/ui/flamegraph.css
+++ b/dial9-viewer/ui/flamegraph.css
@@ -29,6 +29,15 @@
 }
 .fg-search-clear:hover { color: #fff; }
 .fg-search-stats { color: #888; font-size: 0.82em; }
+.fg-spawn-filter {
+    background: #2a2a4a;
+    border: 1px solid #444;
+    color: #ccc;
+    padding: 3px 6px;
+    border-radius: 4px;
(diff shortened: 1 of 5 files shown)


rcoh:
watched the video, this look sweeeet!

rcoh:
one small nit. When I click on a flamegraph node to focus on it, it would be less jarring if the node was the new bottom of the tree (instead of being omitted)

rcoh:
few more requests:
- I think we actually have the full path of the samples. it would be nice if mousing over showed that in a built-in browser tooltip or something for example, the trace shows `BytesToHexChars::next` and I can't see to figure out where that is coming from
- we support docs.rs linking for other spots, would be cool to support it in the flame graph
- did we lose the flamegraph segment that showed threads that are _not_ tokio workers?

yulnr:
> the node was the new bottom of the tree (instead of being omitted)

fixed!

---

> I think we actually have the full path of the samples. it would be nice if mousing over showed that in a built-in browser tooltip or something for example, the trace shows BytesToHexChars::next and I can't see to figure out where that is coming from
> we support docs.rs linking for other spots, would be cool to support it in the flame graph

Good ideas! the tooltip now includes the filename + line number, which can be expanded to full path when the tooltip is pinned. 
You can use shortcuts to pin the tooltip or open the docs.rs link (when available) on a new tab:

https://github.com/user-attachments/assets/2dbc8986-b294-436f-b816-14ef5c87385f

> did we lose the flamegraph segment that showed threads that are not tokio workers?

After pulling main it seems the demo trace in main is not including those samples (already missing in main), this works for my other traces. I'm looking into that

rcoh (approved):

PR #275, opened 2026-04-23
Make worker lanes scrollable in viewer

rcoh:
Plain scroll wheel now scrolls worker lanes vertically instead of zooming. Ctrl/Cmd+scroll zooms the timeline. This allows viewing traces with many workers (e.g. 48) without lanes being clipped.

Also adds a many_workers example that generates a 48-worker trace for testing.

### DIFF
--- a/dial9-viewer/ui/viewer.html
+++ b/dial9-viewer/ui/viewer.html
@@ -551,7 +551,8 @@
             <div id="help-content">
                 <h3>🖱 Mouse</h3>
                 <table>
-                    <tr><td>Scroll wheel</td><td>Zoom in/out (centered on cursor)</td></tr>
+                    <tr><td>Scroll wheel</td><td>Scroll worker lanes vertically</td></tr>
+                    <tr><td>⌘/Ctrl + scroll</td><td>Zoom in/out (centered on cursor)</td></tr>
                     <tr><td>Drag</td><td>Pan left/right</td></tr>
                     <tr><td>Click poll</td><td>Select task &amp; show wake timeline</td></tr>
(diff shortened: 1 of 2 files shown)


dial9-viewer/ui/viewer.html
```diff
@@ -609,7 +609,8 @@
             <div id="help-content">
                 <h3>🖱 Mouse</h3>
                 <table>
-                    <tr><td>Scroll wheel</td><td>Zoom in/out (centered on cursor)</td></tr>
+                    <tr><td>Scroll wheel</td><td>Scroll worker lanes vertically</td></tr>
+                    <tr><td>⌘/Ctrl + scroll</td><td>Zoom in/out (centered on cursor)</td></tr>
```
Fluzko, line 613: This works fine, but it overlaps the time bar behind. Now that we've got two of them seems like something won't scale much if more are needed. No big deal, we can review the UI layout in a future

Fluzko (approved):

dial9-viewer/ui/viewer.html
```diff
@@ -2729,22 +2730,30 @@ <h3>⌨ Keyboard</h3>
                 renderAll();
             }
 
-            // Mouse wheel zoom on main area
+            // Mouse wheel: Ctrl/Cmd+wheel zooms timeline, plain wheel scrolls worker lanes
             document.getElementById("main-area").addEventListener(
                 "wheel",
                 (e) => {
                     if (document.getElementById("flamegraph-panel").style.display === "flex") return;
```
jlizen, line 2737: minor: This isn't new, but the lack of `preventDefault()` means that the scroll will happen behind the timeline which is probably not desirable

jlizen (approved):

PR #276, opened 2026-04-23
Avoid traversing entire directory on local search

rcoh:


dial9-viewer/src/storage.rs
```diff
@@ -313,18 +313,47 @@ impl StorageBackend for LocalBackend {
...
+        return Ok(());
+    }
     let entries = match std::fs::read_dir(dir) {
         Ok(e) => e,
         Err(e) if e.kind() == std::io::ErrorKind::NotFound => return Ok(()),
+        Err(e) if e.kind() == std::io::ErrorKind::PermissionDenied => return Ok(()),
         Err(e) => return Err(StorageError::Other(e.to_string())),
     };
     for entry in entries {
+        *visited += 1;
```
jlizen, line 355: nit: this means we only actually are allowing `MAX_ENTRIES_VISITED` - 1

dial9-viewer/src/storage.rs
```diff
@@ -313,18 +313,47 @@ impl StorageBackend for LocalBackend {
...
+    if depth > MAX_COLLECT_DEPTH
+        || out.len() >= MAX_COLLECT_FILES
+        || *visited >= MAX_ENTRIES_VISITED
+    {
+        return Ok(());
+    }
     let entries = match std::fs::read_dir(dir) {
         Ok(e) => e,
         Err(e) if e.kind() == std::io::ErrorKind::NotFound => return Ok(()),
+        Err(e) if e.kind() == std::io::ErrorKind::PermissionDenied => return Ok(()),
```
jlizen: Probably should have some user-facing output on this case

jlizen (approved):
Seems good. I'd suggest adding some user-facing outputs rather than silently truncating (and similar for permissions errors).

PR #280, opened 2026-04-23
Test/improve testing cpu profiling fallback

yulnr:
Follow-up to #250 (new CPU profiling fallback), this PR does some minor clean up and adds a few tests:

- Tests around the new pieces: sigsegv trap path, and ctimer sample firing under cpu load. 
- Adds fallback path to the bencher overhead tracking.
- One regression fix: `open_perf_event` was wrapping `PermissionDenied` errors via `io::Error::new`, which dropped `raw_os_error` and broke `is_perf_blocked`. Added a fallback detection [test to catch this](perf-self-profile/tests/fallback_detection.rs). 

Will follow up with another one improving the ECS simulation, current check is probably not enough or exactly what ECS restricts.

### DIFF
--- a/perf-self-profile/src/sys/linux/fp_profiler/ctimer.rs
+++ b/perf-self-profile/src/sys/linux/fp_profiler/ctimer.rs
@@ -12,6 +12,16 @@
 //!
 //! ctimer avoids both by binding each timer to a specific tid (`SIGEV_THREAD_ID`)
 //! and charging against per-thread CPU time.
+//!
+//! # Lifecycle                                                                                                                                                                                       
+//!                                                                                   
+//! 1. Call `start(interval_ns)` once in the main thread to install the SIGPROF handler.
+//! 2. Each thread calls `register_thread` from its own context to arm a per-thread
+//! timer, and `unregister_thread` (same thread) to tear it down.
+//! 3. `disable` and `enable` pause sampling without disarming the timers, so resuming doesn't
+//! require re-registration.
+//! Teardown: `disarm_all_timers` disarms timers globally (called automatically on `CtimerSampler` drop)
+//! A subsequent `start` is required to sample again.
 
 use std::cell::Cell;
 use std::io;
--- a/perf-self-profile/src/sys/linux/fp_profiler/ctimer.rs
+++ b/perf-self-profile/src/sys/linux/fp_profiler/ctimer.rs
@@ -12,19 +12,6 @@
 //!
 //! ctimer avoids both by binding each timer to a specific tid (`SIGEV_THREAD_ID`)
 //! and charging against per-thread CPU time.
-//!
-//! # Lifecycle
-//!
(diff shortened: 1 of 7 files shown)


perf-self-profile/src/sys/linux/fp_profiler/ctimer.rs
```diff
@@ -12,19 +12,6 @@
 //!
 //! ctimer avoids both by binding each timer to a specific tid (`SIGEV_THREAD_ID`)
 //! and charging against per-thread CPU time.
-//!
```
rcoh, line 15: did you mean to delete ths?

perf-self-profile/src/sys/linux/fp_profiler/ctimer.rs
```diff
@@ -234,4 +229,56 @@ mod tests {
...
+        let count = SAMPLE_COUNT.load(Ordering::Relaxed);
+
+        unregister_thread();
+
+        RUNNING.store(false, Ordering::Release);
+        DISARM_REQUESTED.store(false, Ordering::Release);
+
+        // lower threshold to account for noise
+        assert!(
+            count >= 150,
```
rcoh, line 290: this might be not low enough but we'll find out when the stress test runs

perf-self-profile/src/sys/linux/fp_profiler/unwind.rs
```diff
@@ -245,4 +245,75 @@ mod tests {
...
+        let fp = guard as usize;
+
+        // sp = fp keeps fp in range, page-aligned.
+        let mut out = [0u64; MAX_FRAMES];
+        let n = unsafe { unwind(0x40_0000, fp, fp, &mut out) };
+
+        assert_eq!(n, 1, "unwind must stop when saved_fp load faults");
+        assert_eq!(out[0], 0x40_0000);
+
+        assert_eq!(unsafe { libc::munmap(guard, ps) }, 0);
```
rcoh: if the test fail before this does this leave things in a weird state?

rcoh (approved):

perf-self-profile/src/sys/linux/fp_profiler/ctimer.rs
jv1i: yeah I thought it was just duplicating what the docs of each module and method say already, but can add it back summarized

perf-self-profile/src/sys/linux/fp_profiler/ctimer.rs
jv1i: I was actually surprised that locally I was getting > 195 every time, I'll monitor it and lower it

perf-self-profile/src/sys/linux/fp_profiler/unwind.rs
jv1i: re-ordered so unmap runs before asserts, so at least the assert panics wouldn't leak

PR #289, opened 2026-04-24
perf: use FxHash for intern_location

### LINKED ISSUE #245: intern_location should use FxHash
spotted it using SipHasher in a flamegraph

namn-grg:
closes #245 

Swap `std::collections::HashMap` for `FxHashMap` (reusing the in-tree `FxHasher` already defined in `dial9-trace-format`)

### DIFF
--- a/dial9-trace-format/src/encoder.rs
+++ b/dial9-trace-format/src/encoder.rs
@@ -18,6 +18,7 @@ use std::sync::Arc;
 ///
 /// For HashMap keys that are already well-distributed (TypeId, Arc<str>), this
 /// avoids hash collisions.
+#[doc(hidden)]
 #[derive(Default)]
 pub struct FxHasher(u64);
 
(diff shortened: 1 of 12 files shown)


dial9-trace-format/src/encoder.rs
```diff
@@ -19,7 +19,7 @@ use std::sync::Arc;
 /// For HashMap keys that are already well-distributed (TypeId, Arc<str>), this
 /// avoids hash collisions.
 #[derive(Default)]
-pub(crate) struct FxHasher(u64);
+pub struct FxHasher(u64);
```
rcoh: Can you make these doc(hidden)? We don't actually want to expose them publicly

rcoh (approved):

rcoh:
Thanks!

rcoh:
CI issue isn't your fault

jlizen:
I can fix CI, will push a change on top of this. We will need two things:
- switch from a loop of publishing each package, to a multi `-p` command (which will use the local registry overlay provided the newer versions are specified explicitly)
- trigger a fresh release which will bump the versions

I've validated that switching to multi `-p` would still catch the regression that this red CI step is guarding against (in #259 )

PR #290, opened 2026-04-24
feat(viewer): resizable sidebar and slightly improved ux

zz85:
- Add draggable resize handle on the sidebar left edge (200px-70vw)
- Show stack trace frames expanded by default instead of collapsed
- Change cursor to pointer when hovering polls with CPU/sched data (so you know it's clickable)
- Wrap toolbar rows instead of hiding overflow
- Add ? keyboard shortcut to toggle the help overlay

### DIFF
--- a/dial9-viewer/ui/viewer.html
+++ b/dial9-viewer/ui/viewer.html
@@ -3210,6 +3210,7 @@
                 let resizing = false;
                 let startX = 0;
                 let startW = 0;
+                let resizeRafId = 0;
 
                 handle.addEventListener("mousedown", (e) => {
                     e.preventDefault();
@@ -3229,6 +3230,12 @@
                         window.innerWidth * 0.7
                     );
                     sidebar.style.width = newW + "px";
+                    if (trace && !resizeRafId) {
+                        resizeRafId = requestAnimationFrame(() => {
+                            resizeRafId = 0;
+                            renderAll();
+                        });
+                    }
                 });
 
                 window.addEventListener("mouseup", () => {
@@ -3526,6 +3533,7 @@
                     }
                     case "?":
                         e.preventDefault();
+                        if (kbSelecting) cancelKbSelection();
                         document.getElementById("help-overlay").classList.toggle("visible");
                         break;
                 }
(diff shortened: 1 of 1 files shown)


dial9-viewer/ui/viewer.html
```diff
@@ -3470,6 +3524,10 @@ <h3>⌨ Keyboard</h3>
                         renderAll();
                         break;
                     }
+                    case "?":
+                        e.preventDefault();
+                        document.getElementById("help-overlay").classList.toggle("visible");
+                        break;
```
jv1i, line 3538: minor: `?` shortcut is clashing with Shift/Alt selection mode. If I press `shift+/` the shift keydown fires first and enters `kbSelecting` mode before `?` toggles help. So then we end up with help open and in selection mode, need to hit `Escape` twice to close the help overlay. (I think we can just cancel `kbSelecting` inside this case)

dial9-viewer/ui/viewer.html
```diff
@@ -3187,6 +3203,44 @@ <h3>⌨ Keyboard</h3>
...
+                });
+
+                window.addEventListener("mouseup", () => {
+                    if (!resizing) return;
+                    resizing = false;
+                    handle.classList.remove("active");
+                    document.body.style.cursor = "";
+                    document.body.style.userSelect = "";
+                    if (trace) requestAnimationFrame(renderAll);
+                });
```
jv1i, line 3248: minor: should we move the `requestAnimationFrame(renderAll)` call to `mousemove` so that timeline re-renders as the sidebar resizes, instead of freezing in place and then jumping? we could throttle it like we do for the tooltip crosshair with a `rafId` flag.

yulnr (commented):
Thanks! this looks good to me, saves some clicking around.
Leaving some minor comments.

dial9-viewer/ui/viewer.html
zz85: good call, mousemove should rerender periodically, but onmouseup will do a final paint

dial9-viewer/ui/viewer.html
zz85: ah good catch, I didn't notice that because I was using `?` to toggle on/off. cancelling  `kbSelecting` works.

yulnr (approved):
Thanks for the changes!

PR #293, opened 2026-04-24
feat: detect uninstrumented task spawns and surface in viewer

yulnr:
## Summary

Adds an `instrumented` flag to tasks spawn events to track wether they were wrapped by `TelemetryHandle::spawn` or not.

When task tracking is enabled, the `on_task_spawn` hook now checks a thread-local flag to classify each spawn as instrumented (`TelemetryHandle::spawn` / `RuntimeTelemetryHandle::spawn`) or raw (`tokio::spawn`). The flag is carried through the wire format as `Option<bool>` on `TaskSpawnEvent` (no version bump - old traces default to instrumented).

**Shutdown summary:** `TelemetryGuard::Drop` now logs a per-location breakdown via `tracing::debug` (visible with `RUST_LOG=dial9_telemetry=debug`, silent otherwise). Example:

image

**Viewer:** task detail panel shows a clickable "no wake data" badge (links to `TelemetryHandle::spawn` docs) for uninstrumented tasks, and hides the previous misleading "0 wakes" counter. 

https://github.com/user-attachments/assets/a96bf39f-c813-4a70-86d7-618d7d5874f9


dial9-tokio-telemetry/src/telemetry/recorder/shared_state.rs
```diff
@@ -43,6 +45,9 @@ pub(crate) struct SharedState {
     /// cycle for queue sampling and metadata generation. `build_and_attach_to_telemetry`
     /// pushes new contexts here so the flush thread picks them up.
     pub(crate) contexts: Mutex<Vec<Arc<RuntimeContext>>>,
+    /// Spawn locations of uninstrumented spawns, appended lock-free at
+    /// runtime and drained at shutdown to build the per-location summary.
+    pub(crate) uninstrumented_spawn_locs: SegQueue<&'static Location<'static>>,
```
jv1i: SeqQueue here means we have to iterate over these at shutdown (as opposed to using a map), but I wanted to avoid any locking structure during spawn.

yulnr:
To showcase this further in the viewer, this adds a button that popups a summary, and also a filter alongside the existing ones: 58785f0

Comitted it separately in case we think this is too noisy. 

https://github.com/user-attachments/assets/6751a29d-012b-49ae-a69c-3ca693d4d818

Open to ideas about the user-facing language for this, "uninstrumented" feels correct to me, but perhaps something like "blind spawns" or "missing wakes" reads better.

dial9-tokio-telemetry/src/telemetry/format.rs
```diff
@@ -176,6 +176,9 @@ pub struct TaskSpawnEvent {
     pub task_id: TaskId,
     /// Interned spawn location.
     pub spawn_loc: InternedString,
+    /// Whether this spawn was instrumented (via `TelemetryHandle::spawn`).
+    /// Defaults to true if not present on the wire.
+    pub instrumented: Option<bool>,
```
rcoh: Why option?

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -185,11 +192,18 @@ fn register_hooks(
...
+
             s5.record_event(RawEvent::TaskSpawn {
                 timestamp_nanos: crate::telemetry::events::clock_monotonic_ns(),
                 task_id,
                 location,
+                instrumented,
             });
+
+            if !instrumented {
+                s5.uninstrumented_spawn_locs.push(location);
```
rcoh: Isn't this going to create a list with an unbounded number of items?

dial9-tokio-telemetry/src/telemetry/format.rs
jv1i: Backward compat with older traces, should we just bump the decoder version instead? wasn't sure if we're ok with that for this feature

to clarify: initially thought missing fields on older traces meant the events would be dropped, but testing now I see the events rather don't decode into their type and get demoted to a custom event (side-note, we might want to protect against that)

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
jv1i: oh crap that's true. I'll think of something, perhaps I do just push to a map behind a mutex, or cap the segqueue (the exact final number of spawns is not the most important, but rather the locations)

jv1i: actually, if we are ok with showing this in the viewer (see my comment above with the screen cap.) then we could also just drop this "report on shutdown", it was my original feature when I thought we would keep this just for local dev (we don't need this twice). are we ok with the additions to the viewer?

rcoh:
Let's fully drop the in-app tracking. We can build it into the viewer and
the analysis toolkit

dial9-tokio-telemetry/src/telemetry/events.rs
```diff
@@ -133,6 +133,8 @@ pub enum TelemetryEvent {
         task_id: TaskId,
         /// Interned spawn location of the task.
         spawn_loc: InternedString,
+        /// Whether this task was spawned via [`TelemetryHandle::spawn`]
+        instrumented: bool,
```
rcoh: this one actually needs to be optional

dial9-tokio-telemetry/src/telemetry/format.rs
rcoh: this one can be required

rcoh (approved):
lgtm. optional is also fine, whatever causes less future headaches :-)

PR #298, opened 2026-04-25
[dial9-viewer] Toolkit: parallel multi-file trace analysis with caching

jlizen:
### Summary

Trace analysis was bottlenecked on a single core when analyzing load test runs with hundreds or thousands of trace files. This adds directory support to the toolkit with parallel parsing, NDJSON caching, and parallel analysis.

Two APIs:

- `parseTrace(path)` returns an async iterable of `ParsedTrace` for both files and directories. Same `for await` pattern for both. For directories, parsing runs in parallel subprocesses and results are cached to `.d9-cache/`.
- `analyzeTraces(path)` returns aggregated results across all files with parallel analysis. Two phases: parse workers populate the cache, then accumulate workers read cached data and compute partial results that the main process merges.

The default report now includes poll duration histograms by spawn location, span duration percentiles (from `#[instrument]`), and scheduling delay histograms, in addition to the existing worker utilization, long polls, task lifecycle, blocking calls, CPU hotspots, and queue depth sections.

### Design decisions

**Subprocess parallelism**: Each file is parsed in its own `child_process.execFile` subprocess. Simple to implement, no shared state concerns, and each process gets its own V8 heap for large traces.

**NDJSON cache format**: Raw `ParsedTrace` data written one JSON record per line. Avoids V8's ~512MB string limit on both write (streaming) and read (line-by-line from Buffer). Tradeoff: cache files are large and somewhat slow to parse, especially for large traces. Future work replaces this with SQLite/Parquet (see #299 ).

**Two-phase parallel streaming analysis**: Parse workers and accumulate workers run in separate phases (not overlapped) to avoid doubling memory usage. The accumulate workers read cached NDJSON, run the full analysis pipeline, and return partial accumulators via stdout. The main process merges partials in constant memory. We could probably make the two phases blend together as parsing work runs out, but I opted to keep things simple.

**Native HDR histograms**: Uses Node's `createHistogram()` for scheduling delays, poll durations by spawn location, and span durations. Exact percentiles with bounded memory.

**Unified API for single and multi-file**: `parseTrace` returns an async iterable for both single files and directories, so recipes work unchanged regardless of input. `for await (const trace of parseTrace(path))` iterates once for a single file, N times for a directory.

### Timings (192 core machine)

| Scenario | Wall | Input files (compressed) | Per input file (avg/max) | Total input size | CPU% | Cache |
|----------|------|-------|--------------------|-------|------|-------|
| Single demo | 0.58s | 1 | 1.3 MB | 1.3 MB | 166% | n/a |
| Small cold | 45.8s | 2071 | 499 KB / 1.0 MB | 1.0 GB | 4408% | 28 GB |
| Small warm | 22.0s | 2071 | 499 KB / 1.0 MB | 1.0 GB | 3999% | 28 GB |
| Small sample 100 | 2.8s | 100 | 499 KB / 1.0 MB | ~49 MB | 3314% | |
| Large cold | 1:39 | 8 | 19.8 MB / 34.8 MB | 159 MB | 701% | 4.2 GB |
| Large warm | 1:29 | 8 | 19.8 MB / 34.8 MB | 159 MB | 557% | |

Large files are bottlenecked by the JS WASM decoder (~90s per 100MB trace). With proper rotation (production default), files stay small and the toolkit handles 2000+ files in under a minute.

### Recipe timings (warm cache, small files, 192 core machine)

| Scenario | Wall | CPU% | Notes |
|----------|------|------|-------|
| `analyzeTraces` (2071 files) | 21.2s | 4141% | parallel analysis |
| `analyzeTraces` sample 100 | 1.3s | 3112% | 2.9M events |
| `for await` worst poll (2071 files) | 2:29 | 134% | sequential per-file |
| `for await` single file | 0.33s | 155% | baseline |

Future scope:
- #299 - nicer query layer to remove the sequential recipes and make bigger traces quicker, probably backed with sqlite or parquet
- #300 - better support for handling full prefixes from s3

### Testing

`test_directory_parse.js`: directory iteration, caching (create, hit, invalidation, force, disable), sampling, progress, parallel/sequential, atomic writes, full analysis pipeline on cached data.

`test_all_skills_snippets.js`: extracts every JavaScript code block from all 6 skills markdown files and runs them against both a single file and a directory. Includes a schema sync test that parses the documented `analyzeTraces` and `ParsedTrace` return keys from `analysis.md` and `loading.md` and asserts they match the actual result bidirectionally.

`test_directory_scale.js`: validates at scale (20 files default, 200 with `--large`).

Agent evals: ran the toolkit skills through an LLM agent to validate that the documentation is sufficient for an agent to use the toolkit end-to-end without prior knowledge. Here is a log, starting with the prompt, showing usage with a large directory: [gist](https://gist.github.com/jlizen/cb4fdaad23bc6a32c0d2d86040c3b9a7). Single-file still works too.

Viewer: made sure the viewer still loads the demo, shows the progress indicator, etc. I didn't directly test with multi-file or s3, but I didn't change any of those code paths.


rcoh:
This approach seems fine for now.

Instead of caching in JSON I think we should actually expose a REPL over the parsed data that agents can execute code against in the long run.

rcoh (approved):
While we are here it would be nice to support allowing inline chunks to be gzipped which would allow the backend to avoid having to decompress when multiple files are loaded. Fine as follow on

jlizen:
> While we are here it would be nice to support allowing inline chunks to be gzipped which would allow the backend to avoid having to decompress when multiple files are loaded. Fine as follow on

Opened #307 about this.

I'll also note the REPL use case in #299 , makes sense

PR #304, opened 2026-04-26
viewer: Retain selection overlay while sidebar is open

zz85:
Keep the blue region highlight visible on the timeline when the sidebar shows flamegraph or blocking calls, so users can see which region they selected. The overlay tracks pan/zoom and is cleared when the sidebar closes. Shift/Alt keypress no longer resets the overlay while a sidebar selection is active.

SCR-20260425-rkgx


rcoh:
Nice! This is a good step towards multi selection/ comparison

rcoh (approved):

PR #306, opened 2026-04-26
Add simple example for local execution

mox692:
I thought it would be nice if we have an example that doesn't depend on AWS services. (From what I understand, the existing example `metrics-service` requires some aws accesses.)
The example itself is fairly simple, it just spawns some tasks and wait until they complete.

This is the tracing result that I got on my local machine from this example.
Screenshot 2026-04-26 at 14 10 59


rcoh (approved):
There a couple of similar examples but nice to have another top level one. Clearly much more discoverable

PR #308, opened 2026-04-27
fix(viewer): correct schedWait unit from µs to ns

olivergillespie:
The viewer tooltip and park visualization multiplied schedWait by 1000, assuming it was in microseconds. It has always been nanoseconds (from sched_wait_ns on the wire). This caused a 1000x overreport of kernel scheduling delays.


jlizen (approved):
Thanks for the fix. Unfortunate that we don't have a great way of testing this, but we'll rewrite the whole viewer sooner or later.

PR #311, opened 2026-04-28
feat: Store S3 metadata into segement metadata

rcoh:
- Make segment metadata writing lazy so late-bound items will still be persisted into segment metadata
- update sgement metadata now merges entries
- copy s3 config options into segment metadata

### DIFF
--- a/dial9-tokio-telemetry/src/background_task/s3.rs
+++ b/dial9-tokio-telemetry/src/background_task/s3.rs
@@ -125,6 +125,17 @@ impl S3Config {
         &self.bucket
     }
 
+    pub(crate) fn as_metadata(&self) -> impl Iterator<Item = (&str, &str)> {
+        [
+            ("bucket", self.bucket.as_str()),
+            ("service_name", self.service_name.as_str()),
+            ("instance_path", self.instance_path.as_str()),
+            ("boot_id", self.boot_id.as_str()),
+        ]
+        .into_iter()
+        .chain(self.prefix.as_ref().map(|p| ("prefix", p.as_str())))
+    }
+
     /// Optional region override for the S3 client.
     pub(crate) fn region(&self) -> Option<&str> {
         self.region.as_deref()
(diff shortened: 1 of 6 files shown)


dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -1842,4 +1891,139 @@ mod tests {
...
+        writer.update_segment_metadata(vec![
+            ("bucket".into(), "my-bucket".into()),
+            ("service_name".into(), "my-svc".into()),
+        ]);
+
+        // Step 2: flush loop merges static + runtime entries (like recorder/mod.rs)
+        let static_metadata = writer.segment_metadata().to_vec();
+        let mut merged = static_metadata.clone();
+        merged.push(("runtime.main".into(), "0,1".into()));
+        writer.update_segment_metadata(merged);
```
jlizen:
```suggestion
          writer.update_segment_metadata(vec![("runtime.main".into(), "0,1".into())]);
```

Right now we are merging s3 + tokio on top of only s3.

We can validate preservation if we remove s3 from the incoming batch.

dial9-tokio-telemetry/src/background_task/s3.rs
```diff
@@ -125,6 +125,17 @@ impl S3Config {
         &self.bucket
     }
 
+    pub(crate) fn as_metadata(&self) -> impl Iterator<Item = (&str, &str)> {
+        [
+            ("bucket", self.bucket.as_str()),
+            ("service_name", self.service_name.as_str()),
+            ("instance_path", self.instance_path.as_str()),
+            ("boot_id", self.boot_id.as_str()),
```
jlizen, line 133: Should we pass region?

jlizen (approved):
Seems nice, we should think about how to let customers attach arbitrary metadata too. Ideally I would think something in runtime config, with a method on the handle to modify.

Anyway for @Fluzko 's work this will be useful

rcoh:
there is an API on RotatingWriter to set your own metadata but agreed its not the easiest

PR #313, opened 2026-04-28
Add shuttle tests of shared_state and local buffers

rcoh:


dial9-tokio-telemetry/src/primitives.rs
```diff
@@ -0,0 +1,188 @@
...
+        unsafe impl<T: Send> Send for Receiver<T> {}
+
+        impl<T> Receiver<T> {
+            pub(crate) fn recv_timeout(
+                &self,
+                _timeout: std::time::Duration,
+            ) -> Result<T, RecvTimeoutError> {
+                // Randomly decide whether to simulate a timeout, giving
+                // the flush loop a chance to execute its body.
+                if shuttle::rand::thread_rng().gen_bool(0.5) {
```
jlizen: Consider raising this to more like `0.8` to get more useful interleavings... our tests will still make progress on interleavings, but 50% chance of blocking means we won't get as many cases of multiple flushes as we would like.

dial9-tokio-telemetry/src/telemetry/recorder/shared_state.rs
```diff
@@ -417,3 +416,257 @@ mod tests {
...
+                Ok(false)
+            }
+        }
+    }
+
+    // ── Decoding ────────────────────────────────────────────────────────
+
+    fn decode_validation_events(data: &[u8]) -> Vec<ValidationEvent> {
+        use dial9_trace_format::decoder::Decoder;
+        let Some(mut dec) = Decoder::new(data) else {
```
jlizen, line 505:
```suggestion
        let Some(mut dec) = Decoder::new(data) else {
              assert!(data.is_empty(), "non-empty segment failed to decode");
```

dial9-tokio-telemetry/src/primitives.rs
```diff
@@ -0,0 +1,188 @@
+//! Cfg-gated concurrency primitives.
+//!
+//! Under normal compilation this re-exports from `std`. With `--cfg shuttle`
```
jlizen, line 3: Should we add clippy disallowed methods to make sure we use these throughout the codebase? Or are we intentionally only using them in some places?

dial9-tokio-telemetry/src/telemetry/recorder/shared_state.rs
```diff
@@ -417,3 +416,257 @@ mod tests {
...
+                ev.id, ev.timestamp_ns
+            );
+        }
+    }
+
+    /// Per-thread events within each segment are in sequence order, and
+    /// segments are temporally ordered per-thread: if thread T's event
+    /// with seq=S is in segment K, no event from thread T with seq < S
+    /// may appear in segment K+1 or later.
+    fn check_segments_temporally_ordered(segments: &[Vec<ValidationEvent>]) {
```
jlizen: This invariant is complex enough that it might be worth adding a negative test for to "test the tester", eg:
```
  #[test]
  #[should_panic(expected = "event leaked across rotation")]
  fn cross_segment_leak_detected() {
      let seg0 = vec![
          ValidationEvent { timestamp_ns: 1, thread_id: 0, seq: 0, id: 0 },
          ValidationEvent { timestamp_ns: 2, thread_id: 0, seq: 2, id: 2 },
      ];
      let seg1 = vec![
          // seq=1 appears AFTER seg0 had seq=2 for the same thread
          ValidationEvent { timestamp_ns: 3, thread_id: 0, seq: 1, id: 1 },
      ];
      check_segments_temporally_ordered(&[seg0, seg1]);
  }
```

We could do similar for the other invariants, but they are a lot simpler.

.github/workflows/ci.yml
```diff
@@ -166,6 +166,23 @@ jobs:
...
+    runs-on: ubuntu-latest
+    env:
+      RUST_BACKTRACE: 1
+      RUSTFLAGS: "--cfg tokio_unstable --cfg shuttle"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.STABLE_TOOLCHAIN }}
+      - uses: Swatinem/rust-cache@v2
```
jlizen, line 180: We probably should pin commit hashes, but anyway that's a pre-existing thing, I'll cut a separate issue.

jlizen (approved):

dial9-tokio-telemetry/src/primitives.rs
rcoh: we are intentionally slowly moving them outwards for the moment

dial9-tokio-telemetry/src/telemetry/recorder/shared_state.rs
rcoh: I ended up just dropping this invariant. This invariant is very nuanced and this invariant as written isn't actually strong enough to be useful. I will write another more sophisticated one in a follow up

PR #316, opened 2026-04-29
fix: Don't register sched events on blocking pool threads

### LINKED ISSUE #314: protect from file descriptor exhaustion when subscribing to perf events
Right now, sampling scheduling events opens a separate fd for each thread (which can exhaust fds). We need a hard limit to protect fds. A reasonable approach is only enabling sampling when we _know_ a thread has become a Tokio worker thread

rcoh:
fixes #314 

The combination of two bugs:
1. perf file descriptors were never closed
2. blocking threads opened their own perf fd

Caused a leak of perf file descriptors over time

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
+++ b/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
@@ -25,37 +25,6 @@ use std::sync::Arc;
 use std::sync::atomic::Ordering;
 use std::time::{Duration, Instant};
 
-/// Guard that cleans up per-thread telemetry state when dropped.
-///
-/// Stored in a thread-local so cleanup happens automatically when the thread
-/// exits — even if Tokio's `on_thread_stop` doesn't fire or panics before
-/// reaching cleanup code. This also covers the calling thread (which never
-/// gets an `on_thread_stop` callback).
-struct ThreadCleanupGuard {
-    shared: Arc<SharedState>,
-}
-
-impl Drop for ThreadCleanupGuard {
-    fn drop(&mut self) {
-        // Clear the per-thread handle so the Arc<SharedState> it holds is released.
-        CURRENT_HANDLE.with(|cell| {
-            *cell.borrow_mut() = None;
-        });
-
-        #[cfg(feature = "cpu-profiling")]
-        {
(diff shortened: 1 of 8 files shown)


dial9-tokio-telemetry/tests/sched_profiler_fd_leak.rs
```diff
@@ -0,0 +1,144 @@
...
+        let mut handles = Vec::new();
+        for _ in 0..num_blocking_tasks {
+            handles.push(tokio::task::spawn_blocking(|| {
+                std::thread::sleep(Duration::from_millis(50));
+            }));
+        }
+        for h in handles {
+            h.await.unwrap();
+        }
+        // Wait for threads to exit and on_thread_stop to fire.
```
jlizen, line 76: nit: this is a bit misleading since we never open fds in the first place for blocking threads, so there is no need to wait on cleanup.

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
jlizen, line 258: nit: this is taking the lock to tear down blocking threads that no longer need cleanup. Not sure if it is worth adding a `runtime_context::is_sched_tracked()` to check the thread-local state to avoid that. Probably not :)

jlizen (approved):

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -241,12 +241,10 @@ fn register_hooks(
...
                     .insert(tid, crate::telemetry::events::ThreadRole::Blocking);
-                if let Ok(mut prof) = s_start.sched_profiler.lock()
-                    && let Some(ref mut p) = *prof
-                {
-                    // Register the current thread for sched event sampling.
-                    let _ = p.track_current_thread();
-                }
+                // Sched event sampling is deferred to register_tid_if_needed(),
+                // which runs only for worker threads on their first poll/park.
+                // This avoids opening perf fds for blocking pool threads.
```
jlizen, line 246: Just to note, we still do have coverage of blocking threads via the cpu profiler's inherit approach. Was my main question while reading this.

PR #318, opened 2026-04-29
Intern stack frames for smaller traces

### LINKED ISSUE #222: Add interning for stack frames
Especially with scheduling events, there are a huge number of duplicates. There are often only hundreds of unique stacks. This compresses very well with GZIP but means that the on-disk traces can get quite large before compression

yulnr:
closes #222 

`CpuSample` currently writes the full stack on every sample, which dominates trace size at hundreds-of-Hz sampling. 
This PR adds interning of stack frames to mitigate this.

## Summary

Adds a `PooledStackFrames` field type alongside the existing `StackFrames` and switches `CpuSample.callchain` to use it, the encoder dedups stacks into a per-segment pool, mostly follows the existing approach interned strings.

Based on testing across 10:1 to 1000:1 samples-per-unique-stack ratios, interned output is 6–16x smaller than inline pre-gzip and ~3x smaller post-gzip.

I'm including a small improvement to `FxHasher::write` in 970aa66628c52e67e60ed2336562330d703e2005:

After adding interning, encoding with `FxHasher` (following what we do for interned strings) was 2-5x slower than inline. Default `HashMap` was _alright_ , but still ended up ~60% slower than inline. 
We could just use the default `HashMap`, but I was surprised that `FxHasher` turned out to be *worse* than default. 
Our local `write` impl looped byte-by-byte, so the new `Box<[u64]>` stack-pool keys paid 8x more ops than necessary on every intern. Making `write` word-wise (chunks of 8 bytes via `from_ne_bytes`) restores parity with inline, and beats it when pool IDs are smaller than the inline frames. Other FxHashMap uses (string_pool, schema name lookups) should get a speed up too on keys ≥8 bytes (though those paths are not as hot really).

Encode-time impact, measured against the inline (pre-interning) baseline:

| bench (encode)        | inline   |  interned + byte-wise | interned + word-wise |
|-----------------------|----------|----------------------|----------------------|
| codec_1M_events       | 22.85ms  | 50.68ms  (+122%)     | 20.20ms  (-12%)      |
| stack_intern 100u/10k | 745us     | 2.55ms   (+242%)     | 587us    (-21%)      |
| stack_intern 500u/10k | 669us    | 2.81ms   (+320%)     | 632us    (-6%)       |
| stack_intern 1000u/10k| 669us    | 3.04ms   (+354%)     | 719us    (+7%)       |

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/format.rs
+++ b/dial9-tokio-telemetry/src/telemetry/format.rs
@@ -412,8 +412,8 @@ pub(crate) fn to_owned_event(
             source: e.source,
             callchain: stack_pool
                 .get(e.callchain)
-                .map(|s| s.to_vec())
-                .unwrap_or_default(),
+                .expect("stack pool entry must exist for CpuSample callchain")
+                .to_vec(),
         },
         TelemetryEventRef::WakeEvent(e) => TelemetryEvent::WakeEvent {
             timestamp_nanos: e.timestamp_ns,
--- a/dial9-tokio-telemetry/src/telemetry/format.rs
+++ b/dial9-tokio-telemetry/src/telemetry/format.rs
@@ -3,11 +3,11 @@ use crate::telemetry::events::CpuSampleSource;
 use crate::telemetry::events::TelemetryEvent;
 use crate::telemetry::task_metadata::TaskId;
 #[cfg(any(feature = "analysis", test))]
-use dial9_trace_format::decoder::StringPool;
+use dial9_trace_format::decoder::{StackPool, StringPool};
 #[cfg(any(feature = "analysis", test))]
 use dial9_trace_format::schema::SchemaEntry;
 use dial9_trace_format::types::{EventEncoder, FieldType, FieldValueRef};
-use dial9_trace_format::{InternedString, StackFrames, TraceEvent, TraceField};
+use dial9_trace_format::{InternedStackFrames, InternedString, TraceEvent, TraceField};
 use serde::Serialize;
 use std::fmt;
 use std::io::{self, Write};
@@ -195,7 +195,7 @@ pub(crate) struct CpuSampleEvent {
     pub tid: u32,
     pub source: CpuSampleSource,
     pub thread_name: Option<InternedString>,
-    pub callchain: StackFrames,
+    pub callchain: InternedStackFrames,
 }
 
 /// Wire-format event for a wake notification.
@@ -248,7 +248,7 @@ pub fn decode_events(data: &[u8]) -> io::Result<Vec<TelemetryEvent>> {
 
     dec.for_each_event(|ev| {
         if let Some(r) = decode_ref(ev.name, ev.timestamp_ns, ev.fields, ev.schema) {
-            events.push(to_owned_event(r, ev.string_pool));
+            events.push(to_owned_event(r, ev.string_pool, ev.stack_pool));
         }
     })
     .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e.to_string()))?;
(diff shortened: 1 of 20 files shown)


dial9-tokio-telemetry/src/telemetry/format.rs
```diff
@@ -406,7 +410,10 @@ pub(crate) fn to_owned_event(r: TelemetryEventRef<'_>, pool: &StringPool) -> Tel
                 .thread_name
                 .and_then(|s| pool.get(s).map(|n| n.to_string())),
             source: e.source,
-            callchain: e.callchain.iter().collect(),
+            callchain: stack_pool
+                .get(e.callchain)
+                .map(|s| s.to_vec())
+                .unwrap_or_default(),
```
rcoh: is unwrap_or_default() correct here? seems like that would hide a bug

dial9-trace-format/js/decode.js
```diff
@@ -7,11 +7,12 @@ const TAG_EVENT = 0x02;
 const TAG_STRING_POOL = 0x03;
 // Tag 0x04 is reserved (formerly SymbolTable, now a schema-based event).
 const TAG_TIMESTAMP_RESET = 0x05;
+const TAG_STACK_POOL = 0x07;
 
 const FieldType = {
   I64: 1, F64: 2, Bool: 3, String: 4,
   Bytes: 5, PooledString: 7, StackFrames: 8, Varint: 9,
-  StringMap: 10, U8: 11, U16: 12, U32: 13,
+  StringMap: 10, U8: 11, U16: 12, U32: 13, PooledStackFrames: 14,
```
rcoh: there is actually a whole at 6, it fits perfectly ther next to pooled string :-)

dial9-trace-format/src/codec.rs
```diff
@@ -79,6 +90,33 @@ pub struct PoolEntryRef<'a> {
...
+        (self.frames_le.len() / 8) as u32
+    }
+
+    /// Iterate the stack frame addresses.
+    pub fn iter(&self) -> crate::types::StackFrameIter<'a> {
+        crate::types::StackFrameIter::new(self.frames_le, self.frame_count())
+    }
+
+    /// Collect into an owned `Vec<u64>`.
+    pub fn to_vec(&self) -> Vec<u64> {
```
rcoh: or I suppose this could return the `StackFrame` type?

dial9-trace-format/src/encoder.rs
```diff
@@ -24,7 +26,12 @@ pub struct FxHasher(u64);
 
 impl Hasher for FxHasher {
     #[inline]
-    fn write(&mut self, bytes: &[u8]) {
+    fn write(&mut self, mut bytes: &[u8]) {
```
rcoh, line 36: there another optimization in the crate that may be worth doing:
https://docs.rs/fxhash/latest/src/fxhash/lib.rs.html#87-105

perf-self-profile/src/offline_symbolize.rs
```diff
@@ -66,14 +66,30 @@ pub fn symbolize_trace_with_maps(
     crate::sys::write_symbol_data(decoder, &addresses, maps, output)
 }
 
-fn collect_stack_frame_addresses(values: &[FieldValueRef<'_>], addresses: &mut BTreeSet<u64>) {
+fn collect_stack_frame_addresses(
+    values: &[FieldValueRef<'_>],
+    stack_pool: &StackPool,
+    addresses: &mut BTreeSet<u64>,
```
rcoh: side note, I don't there there is any actual reason to use a BTreeSet here, we should probably throw a todo in somewhere

rcoh (commented):
Awesome work on this!

rcoh:
a few more follow ups:
- [x] update the spec (SPEC.md)
- [x] add fuzzing of pooled stack frames

perf-self-profile/src/offline_symbolize.rs
jv1i: agree I don't see ordering being necessary, switched to a HashSet already

dial9-trace-format/src/codec.rs
jv1i: I was actually using the vec because I didn't need the type on the caller, refactored a bit now to actually use more the newtype everywhere

dial9-trace-format/js/decode.js
jv1i: moved to 6! Initially I skipped it because of the legacy timestamp comment, checking git history it seems that has been there since the first commit. I'm guessing no early internal users/traces we should worry about?

rcoh: Yeah no issues, 6 was never used in a released version

dial9-trace-format/SPEC.md
```diff
@@ -38,6 +38,8 @@ Every frame begins with a 1-byte tag:
 | `0x03` | String Pool |
 | `0x04` | *(reserved)* |
```
rcoh: seems natural to put StackPool here

rcoh (approved):

PR #328, opened 2026-04-30
fix: Don't leak schedstat file descriptors on thread exit

### LINKED ISSUE #319: on Linux, `SchedStat::read_current()` leaks one FD per thread
https://github.com/dial9-rs/dial9-tokio-telemetry/blob/8d61fb06b7c1f12352ad173812bf4b122b20c892/dial9-tokio-telemetry/src/telemetry/events.rs#L454-L465

Because this uses a RawFd and there is no drop guard, this fd stays open permanently

rcoh:
fixes #319

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/events.rs
+++ b/dial9-tokio-telemetry/src/telemetry/events.rs
@@ -1,7 +1,7 @@
 use crate::telemetry::{format::WorkerId, task_metadata::TaskId};
 use dial9_trace_format::{FieldValue, InternedString};
 use serde::Serialize;
-use std::{os::fd::RawFd, sync::Arc};
+use std::sync::Arc;
 
 /// Role of a thread known to the telemetry system.
 #[cfg(feature = "cpu-profiling")]
@@ -446,12 +446,12 @@ pub(crate) fn clock_pair() -> (u64, u64) {
 
 /// Per-thread scheduler stats from `/proc/<pid>/task/<tid>/schedstat`.
 /// Fields: run_time_ns wait_time_ns timeslices
-#[derive(Debug, Clone, Copy, Default)]
+#[derive(Debug, Clone, Copy)]
 pub(crate) struct SchedStat {
     pub wait_time_ns: u64,
     /// Raw fd backing this read, exposed for FD-lifecycle tests. Not used in production.
-    #[allow(dead_code)]
-    fd: RawFd,
+    #[cfg(test)]
+    fd: std::os::fd::RawFd,
 }
 
 #[cfg(target_os = "linux")]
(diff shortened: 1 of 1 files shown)


dial9-tokio-telemetry/src/telemetry/events.rs
```diff
@@ -457,19 +460,17 @@ impl SchedStat {
...
-                return fd;
+        let fd = SCHED_FD.with(|cell| -> std::io::Result<RawFd> {
+            if let Some(fd) = cell.borrow().as_ref() {
+                return Ok(fd.as_raw_fd());
             }
-            // First call on this thread: open the file
+            // First call on this thread: open the file.
             // SAFETY: SYS_gettid takes no arguments and always succeeds; unsafe is
             // required because syscall() is a raw FFI function with no type checking.
             let tid = unsafe { libc::syscall(libc::SYS_gettid) } as u32;
```
jlizen, line 476: scope creep: should we think about cutting over to `nix` at some point to have safe wrappers around these?

dial9-tokio-telemetry/src/telemetry/events.rs
```diff
@@ -449,6 +449,9 @@ pub(crate) fn clock_pair() -> (u64, u64) {
 #[derive(Debug, Clone, Copy, Default)]
 pub(crate) struct SchedStat {
     pub wait_time_ns: u64,
+    /// Raw fd backing this read, exposed for FD-lifecycle tests. Not used in production.
+    #[allow(dead_code)]
```
jlizen: Can this be `#[cfg(test)]` instead? It's a bit misleading to have this `RawFd` copy that might outlive its backing `OwnedFd`. For the test we want it to probe for a leak, but probably better not to have it available for non-test usage?

dial9-tokio-telemetry/src/telemetry/events.rs
```diff
@@ -449,6 +449,9 @@ pub(crate) fn clock_pair() -> (u64, u64) {
 #[derive(Debug, Clone, Copy, Default)]
```
jlizen: This will default to `0` (stdin) for the `RawFd`, which would be misleading. I'd suggest a custom default impl that defaults to `-1`.

dial9-tokio-telemetry/src/telemetry/events.rs
jlizen, line 502: pre-existing, but if pread returns 0 (doesn't seem likely for schedstat but could in theory), this wuld return a stale `last_os_error`. maybe add a defensive guard?
```
  if n < 0 {
      return Err(std::io::Error::last_os_error());
  }
  if n == 0 {
      return Err(std::io::Error::new(
          std::io::ErrorKind::UnexpectedEof,
          "empty schedstat",
      ));
  }
```

jlizen (approved):

PR #329, opened 2026-04-30
Catch panics in segment processor pipeline

rcoh:
A panicking SegmentProcessor previously killed the entire background worker thread. Now processor.process().await is wrapped with AssertUnwindSafe + catch_unwind so panics are caught, logged, and the segment is removed — the worker continues processing subsequent segments.

Changes:
- Wrap processor.process(data).await with catch_unwind in WorkerLoop::process_segments
- Add `std` feature to futures-util for FutureExt::catch_unwind
- Add test: processor_panic_does_not_kill_worker_loop
- Add test: processor_hang_respects_shutdown_timeout

### DIFF
--- a/dial9-tokio-telemetry/src/background_task/mod.rs
+++ b/dial9-tokio-telemetry/src/background_task/mod.rs
@@ -8,6 +8,7 @@ pub(crate) mod sealed;
 
 use crate::metrics::{Operation, SegmentProcessMetrics, SegmentProcessMetricsGuard};
 use crate::rate_limit::rate_limited;
+use futures_util::FutureExt;
 use metrique::timers::Timer;
 use metrique_writer::BoxEntrySink;
 use pipeline_metrics::{MetriqueResult, PipelineMetrics, StageMetrics};
@@ -599,7 +600,6 @@ impl WorkerLoop {
                 // AssertUnwindSafe: current processors are stateless or have
                 // trivially-recoverable state, so reuse after panic is safe.
                 let process_result = {
-                    use futures_util::FutureExt;
                     match std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
                         processor.process(data)
                     })) {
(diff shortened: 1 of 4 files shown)


dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -613,6 +636,45 @@ impl WorkerLoop {
                         }
                         continue 'next_segment;
                     }
+                    Err(panic_payload) => {
+                        let panic_msg = panic_payload
+                            .downcast_ref::<&str>()
+                            .copied()
+                            .or_else(|| panic_payload.downcast_ref::<String>().map(|s| s.as_str()))
+                            .unwrap_or("unknown panic");
+                        tracing::error!(
```
jlizen: Should we rate limit this?

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -613,6 +636,45 @@ impl WorkerLoop {
...
+                                panicked: true,
+                                panic_message: Some(panic_msg.to_owned()),
+                                pipeline: PipelineMetrics::default(),
+                            }
+                            .append_on_drop(self.metrics_sink.clone()),
+                        );
+                        if let Err(remove_err) = std::fs::remove_file(&segment.path)
+                            && remove_err.kind() != std::io::ErrorKind::NotFound
+                        {
+                            tracing::warn!(target: "dial9_worker", error = %remove_err, path = %segment.path.display(), "failed to remove segment after panic");
```
jlizen: Should we rate limit this?

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -613,6 +636,45 @@ impl WorkerLoop {
...
+                            .unwrap_or("unknown panic");
+                        tracing::error!(
+                            target: "dial9_worker",
+                            processor = processor.name(),
+                            segment = seg_idx + 1,
+                            path = %segment.path.display(),
+                            panic = panic_msg,
+                            "processor panicked, skipping segment"
+                        );
+                        // The original metrics guard was consumed by the
```
jlizen, line 656: This means the original segment process metric will drop and emit without a status, right? Probably not worth the trouble to avoid, just confirming my understanding

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -584,14 +594,27 @@ impl WorkerLoop {
...
                 let proc_start = std::time::Instant::now();
                 tracing::debug!(target: "dial9_worker", processor = processor.name(), segment = seg_idx + 1, "running processor");
-                match processor.process(data).await {
-                    Ok(next) => {
+                // Catch panics in both the synchronous `process()` call
+                // (which builds the future) and during `.await` (polling).
+                // AssertUnwindSafe: current processors are stateless or have
+                // trivially-recoverable state, so reuse after panic is safe.
+                let process_result = {
+                    use futures_util::FutureExt;
```
jlizen: nit: any reason not to do module level? there's no conditional compilation concern here right?

jlizen (approved):
The documented panic safety comment seems fine for now, I don't think we would want to recreate our own processor on a panic. 

But, any thoughts about a stronger guardrail once we have a publicly exposed pipeline system? It seems easy to mess up if you have eg some sort of batching behavior, which certain remote destinations might want?

dial9-tokio-telemetry/src/background_task/mod.rs
rcoh: yeah exactly. a little annoying. It would be interesting to add some sort of metrique level `PanicDetector` that could somehow detect somehow if it was dropped when a panic occured.

PR #332, opened 2026-04-30
fix: Avoid constructing events when telemetry is disabled

rcoh:
To help ensure we don't regress on this, I extracted an `if_enabled` wrapper that can be used as well


dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -383,6 +395,11 @@ impl TelemetryHandle {
         self.shared.enabled.store(false, Ordering::Relaxed);
     }
 
+    /// Returns whether telemetry is enabled
+    pub fn enabled() -> bool {
```
jlizen: Can we do `is_enabled()` instead for symmetry with `shared_state()` and (and also to avoid colliding with #256 's `enabled()` setter)?

There still will be some overlap with #256 's own `is_enabled()` but I suspect you would land first and they can rebase (but either ordering works fine)

jlizen (commented):
Change seems fine but some somewhat load-bearing bike shedding about naming since there is an overlap with #256 .

Also, was there any good way to test these events never land in the buffer? Not seeing anything obvious that wouldn't involve adding a lot more machinery, which I don't think is worth it, but just asking

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
rcoh: yeah is_enabled is fine

jlizen: Ok, you should be good to rebase on top of main and unify with #256 shortly, ping when ready for a new review

jlizen (approved):

PR #334, opened 2026-04-30
don't delete files during an s3 outage

### LINKED ISSUE #320: S3 upload failure will immediately delete the data
S3 PutObject failures return `ProcessErrorKind::Transfer(_)`. The circuit breaker returns `Io(Error::other("circuit breaker open"))`. 

Both paths fall into the delete branch. This means during an S3 outage, every segment arriving at the worker is silently destroyed instead of being kept on disk for retry.

We should wait for the writer to evict them instead of proactively deleting them in the background queue.

rcoh:
fixes #320

### DIFF
--- a/dial9-tokio-telemetry/src/background_task/mod.rs
+++ b/dial9-tokio-telemetry/src/background_task/mod.rs
@@ -13,6 +13,7 @@ use metrique_writer::BoxEntrySink;
 use pipeline_metrics::{MetriqueResult, PipelineMetrics, StageMetrics};
 use std::collections::HashMap;
 use std::future::Future;
+use std::io;
 use std::path::{Path, PathBuf};
 use std::pin::Pin;
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -598,9 +624,10 @@ impl WorkerLoop {
                         data.metrics.pipeline.push(processor.name(), stage);
                         data.metrics.status = Some(MetriqueResult::Failure);
                         data.metrics.total_time.stop();
-                        if matches!(&e.kind, ProcessErrorKind::Io(io) if io.kind() == std::io::ErrorKind::NotFound)
-                        {
+                        if e.kind.already_deleted() {
                             tracing::debug!(target: "dial9_worker", path = %segment.path.display(), "segment evicted during processing, skipping");
+                        } else if e.kind.retryable() {
+                            tracing::debug!(target: "dial9_worker", path = %segment.path.display(), err = ?e.kind, "retryable error, this file will be attempted to process again.")
```
jlizen:
```suggestion
                            tracing::debug!(target: "dial9_worker", path = %segment.path.display(), err = ?e.kind, "retryable error, this file will be attempted to process again.");
```

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -187,7 +203,17 @@ impl From<std::io::Error> for ProcessErrorKind {
 #[cfg(feature = "worker-s3")]
 impl From<aws_sdk_s3_transfer_manager::error::Error> for ProcessErrorKind {
     fn from(e: aws_sdk_s3_transfer_manager::error::Error) -> Self {
-        Self::Transfer(e)
+        let retryable = matches!(
```
jlizen, line 206: Looks like this is downstream of reporting the error to the circuit breaker, which is good. We don't delete the data, but we do still preserve upload backpressure.

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -1156,17 +1186,201 @@ mod worker_pipeline_tests {
...
+        worker.process_open_segments().await;
+
+        check!(
+            seg_path.exists(),
+            "segment should be kept when circuit breaker is open"
+        );
+    }
+
+    /// A NotFound error (evicted segment) is silently skipped — no deletion attempt.
+    #[tokio::test]
```
jlizen, line 1325: would be nice to also have a test on a non-retryable s3 error, but fine if you don't feel like bloating the mock

jlizen (approved):

PR #336, opened 2026-04-30
Add `TelemetryHandle::instrumented_spawn`

rcoh:
I would have preferred to name it `spawn`, but that method is already taken.

### DIFF
--- a/dial9-tokio-telemetry/README.md
+++ b/dial9-tokio-telemetry/README.md
@@ -47,12 +47,8 @@ fn my_config() -> Dial9Config {
 
 #[dial9_tokio_telemetry::main(config = my_config)]
 async fn main() {
-    // your async code here
-    // `TelemetryHandle::current()` returns the per-thread handle for
-    // spawning instrumented sub-tasks:
-    let handle = TelemetryHandle::current();
-    handle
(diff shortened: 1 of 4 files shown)


jlizen (commented):
So this is just a convenience function to avoid calling `TelemetryHandle::current()::spawn()` (which in #256 now also falls through to a non-op handle)?

Are we sure it's really worth the extra API surface? And if yes, any thoughts about just a free `spawn()` function instead? I guess `dial9_tokio_telemetry::spawn()` is a mouthful?

rcoh:
yeah if we make `TelemetryHandle::current` work this is less important? I think a free spawn function is a good idea

jlizen:
non-panicking `TelemetryHandle::current` just landed in main. It now will return an inert handle if not running with dial9.

And then, `handle::spawn()` now matches this function's behavior, but requires `&self`.
https://github.com/dial9-rs/dial9-tokio-telemetry/blob/main/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs#L481

I think a free `spawn` method is the best. We really just need a shorter crate name so we can have `dial9::spawn` ^^

jlizen:
closing in favor of #343

PR #338, opened 2026-05-01
feat!: Include CPU id in CPU profile samples

rcoh:
## Summary

Adds the CPU id that a CPU profile sample was collected on to `CpuSampleEvent`, so the viewer can show which physical CPU each sample came from.

### Breaking changes
This is a breaking change to `perf-self-profile` which will be bumped to 0.4 in the next release

## Changes

**`perf-self-profile` (breaking):**
- `Sample.cpu: u32` → `Sample.cpu: Option<u32>`. The perf backend already gets `Option<u32>` from the kernel via `PERF_SAMPLE_CPU`; previously we were `unwrap_or(0)`-ing away the missing case. The ctimer fallback now returns `None` when `SYS_getcpu` fails instead of silently reporting CPU 0 (ambiguous with real CPU 0).
- Internal `SampleData`/`DrainedSample`/`SlotWriter` thread `Option<u32>` through the lock-free signal-handler-safe ring buffer.

**`dial9-tokio-telemetry`:**
- `CpuSampleData.cpu: Option<u32>` in memory.
- `CpuSampleEvent.cpu: Option<u64>` on the wire — widened to `u64` so it encodes as `OptionalVarint`: 1 byte when absent, 2 bytes total for typical small CPU ids. Narrowed back to `Option<u32>` on decode.
- `TelemetryEvent::CpuSample` gains `cpu: Option<u32>`. Older traces recorded before this field existed decode cleanly as `cpu: None`.
- `RawCpuSample.cpu` threaded through `CpuProfiler::drain` and `SchedProfiler::drain`.

**JS parser:** `CpuSample` objects now expose `cpu: number | null`.

## Tests

- `perf-self-profile/tests/multithread.rs`: asserts every perf sample carries `Some(cpu)` (Linux-only).
- `dial9-tokio-telemetry/src/telemetry/buffer.rs`: two round-trip tests for `CpuSample` — one with `cpu: Some(7)`, one with `cpu: None` — encode through `ThreadLocalBuffer` and decode via `decode_events`.

All 422 tests pass locally, 2 stress iterations clean. `cargo fmt --check` and `cargo clippy --all-targets --all-features` clean.

## Drive-by

Narrow `cfg(test)` on `SchedStat.fd` to `cfg(all(test, target_os = "linux"))` to match its only reader, fixing a pre-existing `dead_code` warning on non-Linux builds.

## Follow-ups

- Demo trace (`dial9-viewer/ui/demo-trace.bin`) not yet regenerated — will do in a follow-up commit on this branch before merge.
- A separate PR will capture CPU id on `WorkerUnpark` so we can see scheduler migrations between park and unpark.


dial9-tokio-telemetry/src/telemetry/format.rs
```diff
@@ -196,6 +196,11 @@ pub(crate) struct CpuSampleEvent {
     pub source: CpuSampleSource,
     pub thread_name: Option<InternedString>,
     pub callchain: InternedStackFrames,
+    /// CPU the sample was taken on, if the backend could determine it.
+    ///
+    /// Widened to `u64` on the wire so the field encodes as `OptionalVarint`:
```
jlizen, line 201: Good approach

dial9-tokio-telemetry/src/telemetry/format.rs
```diff
@@ -414,6 +419,8 @@ pub(crate) fn to_owned_event(
                 .get(e.callchain)
                 .expect("stack pool entry must exist for CpuSample callchain")
                 .to_vec(),
+            // CPU id is varint-encoded as u64 on the wire; real CPU ids fit in u32.
+            cpu: e.cpu.map(|v| v as u32),
```
jlizen, line 423: nit: better to be clearer about bad wire with `u32::try_from()::ok`

jlizen (approved):
++ to the sentinel -> Option<u32>, good breaking change

Issue #189, opened 2026-04-13
Integrate metrique events into dial9


rcoh:
I _think_ the path here is:
1. Add a `value_attr` to `#[metrics(value_derive = dial9::TraceEvent)]` so that metrique _values_ are trace events
2. Add `impl TraceField for <...>` (the types that are metrique value)
3. Add a dial9_sink<E> that is a thread-local sink that records events into the dial9 TL buffer first (borrow) then send it to another global sink.
4. UI integration

https://github.com/dial9-rs/dial9-tokio-telemetry/pull/46 starts this work

jlizen:
Prior art #252

jlizen:
So I think @rcoh 's suggestion (option B below) is viable and "feels" right in that metrique is focused on nice, compile-time codegen with zero overhead. But, the caller ergonomics and maintenance load seems worse than doing runtime registration of the fields (option A below). If we went with A, we could also use that in the tracing layer to avoid a per-emission vec allocation.

---

## Common infrastructure (both options need this)

**Streaming encoder API** in `dial9-trace-format` — allows writing event fields directly to the encoder without collecting into a `Vec<FieldValue>` first. This also eliminates the per-emit allocation in the existing tracing layer.

```rust
// New: stream fields directly, no intermediate Vec
let mut event_enc = enc.begin_event_dynamic(type_id, || schema, ts);
event_enc.write_u64(worker_id);
event_enc.write_f64(latency);
```

**Thread-local sink pattern** — the interceptor/sink runs on the calling thread (worker thread for app metrics, flush thread for internal metrics), captures `worker_id`, `task_id`, and `timestamp_ns` from runtime context, encodes into the TL buffer, then forwards the entry to the global sink (e.g., `BackgroundQueue` for CloudWatch). dial9 encoding is synchronous and cheap (~100-200ns); CloudWatch IO stays on the background thread.

Both options resolve metric values to **f64/string/bool** via metrique's `Value` trait — identical wire output and type fidelity.

---

## Option A: EntryWriter adapter + interceptor

Small metrique-side addition: an `EntryInterceptor` trait and `with_interceptor()` on `EntrySink`. The dial9 side implements the interceptor, which calls `entry.write()` into a `DirectEntryWriter` that streams values straight to the trace encoder.

### User API

```rust
// Metrics struct — no changes
#[metrics(rename_all = "PascalCase")]
struct RequestMetrics {
    operation: Operation,
[...]

PR #346, opened 2026-05-02
docs(design): metrique → dial9 integration

jlizen:
Related: #189

See also: [awslabs/metrique#282](https://github.com/awslabs/metrique/pull/282) (introduces the entry descriptor / source system this design consumes).

### Summary

Design-only PR. No implementation yet; the implementation depends on [metrique#282](https://github.com/awslabs/metrique/pull/282).

The design makes dial9 a peer metrique sink: users configure dial9 once at sink setup and every metrique entry flowing through the configured sink is also recorded into the dial9 trace. A single trace file carries tokio runtime telemetry and per-request application metrics. Dial9 reads a static entry descriptor produced by metrique (one schema per entry type, regardless of optional fields or dynamic `Flex` keys), extracts caller-thread context (`worker_id`, `task_id`, start monotonic timestamp) via a typed source-extraction API, and encodes the user-selected subset of fields via new `List` and `Map` wire variants plus generic schema annotations for units.

### Documents

In-tree permanent:

- `dial9-tokio-telemetry/design/metrique-integration.md`: keeper. Architecture, user-facing API, components, trace format additions, validation, future evolution.

Review-only (deleted as part of sign-off):

- `dial9-tokio-telemetry/design/metrique-integration-review.md`: requirements, non-goals, tradeoffs, design rationale, rejected alternatives (A-Q), feasibility checks.
- `dial9-tokio-telemetry/design/metrique-integration-changelog.md`: what changed from round 1 to round 2 and why.
- `dial9-tokio-telemetry/design/metrique-integration-impl-plan.md`: sequencing, module-level work, public API signatures, risks, testing plan.

### Dependencies

The implementation cannot land before [metrique#282](https://github.com/awslabs/metrique/pull/282) does. Metrique must first ship the descriptor system, source system (`SourceTag` trait with `type Snapshot` and `register_descriptor` hook), field-tag attributes, `no_write`, and the `descriptor()` method on the erased entry vtable. See `metrique/docs/entry-descriptors.md` (in [metrique#282](https://github.com/awslabs/metrique/pull/282)) for the metrique-side design.


dial9-tokio-telemetry/design/metrique-integration-review.md
```diff
@@ -0,0 +1,156 @@
...
+4. **Runtime warning for manual-composition misuse.** Relies on `tracing` being configured. Users running without `tracing` subscribed won't see the warning.
+
+## Key design choices
+
+### Runtime schema discovery rather than compile-time
+
+Two blockers on the compile-time path, either sufficient on its own:
+
+1. **Heterogeneous sinks**. The Global-sink path erases entries to `BoxEntry` before our stream sees them. A `Dial9Sink<E, S>` parameterized by a specific concrete `E` only works for homogeneous pipelines; users mixing metrique struct types through one pipeline would have to split their sinks.
+2. **Open-ended `TraceField` maintenance**. Every `Value` impl in metrique (primitives, unit wrappers, aggregation types) would need a parallel `TraceField` impl. Every user custom type that today works with metrique by implementing `Value` alone would need a new `TraceField` impl to work with dial9. This is a cost users would pay on every custom type they add.
```
rcoh: I figured we would have a blanket impl of TraceField for impl Value fwiw

dial9-tokio-telemetry/design/metrique-integration-review.md
```diff
@@ -0,0 +1,156 @@
...
+- Dial9 publishing `TraceField` implementations for metrique types (`Duration`, `SystemTime`, `Timer`, `TimestampValue`, `FlexEntry<T>`) plus a `TraceField` derive for value-string enums. Users extend for their own types.
+- A compile-time pairing check inside metrique-macro that errors if a struct has `derive(TraceEvent)` without `observe(Dial9)`.
+
+Two independent blockers, either of which would have been decisive:
+
+**Blocker 1: the Global sink path erases `TraceEvent` through `BoxEntry`.** `BoxEntrySink::append_any` takes `impl Entry + Send + 'static` and calls `entry.boxed()` internally, so by the time `TokioContextSink` or `Dial9Stream` receives the entry, the user's concrete type has been erased to `BoxEntry`. `TraceEvent`-ness cannot be recovered from there. Preserving it would require either a parallel object-safe `DynTraceEvent` trait plus a dial9-owned box type (`TraceEvent` is not object-safe today: GAT `Ref<'a>`, generic `encode_fields<W: Write>`), or a metrique-side change so `ServiceMetrics` can attach a sink typed concretely over `TraceEvent + Entry`.
```
rcoh: This is generally true but Encodeable is dyn Safe

dial9-tokio-telemetry/design/metrique-integration.md
```diff
@@ -0,0 +1,236 @@
...
+    #[metrics(flatten)]
+    counters: Flex<u64>,
+}
+```
+
+No attributes, no derives, no per-struct wiring. The design deliberately avoids coupling at the `#[metrics]` layer.
+
+## Why runtime schema discovery
+
+`Dial9Stream` learns each entry's schema at encode time by walking `Entry::write` through an adapter, rather than relying on dial9's compile-time `#[derive(TraceEvent)]` machinery. Two independent blockers, either sufficient on its own:
```
rcoh: Does this handle optional fields?

dial9-tokio-telemetry/design/metrique-integration.md
```diff
@@ -0,0 +1,236 @@
...
+
+1. If the handle is inert (no dial9 runtime attached, or the runtime is disabled): return `Ok(())` immediately. No field walk, no cache lookup, no work.
+2. Walks `entry.write(&mut adapter)` where `adapter` is an internal `EntryWriter` implementation.
+3. The adapter's `config()` method picks `TokioContext` out of the config stream.
+4. The adapter's `value(name, value)` method resolves each metrique `Value` through an internal `ValueWriter` that captures `(FieldType, encoded_bytes, Unit)`.
+5. Field names receive a unit suffix when the unit is not `Unit::None`: `latency_Milliseconds`, `payload_size_Bytes`, and so on. The suffix uses `Unit::name()` verbatim. Some unit names contain `/` (e.g. `Bytes/Second`) or arbitrary user text (`Unit::Custom`), producing suffixes like `throughput_Bytes/Second` that downstream consumers must handle or expect normalization on (see Prospective improvements).
+6. Computes an order-dependent shape fingerprint over `(name, field_type)` pairs. Looks up or registers a `dial9_trace_format::encoder::Schema` in a bounded cache.
```
rcoh: The main challenge here is that with optional fields a realistic struct can have an unbounded number of schemas when discovered in this way.

This is why I was suggesting at one point we expose a schema in metrique land.

The issue is not for a single level of fields but multiple optional nested structs.

What we really want is "tell me all the fields you can possibly produce" as well as: 
- tell me which are optional
- tell me which are dynamic

Another idea I was considering was adding a new DynStruct primitive where the keys are InternedString and the values are TaggedValue (allowing the use of any existing dial9 values). It might be almost as good? Would need prototype and benchmark.

dial9-tokio-telemetry/design/metrique-integration.md
```diff
@@ -0,0 +1,236 @@
...
+- **Flex keys**: each distinct `Flex::new(key)` string is a distinct cache entry. With `M` Flex fields in a struct and `N` distinct keys per field, cardinality is `N^M`.
+
+Thrash cost when the working set exceeds the cap is real, not "graceful". Each cache miss registers a fresh schema with the encoder, which writes a schema frame (roughly 100-150 bytes for a 5-field struct, dominated by the field name strings) and re-interns any previously-evicted field names. On a workload that keeps missing, every emission pays this overhead on top of the event payload itself; the trace wire volume can double or more.
+
+In practice, Flex is used sparingly and with low per-field cardinality, so typical usage stays well under the cap. The prospective Flex field promotion (below) replaces unbounded-key Flex with a single typed-map schema and eliminates the thrash path entirely.
```
rcoh: What if we add a Flag to flex so callers know that it's flexing and can handle it accordingly? (E.g. collect all flex fields into map at end)

rcoh (approved):
Great design. Left a couple of comments but basically all of them I found answers too as I kept reading.

rcoh:
Thinking about it more, I think we may want an opt in flag to make fields part of the dial9 record. You don't want every field in many cases I think.

For example, you may just want a request id for correlation and maybe a KPI or two. But in the general case you might not need the whole wide event, especially if there is a performance implication.

We also need to ensure there is a reasonable way to record context when the event starts.

Adding a flag (which you can see on a full struct or your root struct) would be a fairly non intrusive way to handle it.

dial9-tokio-telemetry/design/metrique-integration-review.md
jlizen: Yeah good clarification on the object safety, but the issue is less about object safety, and more about being able to properly pass through `BoxEntrySink` (which don't have these trait bounds). Ergonomics get pretty poor without that.

dial9-tokio-telemetry/design/metrique-integration.md
jlizen: Currently yes, but like you pointed out, it does cause a potential schema count explosion (2^ number of optional fields)

dial9-tokio-telemetry/design/metrique-integration.md
jlizen: Yeah, I'm starting to think a metrique introspection API would be nice (give me all possible fields, which fields are optional, which are dynamic). And ideally some amount of awareness of enums so we can avoid "super schemas" that are larger on the wire when we might prefer to fan out into smaller structs that are largely independent.

I guess that's true even without enums, if options frequently show up in combination. Perhaps there are some optimizations available for runtime tracking of common schema combinations to slowly narrow in an efficient way.

That in fact might be a reason to prefer runtime discovery... that and because it just is easier for N+1 use case to integrate with, and it has cleaner semantics around sink-side declaration.

But, the filtering problem is a real one, still mulling on it.

dial9-tokio-telemetry/design/metrique-integration-review.md
jlizen: Probably the least bad option to address that is some sort of "TypeId lookup" API on BoxEntry in `metrique-core`, with a dial9 (or other consumer)-side vtable bridge that maps `TypeID -> &dyn X`... which doesn't feel great.

dial9-tokio-telemetry/design/metrique-integration.md
jlizen: Yeah, this is probably a better approach, and then we could have a non-stringly-only-typed StringMap to hold them. My initial design was doing its best to stay on the `dial9` side of the fence, but let me see what is possible if we expose a nicer API on metrique.

(For what's it's worth, I spent a LONG time playing around with some sort of `#[metrics(on_close = Dial9Observer)]` API that had access to the underlying Entry, and couldn't quite get a reasonable API that played nicely both with configuring the observer, and not creating nasty field mapping problems for custom types. (I'm aware that one could have a blanket impl for Value, but then we lose the compile-time schema benefit).

Will poke at both some more.

jlizen:
Thanks Russell... I do agree that this ought to be a requirement:
> good way to opt in on a per-struct and per-field basis

Of course for compile time usage, that's pretty easy. If we stick with runtime, I do like the idea of tags that can be passed through somewhat generically:

```
  #[metrics(trace_streams = [dial9::Dial9Tag])]
  struct RequestMetrics {
      #[metrics(trace_streams = [dial9::Dial9Tag])]
      request_id: String,
      // ...
  }
```

Though we do need to figure out a decent way to have validation for setting the tag but no sink level config (if needed).

jlizen:
For:
> record context when the event starts

I did have some of this as part of earlier iterations. It is very appealing to have a span-like thing. I had punted and focused on making sure we have a good monotonic timestamp of append time (so we can stick a good point on the chart, at least). But we should at least have an evolution path to span.

The easy ways were unpleasant....I don't want invasive wrappers you need to name explicitly, or force different constructors. (Let me add these to the negative requirements.)

What feels better is, customizing `AppendOnDrop` to have a hook somehow (with macro support). Though I hadn't found a way that felt clean yet... I can take another stab. Honestly a good job of that might address a lot of runtime concerns too, if we are doing compile-time codegen...

jlizen:
I think my takeaway is, let me give a thought to what are some clean API changes to make on the `metrique` side, and what they open up on the runtime side on `dial9`.

I was previously pretty focused on "on close"/observer like hook systems on the metrique side. Those ones involved codegen that was `Entry` type aware.  But passing simple, static information through to handle at runtime, is a direction I didn't do much with. Also a new hook interface on `AppendOnDrop`, maybe compile time.

Are you generally ok with runtime interpretation and generally heavier encoding on the runtime thread, if we can meet the requirements? Ultimately I think we can get the wire size to be the same (or better) with cleverness with our `EntryWriter`.

I've found that if there is a good way to model things as `EntryIoStreams`, it opens a lot better composition options with the rest of the `metrique` API.

rcoh:
It's not clear to me why a dial9 span field with a Default impl is major API drawback. Maybe I am missing something

```
struct MyThing {
  #[metrics(flatten)]
  d9: D9Meta,
  ...
}
```

jlizen:
That's true, calling `..Default::default()` isn't the worst thing... These types are all defined in the consumer anyway.

It still would be nicer to have an API that doesn't change consumer usage besides decorating their struct or stream/sink.

Though, also having a good way to fold dial9 context into other format emissions is nice too. So maybe having a library type in your metric definition is the right integration point.

rcoh:
Oh one more thing we need to consider units.

My recommendation is to go simple and have a normative unit for all types, e.g microseconds, Bytes etc. and convert to that unit on storage. You do still need the ability to notate the unit for the value (In the schema?)

jlizen:
For v1 I was just going to do dumb inclusion in field name similar to how our dial9 defined trace schemas work, except uglier. Normative is technically ok for some but it doesn't really work for bytes/sec or custom types. I don't see a whole lot of value of converting compared to just passing through the units directly?

The slightly less ugly thing would be letting the units define a short version (eg `ns`) to have less ugly field name concenation and actually match our hardcoded schemas' styles.

Ultimately probably we want some sort of schema extension to handle these... But felt like scope creep...

jlizen:
Second attempt is ready for review! There is now a "changelog" markdown file, I'd suggest starting there @rcoh .

The new design involves metrique changes, so see https://github.com/awslabs/metrique/pull/282 as well.

dial9-tokio-telemetry/design/metrique-integration-changelog.md
```diff
@@ -0,0 +1,108 @@
...
+
+The round-1 design was functional but could not structurally handle optional-field schema explosion or unbounded Flex keys. It also captured caller-thread context through a sink-wrapper that made dial9 a privileged sink in the composition.
+
+## What changed, by area
+
+### Schema discovery
+
+Round 1: `Dial9Stream` walked `Entry::write` on every event, hashed `(name, field_type)` pairs into a shape fingerprint, and looked up or registered a dial9 schema in a bounded LRU cache.
+
+Round 2: metrique emits a `&'static EntryDescriptor` per macro-derived entry. `Dial9Stream` keys its schema cache on the descriptor pointer. No fingerprinting, no LRU eviction, no thrash path. Hand-written entries return `None` for the descriptor and are skipped with a rate-limited report.
```
rcoh: this is on the `Entry` value itself right? that would allow e.g. `enum`s to produce a different schema for each variant. you may want to consider relaxing the `'static` in favor of making it cheap-to-clone

dial9-tokio-telemetry/design/metrique-integration-changelog.md
```diff
@@ -0,0 +1,108 @@
...
+
+Round 1: dial9 did not expose string interning to the metrique path.
+
+Round 2: `InternString` is a separate field tag. It is orthogonal to `InTrace`; it says "if this emitted field carries string data, route it through dial9's string pool." Applies uniformly, including Flex keys and values when explicitly requested.
+
+### Units
+
+Round 1: units were stamped into field names (`latency_Microseconds`). Ugly and not round-trippable; `Unit::Custom` produced awkward suffixes.
+
+Round 2: units stay first-class in the descriptor (`Option<Unit>` per field). Dial9 emits them as schema-level annotations (`("metrique.unit", "microseconds")`). Fields with no unit cost zero bytes. This generalises to other per-field metadata (display hints, privacy, semantic conventions) without further format churn.
```
rcoh: cool! this also will allow things like `dial9.kpi` to annotate fields that should be graphed, `dila9.span.start` etc.

Eventually even the Tokio fields could use this to make them not "special"

dial9-tokio-telemetry/design/metrique-integration-changelog.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Heterogeneous queues / `BoxEntry` erasure
+
+Round 1: descriptor lookup not a problem in round 1 because there was no descriptor; instead the problem was that `TraceEvent`-ness could not survive `BoxEntry` erasure.
+
+Round 2: descriptor lookup goes through the erased entry vtable via a new method that returns `Option<&'static EntryDescriptor>`. `BoxEntry` size is unchanged. Descriptor-unaware sinks never call the method.
```
rcoh: note that EntryDescriptor should probably include some concept of a canonical name

dial9-tokio-telemetry/design/metrique-integration-changelog.md
```diff
@@ -0,0 +1,108 @@
...
+Round-1 entries are unchanged but do not produce dial9 traces. Users opt in by adding the tags and the `Dial9Context` field.
+
+### Validation
+
+Round 1: no dial9-specific validation story. Misuse was caught if at all by the fact that the runtime-fingerprinted schema looked wrong.
+
+Round 2: three-tier validation.
+
+- Compile-time, in the metrique macro: intrinsic structural checks (duplicate sources, conflicting tags). Independent of any sink.
+- Startup-time, at `Dial9Stream::new`: dial9 registers every descriptor declaring `source(Dial9)` via metrique's `SourceTag::register_descriptor` hook (backed by `linkme` internally). Empty registry at sink construction is `debug_assert!` in debug, rate-limited `tracing::warn!` in release. Per-descriptor first-use checks for structural errors (`InTrace` without `Dial9` source, `InternString` on non-string, `Opaque` in `InTrace`) use `debug_assert!` in debug and rate-limited `tracing::error!` in release. Users who hit legitimate empty-registry false negatives disable per sink via `.startup_discovery(false)`; unsupported targets cfg out the hook entirely.
```
rcoh: I think we can do this lazily. apparently `linkme` has some issues but this could be a follow up anyway

rcoh (commented):
I read the changelog only so far (thanks for that). I like the direction

jv1i:
The overall approach seems sound to me! 
You both have more context so I'm mostly commenting from the side :)

One thing that is not obvious to me is what are we planning/expecting to do with the new information on the trace, at the viewer level (or analysis). 
So I can't easily tell if there's something we should track that we're missing, or if a different structure would be more helpful, etc.
For instance: would it be useful to have an end timestamp by default on `Dial9Context`, like `TimestampOnClose` but monotonic? Thinking of some viewer features: 

- show CPU usage / scheduling wait during a slow request: "a RequestMetrics event took 800ms, what was the runtime doing during that window?" might need (start, end) since the associated task can outlive the metric event

- visualizing two metric events nested on the same task: e.g. an inner DbQueryMetrics opened with its own append_on_drop inside an outer RequestMetrics block on the same handler and both share task_id. Might need an end timestamp on each, to tell "DB query ran inside the request" from "DB query started right after the request finished."

- perhaps showing a metric event lane? (rendering the metric events as spans)

(some of these might be solved somehow else with current dial9 machinery I haven't seen yet)

dial9-tokio-telemetry/design/metrique-integration-changelog.md
jlizen: I'm going to make this opaque to allow future addition of non-`'static` but avoid implementing it for now. Will make it slightly more complex internally but the API stays fine.

dial9-tokio-telemetry/design/metrique-integration-changelog.md
jlizen: Agreed, will add

dial9-tokio-telemetry/design/metrique-integration-changelog.md
jlizen: I'm going to rip this out of initial scope regardless

jlizen:
Thanks Julián!

> For instance: would it be useful to have an end timestamp by default on Dial9Context, like TimestampOnClose but monotonic?

Ah yeah, I was intending to have this but dropped it during an iteration, adding back.

I intentionally didn't go very deep into the viewer, but i can add some top-level discussion of what shape the viewer wants without focusing on the specifics of visualization.

jlizen:
This is ready for fresh eyes @rcoh / @yulnr , I'll get started on the schema field extensions in the meantime.

There is a glossary now, and also a changelog since last review.

yulnr (approved):

PR #349, opened 2026-05-04
feat: support tracing futures in JoinSet and external spawn APIs

### LINKED ISSUE #301: make Traced future public
Right now, there is no way to trace a future without spawning it. This makes it impossible to trace futures that are spawned into a joinset

noxware:
## Current solution

Exposes a `handle.spawn_with(fut, |f: TracedFuture| { ... })` to allow spawning instrumented/traced futures through custom spawn APIs (like `JoinSet::spawn`).

## Example

```rust
let mut set: JoinSet<()> = JoinSet::new();
handle.spawn_with(fut, |f| set.spawn(f));
```

## Other
Closes #301

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
+++ b/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
@@ -33,9 +33,9 @@ crate::primitives::thread_local! {
     /// cleared in `on_thread_stop`. Enables [`TelemetryHandle::current`].
     static CURRENT_HANDLE: RefCell<Option<TelemetryHandle>> = const { RefCell::new(None) };
 
-    /// Set by `TelemetryHandle::spawn()` before calling `tokio::spawn()`,
-    /// so the `on_task_spawn` hook can distinguish instrumented from raw spawns.
-    static INSTRUMENTED_SPAWN: Cell<bool> = const { Cell::new(false) };
+    /// Nest count for [`InstrumentedSpawnGuard`]. `on_task_spawn` treats
+    /// any value `> 0` as an instrumented spawn.
+    static INSTRUMENTED_SPAWN: Cell<u32> = const { Cell::new(0) };
 }
 
 // ---------------------------------------------------------------------------
@@ -204,7 +204,7 @@ fn register_hooks(
             s5.if_enabled(|buf| {
                 let task_id = TaskId::from(meta.id());
                 let location = meta.spawned_at();
-                let instrumented = INSTRUMENTED_SPAWN.with(|f| f.get());
+                let instrumented = INSTRUMENTED_SPAWN.with(|f| f.get()) > 0;
                 let timestamp_ns = crate::telemetry::events::clock_monotonic_ns();
                 buf.record_encodable_event(&runtime_context::TaskSpawn {
                     timestamp_ns,
@@ -506,15 +506,66 @@ impl TelemetryHandle {
     {
         match self.traced_handle() {
             Some(traced_handle) => {
-                let _guard = InstrumentedSpawnGuard::set();
+                let _guard = InstrumentedSpawnGuard::enter();
                 tokio::spawn(async move {
-                    let task_id = tokio::task::try_id().map(TaskId::from).unwrap_or_default();
+                    let task_id = TaskId::from(tokio::task::id());
                     let inner = wrap_task_dumped(future, traced_handle.shared.clone(), task_id);
-                    crate::traced::Traced::new(inner, traced_handle, task_id).await
+                    crate::telemetry::Traced::new(inner, Some(traced_handle)).await
                 })
             }
             None => tokio::spawn(future),
         }
+    }
+
+    /// Spawn a wake-tracked future through a user-supplied spawn function.
+    ///
+    /// `spawn_fn` is invoked synchronously with a [`Traced<F>`] that
+    /// the caller is expected to spawn immediately. The closure's return
+    /// value is forwarded back to the caller, so you can keep the
+    /// [`tokio::task::JoinHandle`], [`tokio::task::AbortHandle`], or
+    /// whatever the spawn function returns.
+    ///
+    /// # Examples
+    ///
+    /// Spawn into a [`tokio::task::JoinSet`]:
+    ///
+    /// ```rust,no_run
+    /// # use dial9_tokio_telemetry::telemetry::TelemetryHandle;
+    /// # use tokio::task::JoinSet;
+    /// # async fn work() {}
+    /// # async fn demo() {
+    /// let handle = TelemetryHandle::current();
+    /// let mut set: JoinSet<()> = JoinSet::new();
+    /// handle.spawn_with(work(), |f| set.spawn(f));
+    /// # }
+    /// ```
+    ///
+    /// Spawn into a [`tokio::task::JoinSet`] on a specific runtime:
+    ///
+    /// ```rust,no_run
+    /// # use dial9_tokio_telemetry::telemetry::TelemetryHandle;
+    /// # use tokio::runtime::Runtime;
+    /// # use tokio::task::JoinSet;
+    /// # async fn work() {}
+    /// # fn demo(runtime: &Runtime) {
+    /// let handle = TelemetryHandle::current();
+    /// let mut set: JoinSet<()> = JoinSet::new();
+    /// handle.spawn_with(work(), |f| set.spawn_on(f, runtime.handle()));
+    /// # }
+    /// ```
+    ///
+    /// [`Traced<F>`]: crate::telemetry::Traced
+    pub fn spawn_with<F, S>(
+        &self,
+        future: F,
+        spawn_fn: impl FnOnce(crate::telemetry::Traced<F>) -> S,
+    ) -> S
+    where
+        F: std::future::Future + Send + 'static,
+        F::Output: Send + 'static,
+    {
+        let _guard = InstrumentedSpawnGuard::enter();
+        spawn_fn(crate::telemetry::Traced::new(future, self.traced_handle()))
     }
 }
 
@@ -565,20 +616,20 @@ where
     TelemetryHandle::current().spawn(future)
 }
 
-/// RAII guard that sets `INSTRUMENTED_SPAWN` to `true` on creation and
-/// resets it to `false` on drop, even if `tokio::spawn` panics.
+/// RAII guard that increments `INSTRUMENTED_SPAWN` on creation and
+/// decrements it on drop, even if the protected closure panics.
 struct InstrumentedSpawnGuard;
 
 impl InstrumentedSpawnGuard {
-    fn set() -> Self {
-        INSTRUMENTED_SPAWN.with(|c| c.set(true));
+    fn enter() -> Self {
+        INSTRUMENTED_SPAWN.with(|c| c.set(c.get().saturating_add(1)));
         Self
     }
 }
 
 impl Drop for InstrumentedSpawnGuard {
     fn drop(&mut self) {
-        INSTRUMENTED_SPAWN.with(|c| c.set(false));
+        INSTRUMENTED_SPAWN.with(|c| c.set(c.get().saturating_sub(1)));
     }
 }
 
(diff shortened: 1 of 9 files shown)


dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -513,6 +513,61 @@ impl TelemetryHandle {
...
+    ) -> impl std::future::Future<Output = F::Output> + Send + 'static
+    where
+        F: std::future::Future + Send + 'static,
+        F::Output: Send + 'static,
+    {
+        let traced_handle = self.traced_handle();
+        async move {
+            match traced_handle {
+                Some(handle) => {
+                    let task_id = tokio::task::try_id().map(TaskId::from).unwrap_or_default();
```
rcoh: default is almost certainly not a useful fallback to store here

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -513,6 +513,61 @@ impl TelemetryHandle {
...
+
+    /// Run `f` with the instrumented-spawn flag set so any `tokio::spawn`
+    /// inside it produces a `TaskSpawn` event with `instrumented = true`.
+    /// Nestable.
+    pub fn with_instrumented_spawn<R>(&self, f: impl FnOnce() -> R) -> R {
+        let _guard = InstrumentedSpawnGuard::enter();
+        f()
+    }
+
+    /// Spawn `future` into `set` with wake tracking and the `instrumented`
```
rcoh: this should be `set.span(handle.trace(fut)` right? In any case, this is probably better at an extension trait or an example (if at all)

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -589,6 +644,63 @@ impl RuntimeTelemetryHandle {
             None => self.runtime.spawn(future),
         }
     }
+
+    /// See [`TelemetryHandle::trace`].
+    pub fn trace<F>(
```
rcoh: this code looks identical to the other code?

noxware: Yep, replicated between `RuntimeTelemetryHandle` and `TelemetryHandle`, the same way as `spawn`. 

An idea to improve this in the future: We could rename `TelemetryHandle` to something like `ThreadTelemetryHandle` so then we can have a trait called `TelemetryHandle` to enforce "how a telemetry handle should behave". WDYT?

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
noxware: I agree.

Please note that `spawn` was doing the same and I just replicated that. But I can give it some thought to improve it across both use cases.

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
rcoh: hmmm...yeah ideally for 0.4 we will find a clear path to unify and deduplicate these APIs. In the meantime, I would extract a function like `fn trace(f, traced_handle)` so that both callsites can use the same actual impl without having to duplicate the code

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -513,6 +513,61 @@ impl TelemetryHandle {
...
+    pub fn with_instrumented_spawn<R>(&self, f: impl FnOnce() -> R) -> R {
+        let _guard = InstrumentedSpawnGuard::enter();
+        f()
+    }
+
+    /// Spawn `future` into `set` with wake tracking and the `instrumented`
+    /// `TaskSpawn` flag. Panics under the same conditions as
+    /// [`tokio::task::JoinSet::spawn`].
+    #[track_caller]
+    pub fn spawn_in_joinset<F, T>(
```
rcoh: lets delete this and replace it with an `# Examples` section on the `trace` function

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
noxware: The ergonomic would be equivalent to

```rust
handle.with_instrumented_spawn(|| set.spawn(handle.trace(future)));
```

I can move it to an extension trait or remove it completely, just say the words :)

rcoh: yeah lets remove it and handle this in docs / examples

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
noxware: Done. I handled `None` inside `trace()` by allowing the future to run transparently with a warning in that case. I suppose this is consistent with the behavior in other parts of the crate and avoids blowing up the user's program.

I went a bit further and changed the existing `spawn()` functions to use `tokio::task::id()` instead of `tokio::task::try_id()`, as in those places, `None` will never be returned, and `try_id` with `.unwrap_or_default()` confused me.

```rust
// always
tokio::spawn(async {
  // appropriate use, immediately, will not panic
  tokio::task::id()
  // ...
}
```

@rcoh WDYT?

noxware:
@rcoh thank you for the review, it should be ready for another pass 🙌

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -504,15 +504,77 @@ impl TelemetryHandle {
...
+    }
+
+    /// Run `f` with the instrumented-spawn flag set so any `tokio::spawn`
+    /// inside it produces a `TaskSpawn` event with `instrumented = true`.
+    /// Nestable.
+    ///
+    /// Combine with [`trace`](Self::trace) to instrument futures spawned
+    /// through APIs other than [`spawn`](Self::spawn) (e.g.
+    /// [`tokio::task::JoinSet::spawn`]).
+    pub fn with_instrumented_spawn<R>(&self, f: impl FnOnce() -> R) -> R {
```
rcoh: Add an example here. This does nothing except set the thread local so I'm not really sure this API is required.

If we are going to trace the future anyway can we set the TL there?

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -504,15 +504,77 @@ impl TelemetryHandle {
...
+    F: std::future::Future + Send + 'static,
+    F::Output: Send + 'static,
+{
+    match traced_handle {
+        Some(handle) => match tokio::task::try_id().map(TaskId::from) {
+            Some(task_id) => crate::traced::Traced::new(future, handle, task_id).await,
+            None => {
+                rate_limited!(Duration::from_secs(60), {
+                    tracing::warn!(
+                        "trace() future awaited outside a Tokio task context; running future without wake tracking"
```
rcoh: The relevant action here is actually polling

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -580,15 +642,57 @@ impl RuntimeTelemetryHandle {
         match &self.traced {
             Some(traced) => {
                 let traced = traced.clone();
-                let _guard = InstrumentedSpawnGuard::set();
+                let _guard = InstrumentedSpawnGuard::enter();
                 self.runtime.spawn(async move {
-                    let task_id = tokio::task::try_id().map(TaskId::from).unwrap_or_default();
+                    let task_id = TaskId::from(tokio::task::id());
```
rcoh: This can panic but is generally safe since we are calling tokio::spawn

Still, it would be good to justify this as unreachable in a comment

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -580,15 +642,57 @@ impl RuntimeTelemetryHandle {
...
                     crate::traced::Traced::new(future, traced, task_id).await
                 })
             }
             None => self.runtime.spawn(future),
         }
     }
+
+    /// See [`TelemetryHandle::trace`].
```
rcoh: Having a duplicate sentence here is helpful to avoid having to jump

rcoh: Having a duplicate sentence here is helpful to avoid having to jump

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -580,15 +642,57 @@ impl RuntimeTelemetryHandle {
...
+    where
+        F: std::future::Future + Send + 'static,
+        F::Output: Send + 'static,
+    {
+        trace_future(future, self.traced.clone())
+    }
+
+    /// See [`TelemetryHandle::with_instrumented_spawn`]. Sets the flag on
+    /// the calling thread, not on the runtime's workers.
+    pub fn with_instrumented_spawn<R>(&self, f: impl FnOnce() -> R) -> R {
```
rcoh: Same here, it is not clear to me this needs to be a public function

rcoh (approved):
Changes look good. Can you figure out how to remove the need to have a separate public with_instrumented_spawn API?

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
noxware: I gave it a second thought but I came to the same conclusion. 

In order to fully replicate what `TelemetryHandle::spawn` does, with arbitrary `spawn` functions we need to follow these steps:
- Construct the wrapper `Traced` future so wake events are emitted.
- Set `INSTRUMENTED_SPAWN` **just before** calling any `spawn` function.
	- This is to emit `TaskSpawn { instrumented: true }` at `on_task_spawn`.
- Call the arbitrary `spawn` function (like the `JoinSet` one).
- **Immediately**  unset `INSTRUMENTED_SPAWN` right after the `spawn` call to avoid affecting further unrelated `spawn` calls.

We can't do this in `trace` because:
- If we do it inside the future block, it will not be run until `spawn` has been already called to drive it.
- If we do it synchronously, immediately when the function is called, but before returning the future, then, other unrelated spawns between the future construction and the `spawn` consuming it will be incorrectly marked as `instrumented`. In other words, a simple `set.spawn(handle.trace(fut))` may look like it "works", but something like this will give incorrect behavior:

```rust
let traced = handle.trace(fut); // sets INSTRUMENTED_SPAWN
tokio::spawn(other); // unrelated, incorrectly affected by INSTRUMENTED_SPAWN
set.spawn(traced); // the traced future unsets INSTRUMENTED_SPAWN when polled
```

> Note: According to [tokio docs](https://docs.rs/tokio/latest/tokio/task/fn.spawn.html), "It is guaranteed that spawn will not synchronously poll the task being spawned". So that "unset on poll" behavior is quite fragile, probably incorrect.

I see 3 options:
- Keeping the closure based API.
- Exposing `InstrumentedSpawnGuard` instead as a public RAII API.
- Removing the closure and doing nothing to handle this (in this PR at least). Wake events will still be emitted, but `TaskSpawn` will be emitted with `instrumented: false` (I suppose), so it may affect the viewer (probably).

Tell me what you prefer.

jv1i: As this stands now in the PR the spawns that go through `trace` without `with_instrumented_spawn` will be marked always as unistrumented and show as such in the viewer. So I don't think is ideal to give two options to users one where `instrumented` will be misleading and one where it wont. Also publicly  naming it `with_instrumented_spawn` is a bit misleading, if someone is using trace the spawns will be instrumented either way, they just won't be marked as such.
So I wouldn't do 1 or 3.

jv1i: Would this still be useful if `trace` didn't return a `Future` at all? Right now `trace() -> impl Future` is spawnable as-is, which is part of what creates the labeling gap. If `trace` instead takes the spawn fn and returns whatever it returns, the caller never gets a future to mis-spawn:

```rust
pub fn trace<F, S>(
    &self, fut: F,
    spawn_fn: impl FnOnce(TracedFuture<F>) -> S,
) -> S {
    let _g = InstrumentedSpawnGuard::enter();
    spawn_fn(TracedFuture { inner: Traced::new(fut, ...) })
}
```

Usage:
```rust
handle.trace(fut, |f| set.spawn(f));
handle.trace(fut, |f| set.spawn_on(f, rt.handle()));
handle.trace(fut, tokio::spawn);
```

`TracedFuture<F>` is opaque, impls `Future` so the closure can spawn it, and`Traced` stays private. 

Or, expose `Traced` directly as in the original intent? If that still problematic we could re-consider some earlier detection options like deriving `instrumented` from a separate `TaskInstrumented` event emitted on first poll (to drop the TL flag, though these other options have downsides). But I don't think that should be necessary for this.

noxware: @yulnr  I just thought about something similar (but your idea looks better). The only relevant issue I see with `handle.trace(fut, spawn)` is that it forces you to `spawn` instead of `await` inside an already active future. 

Is that an issue? Or the use case for awaiting a traced future inside another future doesn't exist? 

EDIT: I think it's fine, just asking to confirm.

jv1i: Mh what would be the problem for that? or do you mean someone having the need to explicitly await inline?
I think awaiting inline could have another issue actually with duplicated wake events if the outer task is instrumented.

One issue with what I suggested might be tracking spawn loc if someone passes an fn-item instead of a closure (e.g. `tokio::spawn`), I think `#[track_caller]` wouldn't help there. But that use case should be covered by `handle.spawn()`, so perhaps we could document that (not sure if we can type against it on `trace`)?.

noxware:
Progress at this point:
- Implemented the solution using public `Traced` as the concrete type.
- Rebased dropping old implementation commits to keep the Git history clean.

However, `main` has a `TaskDumped` in `spawn` which should be considered in the solution.

noxware:
Clean PR at #392

PR #355, opened 2026-05-05
Expose runtime pipeline

### LINKED ISSUE #144: Feature: GCS and/or out-of-core trace file upload?
The S3 support here looks really nice to take advantage of, especially now that I have working CPU profiles integrated in my system. Would it be possible to extend this to something like GCS as well? In a project at `$WORK` we use `object_store` for these purposes, which works fine, but I'm not particularly picky about that, and I'd like to investigate using dial9 there.

More broadly, it would be kind of nice if you could implement this feature out-of-band if this isn't already possible? Maybe it is possible just with a bit more work today? You could imagine various other transport mechanisms (FTP, SSH copy, etc) to push traces around. Or alternative imposed constraints; as an example, [in my repository](https://github.com/thoughtpolice/a) it is a bit difficult to use certain crates like `aws-lc-rs` or `ring` (for Reasons), so I typically use a very specific TLS stack (BoringSSL) with appropriate adapters. Implementing upload on my side would let me sidestep possible complications, e.x. `object_store` depends on `ring`.

Fluzko:
closes #144 

## Summary
Open up the worker pipeline so users can plug in their own segment processors. Previously, the worker had a hard-coded chain `([Symbolize?, Gzip, S3 or WriteBack])` and S3 was the only supported sink. This PR turns the chain into a first-class public API and reorganizes the builder around it.
                                                                                                        
###  Core decisions  
  1. Make `SegmentProcessor` a public trait, with a `PipelineBuilder` for composition. Users can now implement processors and chain them with `.pipe(...)` **(Are we ok with this naming?)** alongside built-ins `(.gzip(), .write_back(), .s3(), .symbolize())`. `SegmentData`, `ProcessError`, and `SealedSegment` are also lifted to public, with accessors and constructors, so any third-party processor has what they need without reaching into internals.                                                         
  2. PipelineBuilder doesn't add any steps, leaving the user full control over the pipeline. `PipelineBuilder::s3()` deliberately does not auto-gzip; the user chains `.gzip().s3(cfg)` if they want compressed uploads. Hiding implicit steps in the primitive makes downstream surprises hard to debug. Presets (with_s3_uploader) are the place to compose, the builder is the place to be literal.
  3. Pipeline strategy is enforced at the type level. `TracedRuntimeBuilder<P, M>` gains a second marker: PipelineUnset > either `PipelineS3` (via `with_s3_uploader`) or `PipelineCustom` (via `with_custom_pipeline`). The two are mutually exclusive at compile time, so misconfigurations like "S3 preset and a custom chain" become unrepresentable. Once in PipelineS3, only S3-shaped tweaks (`with_s3_client`, replacing the uploader) remain in scope; once in `PipelineCustom`, the pipeline is frozen.
  4. `S3PipelineUploader::new` is now synchronous;  Previously new was async, which forced a tokio runtime at builder time. The uploader now holds a `Pending { config, client: Option<_> } / Ready { ... }` state and resolves the client + bucket region on the first `process()` call, inside the worker's own runtime. The whole runtime-builder API can stay sync end-to-end.                                                                                                                        
  5. `with_s3_uploader` and `with_s3_client` are call-order independent. The previous shape silently dropped the client when `with_s3_client` was followed by `with_s3_uploader` (the replacement uploader was rebuilt without it). Fixed by `S3PipelineUploader::take_client`, which carries a pending client across uploader replacement. There's a regression test for it.
  6. The worker is now pipeline-agnostic. `BackgroundTaskConfig` no longer carries s3, client, or symbolize fields. It takes a `Vec<Box<dyn SegmentProcessor>>` and runs them. Strategy-to-processor desugaring lives in one place (assemble_processors), with a documented behaviour matrix:

### DIFF
--- a/dial9-tokio-telemetry/src/background_task/mod.rs
+++ b/dial9-tokio-telemetry/src/background_task/mod.rs
@@ -185,8 +185,11 @@ impl ProcessError {
 #[non_exhaustive]
 pub enum ProcessErrorKind {
     /// The processor hit an `std::io::Error`.
+    #[non_exhaustive]
     Io(std::io::Error),
-    /// An S3 transfer manager error.
+
+    /// An error transferring data off the host
+    #[non_exhaustive]
     Transfer {
         /// Underlying error source.
         source: Box<dyn std::error::Error + Send + Sync>,
@@ -290,6 +293,25 @@ pub trait SegmentProcessor: Send {
 /// ([`gzip`](Self::gzip), [`write_back`](Self::write_back),
 /// [`s3`](Self::s3), [`symbolize`](Self::symbolize)); custom processors
 /// are added with [`pipe`](Self::pipe).
+///
+/// # Example
+///
+/// ```ignore
+/// struct Logger;
+/// impl SegmentProcessor for Logger {
+///     fn name(&self) -> &'static str { "logger" }
+///     fn process(&mut self, data: SegmentData)
+///         -> Pin<Box<dyn Future<Output = Result<SegmentData, ProcessError>> + Send + '_>>
+///     {
+///         Box::pin(async move {
+///             println!("segment {} ({} bytes)", data.segment().index(), data.bytes().len());
+///             Ok(data)
+///         })
+///     }
+/// }
+///
+/// builder.with_custom_pipeline(|p| p.pipe(Logger).gzip().write_back())
+/// ```
 #[must_use]
 pub struct PipelineBuilder {
     processors: Vec<Box<dyn SegmentProcessor>>,
--- a/dial9-tokio-telemetry/src/background_task/mod.rs
+++ b/dial9-tokio-telemetry/src/background_task/mod.rs
@@ -6,6 +6,8 @@ pub(crate) mod pipeline_metrics;
 pub mod s3;
 pub(crate) mod sealed;
 
+pub use sealed::SealedSegment;
+
 use crate::metrics::{Operation, SegmentProcessMetrics, SegmentProcessMetricsGuard};
 use crate::rate_limit::rate_limited;
 use futures_util::FutureExt;
@@ -23,10 +25,8 @@ pub(crate) const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(1);
 
 /// Configuration for the in-process worker pipeline.
 ///
-/// Only `trace_path` and `s3` are required. Optional fields:
-///
-/// - `poll_interval`: how often to check for sealed segments (default: 1 second)
-/// - `client`: pre-built `aws_sdk_s3::Client` for custom credentials or endpoints
+/// The pipeline is composed of a sequence of [`SegmentProcessor`]s supplied
+/// via `processors`. When none are provided the worker runs no processing.
 #[derive(bon::Builder)]
 #[builder(on(String, into))]
 pub struct BackgroundTaskConfig {
@@ -36,19 +36,9 @@ pub struct BackgroundTaskConfig {
     /// How often the worker checks for sealed segments. Defaults to 1 second.
     #[builder(default = DEFAULT_POLL_INTERVAL)]
     poll_interval: Duration,
-    /// S3 upload configuration. When `None`, the worker symbolizes and
-    /// gzip-writes back to disk without uploading.
-    #[cfg(feature = "worker-s3")]
-    s3: Option<s3::S3Config>,
-    /// Pre-built S3 client. When provided, the worker uses this client
-    /// instead of building one from `aws_config::load_defaults`.
-    /// Region auto-detection still applies unless `region` is set on `S3Config`.
-    #[cfg(feature = "worker-s3")]
-    client: Option<aws_sdk_s3::Client>,
-    /// When true, run the symbolize processor on each segment.
+    /// The processor pipeline executed for each sealed segment, in order.
     #[builder(default)]
-    #[allow(dead_code)]
-    symbolize: bool,
+    processors: Vec<Box<dyn SegmentProcessor>>,
     /// Metrics sink. Defaults to [`DevNullSink`](metrique_writer::sink::DevNullSink).
     #[builder(default = metrique_writer::sink::DevNullSink::boxed())]
     metrics_sink: BoxEntrySink,
@@ -94,12 +84,6 @@ impl BackgroundTaskConfig {
             }
         }
     }
-
-    /// S3 upload configuration.
-    #[cfg(feature = "worker-s3")]
-    pub fn s3(&self) -> Option<&s3::S3Config> {
-        self.s3.as_ref()
-    }
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -142,16 +163,35 @@ impl std::fmt::Debug for SegmentData {
...
+}
+
+/// Kind of failure reported by a [`SegmentProcessor`].
 #[derive(Debug)]
-pub(crate) enum ProcessErrorKind {
+#[non_exhaustive]
+pub enum ProcessErrorKind {
+    /// The processor hit an `std::io::Error`.
     Io(std::io::Error),
+    /// An S3 transfer manager error.
```
rcoh: "An error transferring data off the host"

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -142,16 +163,35 @@ impl std::fmt::Debug for SegmentData {
...
+    }
+}
+
+/// Kind of failure reported by a [`SegmentProcessor`].
 #[derive(Debug)]
-pub(crate) enum ProcessErrorKind {
+#[non_exhaustive]
+pub enum ProcessErrorKind {
+    /// The processor hit an `std::io::Error`.
     Io(std::io::Error),
```
rcoh, line 186: throw `#[non_exhaustive]` on the variants as well

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -243,31 +283,90 @@ pub(crate) trait SegmentProcessor: Send {
...
+            processors: Vec::new(),
+        }
+    }
 
+    pub(crate) fn into_processors(self) -> Vec<Box<dyn SegmentProcessor>> {
+        self.processors
+    }
+
+    /// Append a user-supplied [`SegmentProcessor`] to the pipeline.
+    pub fn pipe<S>(mut self, processor: S) -> Self
```
rcoh, line 329: `pipe` seems ok. `chain` would also match with the stdlib method for chaining iterators but its not exactly the same concept. `pipe` is the best name so far anyway

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -243,31 +283,90 @@ pub(crate) trait SegmentProcessor: Send {
...
-/// Build the processor pipeline based on config flags and available features.
-async fn build_pipeline(_config: &mut BackgroundTaskConfig) -> Vec<Box<dyn SegmentProcessor>> {
-    let mut pipeline: Vec<Box<dyn SegmentProcessor>> = Vec::new();
+/// Closure-scoped builder for assembling a custom processor pipeline.
+///
+/// Obtained via `with_custom_pipeline(|p| ...)` on the runtime builder.
+/// Built-in processors are reachable through dedicated methods
+/// ([`gzip`](Self::gzip), [`write_back`](Self::write_back),
+/// [`s3`](Self::s3), [`symbolize`](Self::symbolize)); custom processors
+/// are added with [`pipe`](Self::pipe).
```
rcoh, line 292: adding a short example here would be helpful

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -779,7 +958,37 @@ impl SegmentProcessor for S3PipelineUploader {
...
         Box::pin(async move {
-            if !self.circuit_breaker.should_attempt() {
+            // Lazy init: clone the config + client and run `initialize`
+            // without mutating `self.state`. If the init future panics or
+            // is cancelled mid-await, the worker's outer `catch_unwind`
+            // recovers and `self.state` stays `Pending`, so the next
+            // segment will retry. Mutating before the await would leave
+            // the uploader stuck in a transient state forever.
+            if let S3UploaderState::Pending { s3_config, client } = &self.state {
+                let cfg = s3_config.clone();
```
rcoh, line 996: this makes sense. Having a method to smoke test that triggers this becomes even more important. Something like `async fn smoke_test()` on the telemetry handle would be good. No need to add it in this PR but we should tackle that soon after.

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -839,8 +839,31 @@ pub struct NoTracePath;
 #[derive(Debug)]
 pub struct HasTracePath;
 
+/// Marker: no pipeline strategy has been chosen yet. From this state the
+/// builder can transition to either S3 (via `with_s3_uploader`) or a custom
+/// pipeline (via `with_custom_pipeline`).
+#[derive(Debug)]
+pub struct PipelineUnset;
```
rcoh, line 878: `#[non_exhaustive]` on these I think? There is no reason to ever create them since they are "type only"

dial9-tokio-telemetry/src/legacy_config.rs
```diff
@@ -16,10 +16,30 @@ use std::path::PathBuf;
...
 
 use crate::telemetry::recorder::{
-    HasTracePath, TelemetryGuard, TracedRuntime, TracedRuntimeBuilder,
+    HasTracePath, PipelineUnset, TelemetryGuard, TracedRuntime, TracedRuntimeBuilder,
 };
 use crate::telemetry::writer::RotatingWriter;
 
+/// Type-erased terminal step for a [`TracedRuntimeBuilder`]: hides the
+/// pipeline-mode marker `M` so [`Dial9Config`] can stay non-generic.
+trait BuildTracedRuntime: Send {
```
rcoh, line 25: what do we gain from having this trait?

rcoh (commented):
overall looks great! some small comments

dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -111,21 +95,58 @@ impl BackgroundTaskConfig {
...
+    pub fn bytes_mut(&mut self) -> &mut Vec<u8> {
+        &mut self.bytes
+    }
+
+    /// Take ownership of the payload bytes, leaving an empty `Vec` in its place.
+    pub fn take_bytes(&mut self) -> Vec<u8> {
+        std::mem::take(&mut self.bytes)
+    }
+
+    /// Replace the payload bytes.
```
rcoh: I think it may be worth doing a setup PR to this one where we change the bytes management to be `Bytes` and possibly add the ability to do SegmentedBytes. Not a huge issue because this API is not going to have a lot of consumers and we can alter it in 0.4

rcoh:
ah one last thing, I think from reading (but easiest to write tests I think), that in some cases we aren't setting SegmentMetadata and in others we are. worth evaluating

dial9-tokio-telemetry/src/legacy_config.rs
Fluzko: The main idea is to erase the pipeline-mode marker `M` from `TracedRuntimeBuilder<HasTracePath, M>`.
Concretely, we gain `Dial9Config` stays non-generic (`pub struct Dial9Config(Inner)` with no `<M>`). 

Otherwise the type would be breaking because you'd have to do:
```rust
fn my_config() -> Dial9Config<PipelineS3<GzipProcessor>> {
  // ...
}
```
Or do `Dial9Config<_>` and hope inference works as expected. This is one of the drawbacks of the typestate approach.
Open to discuss around a better approach if you see a better one.

dial9-tokio-telemetry/src/background_task/mod.rs
Fluzko: created #361

Fluzko:
> ah one last thing, I think from reading (but easiest to write tests I think), that in some cases we aren't setting SegmentMetadata and in others we are. worth evaluating

Good catch, I've created a knob for users to manage the metadata (always with _you're in full-control_ in mind) and added a few tests let me know if this is enough or would you like to go further [19142ca](https://github.com/dial9-rs/dial9-tokio-telemetry/pull/355/commits/19142cad7c735d061bf1a1825dbe99484e989871)

dial9-tokio-telemetry/src/background_task/mod.rs
Fluzko: I've done it here [7413524](https://github.com/dial9-rs/dial9-tokio-telemetry/pull/355/commits/74135248a4c513efab91a6a9067a9bd60aa3585a) since was pretty achievable with not much work. It pretty much introduces a `Payload` which handles the conversion

rcoh (approved):

PR #360, opened 2026-05-06
benches: add iai-callgrind PR gate, retire criterion CI

### LINKED ISSUE #325: Make current benchamarks runnable on CI without being noisy
We've recently disabled the bencher jobs on CI https://github.com/dial9-rs/dial9-tokio-telemetry/pull/296 due to them being quite noisy not yet providing much value. 

We could something like https://docs.rs/iai-callgrind/latest/iai_callgrind/ for micro benches in a way that will run more consistently on CI compared to the current setup on top of gh actions.

For integration and more e2e benchmarks we might want to use a dedicated EC2 or similar.

yulnr:
closes #325

Replaces noisy wall-clock benches on shared ubuntu-latest with a deterministic instruction-count gate. iai-callgrind runs each bench once under callgrind, so same code + same input gives byte-identical counts. 
A 1% threshold should catch real regressions without false fires, confirmed Bencher alarms for a %1.03 regression (see bencher report history below)

**Unoutched** (will follow up on, likely with a dedicated runner):

- `overhead_bench` (TCP echo, 50 clients × 60s) and `e2e_workload` (mixed CPU+IO), kept disabled. both measure throughput / latency under real IO and probably need pinned hardware to be useful. (and would probably be too slow with iai_callgrind for PR checks)
- left the multithread bench on `tracing_layer_bench`, with valgrind serializing thread execution we can't observe real world wall-time scaling under contention. (could still track per-thread instruction counts if wanted)

**EDIT**: This now gates all 6 iai-callgrind measures the adapter emits (instructions, L1/RAM hits, total r+w, estimated cycles at 1%; LL hits at 5% - small absolute counts amplify noise). 
It's the same cost as gating only instructions, and the adapter posts all 6 regardless, so it seems simpler to actually track them than to do some workaround to have Bencher not complain about un-thresholded ones (write a custom json adapter, update some settings on Bencher UI). If we find it to noisy we drop back to just instruction counts and handle this more manually.

**Misc.**
Split into benchamrks.yml and benchamrks-pr.yml so that we don't get the noisy skipped `main` jobs on every PR checks

Last job ran in 2m 42s, not too bad.


jv1i:
Nice, reverting one of the latest optimizations we did to `FxHasher` shows the above alert. (won't see the commit because I force-pushed to not mess with the git history of that file)

Edit: bencher comment got edited after each run so you can check the history if you want to see the alert.

rcoh (approved):
awesome!

PR #362, opened 2026-05-06
Memory profiling design

rcoh:


dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+  background `SymbolizeProcessor`.
+- **Install-once, captured handle.** `MemoryProfiler::install(handle)`
+  sets a process-global static that the allocator reads. The captured
+  handle routes all sampled events through the central recorder —
+  allocations on any thread (tokio workers, blocking pool, user OS
+  threads, library-spawned threads) get recorded identically.
+
+## Non-goals
+
+- Tracking every allocation. We sample.
```
jlizen, line 51: Non-goal of, seeing contents of sampled memory?

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+- Replacing heaptrack / valgrind. Those record every allocation and are
+  fine for dev-time analysis. Dial9 is for production, always-on.
+- Detecting use-after-free, double-free, or other memory safety bugs.
+  That's what Miri and ASAN are for.
+- Tracking stack allocations or `mmap`-backed memory outside the
```
jlizen, line 62: Fine to be out of scope, but we'll want to have an evolution path to it.

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+    while next <= 0 {
+        next += draw_exponential(sample_rate_bytes);
+    }
+    next_sample_bytes.set(next);
+
+    record_sample(size, capture_stack());
+}
+```
+
+Yes — this matches the structure you sketched. Two notes on that sketch:
```
jlizen: Needs cleanup

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+
+- The `while` loop is important. A single huge allocation (or a very low
+  sample rate) can push `next_sample_bytes` below zero by more than one
+  draw's worth. We need to keep drawing until the counter is positive
+  again so we don't sample the *next* allocation immediately too.
+- `remaining` has to be `i64` (or signed `isize`). Subtracting `usize`
```
jlizen, line 105: might be worth noting this locally where we first declare the i64

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+  zero-cost wrapper.
+- Stacks tied to the worker thread + task that performed the allocation, so
+  allocation hot paths can be attributed to specific tasks or poll ranges.
+- Optional live-set tracking for leak detection. Off by default — it adds
+  a per-allocation hashmap lookup and a concurrent hashmap.
+- No symbolication on the hot path. Addresses flow through the existing
+  background `SymbolizeProcessor`.
+- **Install-once, captured handle.** `MemoryProfiler::install(handle)`
+  sets a process-global static that the allocator reads. The captured
+  handle routes all sampled events through the central recorder —
```
jlizen: i might split "install once captured handle" and "instruments all threads, not just tokio workers" as separate goal points

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+  allocations on any thread (tokio workers, blocking pool, user OS
+  threads, library-spawned threads) get recorded identically.
+
+## Non-goals
```
jlizen, line 49: Are there any non-goals mentioning here wrt randomness properties?

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+
+The design mirrors jemalloc's and Go's allocation profilers: **geometric
+(Poisson) sampling** keyed on allocation size. The expected sampling rate
+is 1 sample per N bytes allocated, regardless of object size distribution.
+This gives unbiased size-weighted profiles at bounded overhead.
+
+## Goals
+
+- Always-on in production, with sub-1% overhead at default sampling rates.
+- Works with any `GlobalAlloc` (system, jemalloc, mimalloc, etc.) via a
```
jlizen, line 31: Are we giving up anything by not delegating to jemalloc/mimalloc's inbuilt profiling?

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+pointer unwinder we already use for CPU profiling. The viewer and analysis
+toolkit gain allocation flamegraphs, per-task allocation totals, and leak
+detection.
+
+The design mirrors jemalloc's and Go's allocation profilers: **geometric
+(Poisson) sampling** keyed on allocation size. The expected sampling rate
+is 1 sample per N bytes allocated, regardless of object size distribution.
+This gives unbiased size-weighted profiles at bounded overhead.
+
+## Goals
```
jlizen, line 28: Do we want to take a goal related to analysis toolkit/viewer support? Or should that be out of scope for now?

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+
+**Known bias: `MAX_FRAME_SIZE`.** The underlying unwinder stops walking
+if `saved_fp - fp > 256 KiB` (this cap rejects wild pointers that happen
+to be above `fp` but aren't real frames). For CPU profiling the
+occasional truncation caused by a large on-stack future or struct is
+rare enough to ignore; for *allocation* profiling the bias is
+systematic — allocations performed inside functions with unusually
+large stack frames (large `Box::pin(future)` state machines, large
+`[u8; N]` locals) will consistently have their stacks cut off at the
+big frame. We accept this for the MVP and document it as a known
```
jlizen: So the future mitigation we see is to just dial up the max size? Is there a reason not to give headroom in the first place?

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+limitation. Raising the cap or making it configurable in the unwinder
+is future work.
+
+**Why a handle-returning `install`, not auto-install-on-first-capture?**
```
jlizen, line 251: Big ++ to active `install` call rather than lazy capture, much simpler.

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+and applies it uniformly to every `AllocEvent` in that segment when
+computing unbiased byte totals. Per-event storage of the weight
+would add 1–9 LEB128 bytes to every sampled allocation for data
+that never varies within a file — wasteful at steady-state trace
+rates.
+
+**Can the rate change at runtime?** Not within a single trace file.
+If we later want a `set_sample_rate_bytes` API, it will:
+
+1. Atomically update the rate held in `MemoryProfilerInner`.
```
jlizen, line 359: Seems fine as an evolution path

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+were sampled:
+- Old sampled + new sampled: emit free-of-old, emit alloc-of-new.
+- Old unsampled + new sampled: emit alloc-of-new only.
+- Old sampled + new unsampled: emit free-of-old only.
+- Neither sampled: emit nothing.
+
+Crucially, jemalloc treats the new side as a *fresh* sampling decision
+even when `realloc` doesn't move the pointer. Same as a new `malloc`.
+
+**Our plan: follow jemalloc.** Treat `realloc(p, n_bytes)` as:
```
jlizen, line 450: ++ to sticking to convention

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+    Dial9Allocator::new(tikv_jemallocator::Jemalloc);
+```
+
+The wrapper is zero-cost when memory profiling isn't attached (see §6).
+
+---
+
+## 5. Ring-buffer hand-off (considered, rejected for MVP)
+
+Your idea: have the allocator hook push raw data to a ring buffer, then
```
jlizen: needs cleanup

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+the ring buffer design.
+
+**The catch:** the ring buffer entries need to store the captured
+stack (N × u64 addresses) plus size, timestamp, `tid`. That's
+~150-250 bytes per sample. A lock-free SPSC queue per thread sized at
+256 entries = ~50 KiB per thread. Doable but non-trivial memory.
+
+**Decision for MVP: don't build the ring buffer.** Use a thread-local
+reentrancy guard (§6) and the existing thread-local buffer
+infrastructure. Revisit if:
```
jlizen: Thanks for specifying how we are evaluating building the ring buffer. Is that evaluation going to be done as part of initial implementation?

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+```
+
+The flag is checked with a single relaxed TLS load.
+
+**Consequence:** we don't see dial9's own internal allocations in the
+profile. 
+
+### Consequence: dropped events
+
+Sampled allocations that happen *during* another sampled allocation's
```
jlizen, line 665: Seems fine, having the metric is good

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+**Interaction with the `tracing-layer` feature.** The guard drops
+*any* recording work that happens inside a sampled allocation's stack
+capture + encoding, not just dial9's internal allocations. If a user
+has both `tracing-layer` and memory profiling enabled and a
+`tracing::span!` enter/exit happens to land on a thread that is
+currently ~1 µs deep in `on_alloc`'s hot path (stack capture +
+encode), that span event is dropped. At 2000 samples/sec and typical
+tracing-span rates the collision rate is still tiny (<0.01%), but
+this is a real and non-obvious behavior that operators reviewing
+dropped-event counters need to know about. Noted here because it's
```
jlizen: which again will show up due to `dropped_samples` yes?

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+pub enum InstallError {
+    AlreadyInstalled,
+    Unwinder(std::io::Error),
+}
+
+/// Dropping this guard does NOT uninstall the profiler. The static
+/// state lives until process exit. The guard exists to make the
+/// API familiar (RAII) and to hold a lifetime if we add a pause/
+/// resume later.
+pub struct MemoryProfilerGuard { _private: () }
```
jlizen, line 817: ++ to shipping with the mostly useless guard

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+    timestamp_ns: u64,
+    // No stack frames here. The per-live-alloc memory cost of storing
+    // a full call stack (64 × 8 B) dominates total liveset overhead.
+    // Stack-attributed leak analysis joins FreeEvent→AllocEvent by
+    // `addr` inside a single retained segment instead.
+}
+
+// Shared, concurrent: all threads alloc; all threads free.
+// Lives inside MemoryProfilerInner.
+liveset: DashMap<usize, LivesetEntry>,
```
jlizen: I worry about using DashMaps inside an allocator since it is so deadlock prone.

Did you look at crossbeam's new skiplist library? That might be less footgun prone and runs leaner, if the API works for us.

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+uncontended).
+
+Per **any** dealloc (not just sampled): one hash + one shard lock +
+map lookup. If the pointer isn't in the liveset (typical — 99.9% of
+deallocs), this is a lock+lookup+miss on a cold key. ~40-60ns on
+DashMap.
+
+Two optimizations to consider:
+
+1. **Bloom filter in front of DashMap.** Sampled `alloc` inserts into
```
jlizen: You know I love my bloom filters. I'd look at how much complexity this actually adds, seems like a good best practice to me.

Though, I guess crossbeam-skiplist might make lookups cheaper anyway.

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+
+Target: <1% at default settings (512 KiB sample rate, no liveset).
+
+Per-allocation fast path (unsampled, ~99.9% of calls):
+- 1 atomic load of `ACTIVE` pointer (~1ns)
+- 1 TLS load of `IN_ALLOC_HOOK` (~2ns)
+- 1 subtract + compare on per-thread `next_sample_bytes` (~1ns)
+- Return
+
+Total: ~5ns added per alloc. For a service doing 1M allocs/sec that's
```
jlizen, line 1109: This would be good to contextualize with typical overall allocator latencies

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+0.5% CPU. Most services do far fewer.
+
+Per-sampled allocation (~0.1% of calls at 512 KiB):
+- Stack capture (~110 ns for 20 frames, measured warm-cache; ~200–400
+  ns likely in production with cold caches)
+- `record_event` into TL buffer (~100-200ns)
+- Optional timestamp call (~25ns)
+- Liveset insert if enabled (~30-50ns)
+
+At 2000 samples/sec: well under 1ms/sec CPU, comfortably under budget.
```
jlizen: 1ms/sec -> 1% cpu?

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+### Allocation flamegraph
+
+Same UX as CPU flamegraph — click a poll, or shift-drag a time range,
+see a flamegraph. The sample value is
+`AllocEvent.size * weight_correction(AllocEvent.size, sample_rate)`,
+where `sample_rate` is read once per segment from `SegmentMetadata`.
+Code lives alongside the existing flamegraph, driven by a new
+aggregator that consumes `AllocEvent` instead of `CpuSample`.
+
+### Per-task allocation chart
```
jlizen, line 1163: this will be great. might be getting to be time for collapsible panes... or some other way of avoiding data overload

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+    task = poll.task_id
+    agg[task] += e.size * weight_correction(e.size, sample_rate)
+```
+
+This is the same join the viewer already does between `CpuSample` and
+polls. Allocations with no containing poll go into an "unassociated"
+bucket — common for background threads, first allocations on new
+workers, etc.
+
+### Leak view (liveset only)
```
jlizen, line 1182: great

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+  `tests/memory_profiling_*.rs` files pick up a fresh state
+  automatically. This is the approach the existing S3 worker tests
+  use (see `tests/s3_integration.rs`).
+- Contributors who run tests via `cargo test` (which reuses the
+  process across tests within a binary) need to structure tests
+  within a single file so they share a consistent profiler
+  configuration, or move into separate `tests/` files. This is
+  called out in the tests' module docs.
+
+### Test categories
```
jlizen, line 1249: I was expecting to see some shuttle mention here?

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+   compute the net-bytes delta.
+
+---
+
+## 13. Open questions
+
+1. **Do we want per-alloc allocator latency?** Wrapping `self.0.alloc`
+   with a `clock_monotonic_ns` pair gives allocator latency, which is
+   interesting for jemalloc/mimalloc fragmentation investigations.
+   Cost: +50ns per sampled alloc (two vDSO calls). Easy to add as a
```
jlizen, line 1298: Seems nice, but maybe at an even lower sample rate?

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+   field later.
+
+2. **MUSL / static builds.** The `safe_load` trampoline + SIGSEGV chain
```
jlizen, line 1304: Are you planning to test this for release? I'd err on the side of conditionally compiling out support if not

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+   works on glibc. Need to verify on musl (Alpine containers). If it
+   doesn't, fall back to the `backtrace` crate on init — slower but
+   allocation-free if called with a preallocated buffer.
+
+3. **Multi-runtime selection.** With a single captured handle, all
```
jlizen, line 1312: Yeah let's wait for a customer, this seems solveable.

jlizen (commented):
I'm on board with the design, my main load bearing question is around usage of dashmap

dial9-tokio-telemetry/design/memory-profiling.md
rcoh: Worth evaluating, yeah I don't like dashmap

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+    unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
+        hook::on_dealloc(ptr, layout.size());
+        unsafe { self.0.dealloc(ptr, layout) };
+    }
+
+    unsafe fn realloc(
+        &self, ptr: *mut u8, old_layout: Layout, new_size: usize
+    ) -> *mut u8 {
+        hook::on_dealloc(ptr, old_layout.size());
+        let new_ptr = unsafe { self.0.realloc(ptr, old_layout, new_size) };
```
jlizen, line 520: if `self.0.realloc` returns null, that means that the former pointer is still live. But we already record it as freed above. We need to check for non-null first.

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+```rust
+thread_local! {
+    static IN_ALLOC_HOOK: Cell<bool> = const { Cell::new(false) };
+}
+
+fn on_alloc(ptr: *mut u8, size: usize) {
+    if IN_ALLOC_HOOK.with(|f| f.replace(true)) {
+        return; // reentrant — skip
+    }
+    // ... sampling decision, stack capture, event emission ...
```
jlizen, line 653: if we have a panic here, we will keep the flag as `true` and silently drop all other samples in the thread.

We should instead have a `struct ReentrancyGuard` that clears on drop

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+Incidental benefit: skipping the `current()` TLS lookup saves a few
+ns per sampled alloc. Not the reason, but nice.
+
+### `MemoryProfiler::install()` flow
+
+```rust
+static ACTIVE: OnceLock<MemoryProfilerState> = OnceLock::new();
+
+impl MemoryProfiler {
+    pub fn install(
```
jlizen, line 782: this uses an underlying `unsafe install_handler` API, we should document why this method has a safe API

dial9-tokio-telemetry/design/memory-profiling.md
```diff
@@ -0,0 +1,1224 @@
...
+allocations performed by dial9 itself while recording an event. Without
+a guard, we recurse into `on_alloc` from inside `on_alloc` and either
+deadlock or blow the stack.
+
+Concrete cases:
+- `Encoder::intern_string` calls `s.to_string()` and `HashMap::insert`
+  — both allocate.
+- Growing the thread-local buffer's `Vec<u8>` allocates.
+- The liveset `DashMap::insert` allocates.
+- tracing logs allocate.
```
jlizen, line 616: also dial9's own SIGPROF handler could allocate

and first initialization run of the `thread_local!` blocks

dial9-tokio-telemetry/design/memory-profiling.md
jlizen, line 49: Tracking during process teardown as out of scope?

dial9-tokio-telemetry/design/memory-profiling.md
jlizen: ref https://docs.rs/crossbeam-skiplist/latest/crossbeam_skiplist/

jlizen: another option worth looking at is `papaya`

dial9-tokio-telemetry/design/PROGRESS-memory-profiling-cr.md
```diff
@@ -0,0 +1,79 @@
+# Memory Profiling Design - CR Feedback Progress
```
jlizen: This probably doesn't need to be checked in.

jlizen (approved):
Changes look good to me

PR #366, opened 2026-05-07
feat!: Add TAG_SCHEMA_ANNOTATIONS frame and SchemaEntry::annotations

jlizen:
Related: dial9-rs/dial9-tokio-telemetry#346 (metrique integration design)

### Summary

Adds wire-format support for schema-level annotations, the first of two trace-format extensions needed for metrique integration. Annotations carry `(field_index, key, value)` tuples that decoders can surface as per-field metadata: units today (`metrique.unit` = `microseconds`), display hints, semantic-convention labels, privacy tiers, and similar additions tomorrow. Fields without annotations cost zero bytes on the wire.

The annotation payload rides in a new top-level frame tag (`TAG_SCHEMA_ANNOTATIONS = 0x07`) that follows the schema frame it annotates. Framing decisions:

1. Annotations live in a separate frame rather than inside the schema frame body. We could inline them (new decoders handle both old and new formats fine), but a separate frame keeps the schema frame codec unchanged, avoids touching the JS parser's schema decode path, and allows multiple annotation frames per schema (the decoder accumulates). The separation also cleanly models the concern split: schema defines structure, annotations carry metadata.

2. Multiple annotation frames per schema are permitted. The decoder accumulates; the encoder emits one frame per schema registration but the format does not mandate it. I suspect that this will prove useful over time if we want to do clever things around metadata stickiness.

3. `FieldAnnotation` is intentionally generic (`String` key, `String` value) so one mechanism serves units, display hints, semantic conventions, privacy labels, and future metadata without further format churn.

### Tradeoffs

- **Separate frame vs inlined**: a separate frame adds a tag to the format but keeps the schema codec simple and the JS parser untouched. Since the viewer ships in lockstep with producers, backward compat with old decoders is not a concern; the real constraint is new decoders reading old traces (which works: no annotation frame means empty annotations).

- **Generic key-value strings**: flexible at the cost of being untyped. Consumers interpret by key (`metrique.unit`, `dial9.kpi`, etc.). A typed hierarchy was considered but would lock us into today's uses and fight the "add new metadata without format churn" requirement.

- **Breaking change**: This will need to go in `0.4.x` due to the new `annotations` field on `SchemaEntry` and various other places. I plan pull off the bandaid and tag that `#[non_exhaustive] and provide some better constructors, but I'll do it in a separate PR that adds support for non-stringly maps and lists.

### Testing

Round-trip of a schema with annotations; verify annotations attach to the right schema and field index. Empty-annotation schemas produce byte-parity output with the existing format. Interleaving scenarios: two schemas in one trace, one annotated, one not; decoder routes annotations correctly. Silent truncation: truncate the byte stream before the annotation frame; decoder halts cleanly at end of stream. Unknown type_id in an annotation frame: decoder skips leniently and continues. All existing trace-format round-trips continue to pass.

### Out of scope

- Annotation rendering in the dial9 viewer (follow-up).
- `#[non_exhaustive]` on `FieldType`, `FieldDef`, `SchemaEntry`: these land with the field-types PR.
- TraceField impls, metrique adapter.
- Format version bump, Cargo.toml change, CHANGELOG entry (handled at release time).

### DIFF
--- a/dial9-trace-format/SPEC.md
+++ b/dial9-trace-format/SPEC.md
@@ -38,8 +38,7 @@ Every frame begins with a 1-byte tag:
 | `0x03` | String Pool |
 | `0x04` | Stack Pool |
 | `0x05` | Timestamp Reset |
-| `0x06` | *(reserved)* |
-| `0x07` | Schema Annotations |
+| `0x06` | Schema Annotations |
 
 Unknown tags **must** cause the decoder to stop (the stream cannot be advanced without knowing the frame size).
 
@@ -151,14 +150,14 @@ Total: **9 bytes**.
 
 After decoding this frame, the decoder sets `timestamp_base_ns = timestamp_ns`. The next event's `timestamp_delta_ns` is relative to this new base.
 
-### Schema Annotations Frame (`0x07`)
+### Schema Annotations Frame (`0x06`)
 
 Carries per-field metadata for a previously-registered schema. Annotations are key-value string pairs attached to individual fields by index. A schema with no annotations produces no annotation frame.
 
 | Field | Type | Description |
 |-------|------|-------------|
-| tag | u8 | `0x07` |
-| type_id | u16 | References a schema's `type_id` |
+| tag | u8 | `0x06` |
+| type_id | LEB128 u64 | References a schema's `type_id` (varint-encoded to allow future overflow beyond u16) |
 | count | u16 | Number of annotation entries |
 | entries | [FieldAnnotation; count] | Annotation entries (see below) |
 
--- a/dial9-trace-format/SPEC.md
+++ b/dial9-trace-format/SPEC.md
@@ -39,6 +39,7 @@ Every frame begins with a 1-byte tag:
 | `0x04` | Stack Pool |
 | `0x05` | Timestamp Reset |
 | `0x06` | *(reserved)* |
+| `0x07` | Schema Annotations |
 
 Unknown tags **must** cause the decoder to stop (the stream cannot be advanced without knowing the frame size).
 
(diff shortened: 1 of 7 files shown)


dial9-trace-format/SPEC.md
```diff
@@ -39,6 +39,7 @@ Every frame begins with a 1-byte tag:
 | `0x04` | Stack Pool |
 | `0x05` | Timestamp Reset |
 | `0x06` | *(reserved)* |
+| `0x07` | Schema Annotations |
```
rcoh: you can use 0x06

dial9-trace-format/src/codec.rs
```diff
@@ -78,6 +79,10 @@ pub(crate) enum Frame {
     StringPool(Vec<PoolEntry>),
     StackPool(Vec<StackPoolEntry>),
     TimestampReset(u64),
+    SchemaAnnotations {
+        type_id: WireTypeId,
+        annotations: Vec<FieldAnnotation>,
```
rcoh, line 83: I realize this is already the pattern, but its a little annoying that we need to allocate a vec here. presumably this happens sufficiently rarely right now

dial9-trace-format/src/schema.rs
```diff
@@ -7,6 +7,34 @@ use crate::codec::WireTypeId;
...
+/// after the schema frame they belong to. They carry metadata such as units,
+/// display hints, or semantic-convention labels.
+#[non_exhaustive]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FieldAnnotation {
+    /// Index of the field this annotation applies to (0-based, matching the
+    /// field order in [`SchemaEntry::fields`]).
+    pub field_index: u16,
+    /// Annotation key (e.g. `"metrique.unit"`).
+    pub key: String,
```
rcoh: should almost certainly be `Cow`? I imagine this is static in practice

dial9-trace-format/src/schema.rs
```diff
@@ -7,6 +7,34 @@ use crate::codec::WireTypeId;
...
 use crate::types::FieldType;
 
+/// A per-field annotation carrying arbitrary key-value metadata.
+///
+/// Annotations are emitted in a separate frame (`TAG_SCHEMA_ANNOTATIONS`)
+/// after the schema frame they belong to. They carry metadata such as units,
+/// display hints, or semantic-convention labels.
+#[non_exhaustive]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FieldAnnotation {
```
rcoh, line 18: I don't love that we have our specific types in the public UI that would make it hard to, e.g. switch to SmallStr in the future

dial9-trace-format/SPEC.md
```diff
@@ -150,6 +151,31 @@ Total: **9 bytes**.
...
 After decoding this frame, the decoder sets `timestamp_base_ns = timestamp_ns`. The next event's `timestamp_delta_ns` is relative to this new base.
 
+### Schema Annotations Frame (`0x07`)
+
+Carries per-field metadata for a previously-registered schema. Annotations are key-value string pairs attached to individual fields by index. A schema with no annotations produces no annotation frame.
+
+| Field | Type | Description |
+|-------|------|-------------|
+| tag | u8 | `0x07` |
+| type_id | u16 | References a schema's `type_id` |
```
rcoh: lets use u64 here to allow for an eventual support of overflow type_ids (we can say that a type_id of u16::MAX overflows into the next byte) but since these reset frames are sufficiently rare, there is no reason not to use a varint u64

rcoh (approved):
Looks good. My main question is how we ensure that callers write the annotations any time that a schema gets written and what we imagine the external API for this ends up looking like.

This seems like a change we can safely release as 0.4 of the format without bumping the main crate

dial9-trace-format/src/codec.rs
jlizen: Pre-existing pattern, cold path, agreed that it's irritating

I removed it at one point but wasn't happy with the design

dial9-trace-format/src/schema.rs
jlizen: i'll make them opaque, thanks

PR #369, opened 2026-05-07
task dumps: Switch to poisson sampling & libunwind

rcoh:
Two improvements from the first cut:
1. Switch to poisson sampling (we'll also use this in memory profiling)
2. use libunwind directly instead of backtrace::trace

### DIFF
--- a/dial9-tokio-telemetry/src/task_dumped.rs
+++ b/dial9-tokio-telemetry/src/task_dumped.rs
@@ -1,5 +1,5 @@
 //! `TaskDumped<F>` wraps a future and captures async backtraces at yield
-//! points when the task stays idle longer than the configured threshold.
+//! points using geometric (Poisson) sampling keyed on idle duration.
 //!
 //! This wrapper is intentionally separate from [`crate::traced::Traced`]: the
 //! wake-event capture in `Traced` runs on every instrumented spawn regardless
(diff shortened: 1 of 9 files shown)


dial9-tokio-telemetry/src/task_dumped.rs
```diff
@@ -115,17 +167,19 @@ impl<F: Future> Future for TaskDumped<F> {
                 *this.task_id,
                 this.pending_capture_ts.unwrap().get(),
             );
+            // Redraw: loop in case the draw is smaller than the deficit
+            // (rare, but possible for very large idles / small means).
+            while *this.next_sample_ns <= 0 {
+                *this.next_sample_ns += this.rng.draw_exponential_ns(*this.sample_mean_ns);
```
jlizen: this could be a pretty tight loop for small `sample_mean_ns`. Would it suffice to have just a single pass:


```
  if *this.next_sample_ns <= 0 {
      *this.next_sample_ns = this.rng.draw_exponential_ns(*this.sample_mean_ns);
  }
```
?

dial9-tokio-telemetry/src/task_dumped.rs
```diff
@@ -205,39 +259,15 @@ impl FrameBuf {
...
-///
-/// Held behind the `backtrace` crate's process-wide mutex. This will be optimized in
-/// https://github.com/dial9-rs/dial9-tokio-telemetry/issues/357. This is not a
-/// complete blocker for the MvP because no code is utilizing the symbolizing from
-/// `backtrace` so the mutex is not actually held for long periods of time. Nevertheless,
-/// this will be a fast follow.
+/// Walk the stack, collecting instruction pointers between `leaf_addr` and
+/// `root_addr`. Calls `_Unwind_Backtrace` directly via [`crate::unwind`],
+/// bypassing the `backtrace` crate's process-wide mutex.
 fn capture_frames(
```
jlizen, line 258: is this worth just inlining?

dial9-tokio-telemetry/src/task_dumped.rs
```diff
@@ -258,3 +288,43 @@ impl Encodable for TaskDumpData<'_> {
         });
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::SplitMix64;
```
jlizen, line 287: good to have a test that verifies the counter  recovers correctly after a large deficit, eg

```
  #[test]
  fn redraw_loop_recovers_from_large_deficit() {
      let mut rng = SplitMix64::new(99);
      let mean_ns: u64 = 10_000_000; // 10ms

      // Simulate: counter starts at a small draw, then a huge idle
      // drives it deeply negative.
      let mut next_sample_ns = rng.draw_exponential_ns(mean_ns);
      let huge_idle: i64 = 5_000_000_000; // 5 seconds
      next_sample_ns -= huge_idle;
      assert!(next_sample_ns < 0);

      // Redraw loop (mirrors the poll path logic).
      let mut iterations = 0u32;
      while next_sample_ns <= 0 {
          next_sample_ns += rng.draw_exponential_ns(mean_ns);
          iterations += 1;
      }

      // Should recover in a bounded number of iterations (~500 for 5s/10ms).
      assert!(next_sample_ns > 0);
      assert!(
          iterations < 2000,
          "took {iterations} iterations to recover; expected ~500"
      );
  }
```

jlizen (approved):
Looks great. How'd it do in the bench?

rcoh:
there's basically no difference in benchmarking because the mutex is effectively uncontended unless you are also using backtrace to symbolize

dial9-tokio-telemetry/src/task_dumped.rs
rcoh: I ended up using slightly different logic to reset basically equivalent to saturating_sub on the next_sample_ns

PR #372, opened 2026-05-08
fix: use real waker in task dump capture to prevent lost wakes

jlizen:
I couldn't reproduce the issue, so I didn't validate the fix directly. However, it does seem strictly better than the old way for now.

@rcoh if you have a working reproduction, would welcome it, especially if it is reachable in unit tests


rcoh (approved):
It reproduced reliably when I enabled task dumps for the example app

PR #374, opened 2026-05-08
feat: write dial9 crate README

jlizen:
self-explanatory


rcoh (approved):

rcoh:
As part of doing this, we need to update that readme to explain that if you want to integrate dial9 you need to use `dial9-tokio-telemetry` so that folks aren't confused

jlizen:
@rcoh i rewrote the readme and stuck it in dial9, and then turned dial9-viewer into a pointer. Mind taking another look?

rcoh (approved):

PR #375, opened 2026-05-08
fix: eliminate task dump busy loop and move dl_iterate_phdr off hot path

rcoh:
## Summary

Two fixes for task dump overhead that was causing significant performance degradation:

### 1. Break the capture→wake→poll busy loop

`trace_with` re-polls the inner future with the real waker, which immediately fires and re-schedules the task. This caused an infinite loop: capture → wake → poll → capture → …

Fix: `just_captured` flag suppresses capture on the spurious wake, limiting overhead to one extra poll per capture point.

### 2. Two-phase stack walking (lock-free hot path)

`_Unwind_FindEnclosingFunction` calls `dl_iterate_phdr` which takes a process-wide mutex. Previously this was called per-frame on every capture (every yield point).

Now:
- **Capture** (every yield): only `_Unwind_GetIP` — no locks
- **Emit** (Poisson-sampled, rare): `_Unwind_FindEnclosingFunction` trims to root/leaf boundaries

### Results

| Metric | Before fix | After fix | Baseline (no dumps) |
|--------|-----------|-----------|---------------------|
| POST p99 | 5.5ms | **2.5ms** | 1.85ms |
| Throughput | 6,361 | **6,963** | 6,976 |
| Sched delays >1ms | 14,093 | **373** | 262 |
| Worker util | 11%/99.7% | **17%/16%** | 12%/12% |
| CPU: mutex_lock | 14.3% | **0%** | 0% |

### Also

- Adds `--no-task-dumps` flag to metrics-service for A/B testing


dial9-tokio-telemetry/src/task_dumped.rs
```diff
@@ -207,16 +222,42 @@ impl FrameBuf {
...
         shared.if_enabled(|buf| {
-            for i in 0..self.offsets.len() {
-                let start = self.offsets[i];
-                let end = self.offsets.get(i + 1).copied().unwrap_or(self.ips.len());
-                buf.record_encodable_event(&TaskDumpData {
-                    timestamp_ns: capture_ts,
-                    task_id,
-                    callchain: &self.ips[start..end],
-                });
+            // Each chain is stored as 3 entries in offsets:
```
jlizen: This is clever, though I wonder if it's worth the extra complexity for the cold path compared to just a named struct. I worry about it being tough to maintain if we ever need to eg add fields.

```
  struct ChainMeta {
      ip_start: usize,
      root_addr: usize,  // 0 = None
      leaf_addr: usize,
  }

  struct FrameBuf {
      ips: Vec<u64>,
      chains: SmallVec<[ChainMeta; 4]>,  // was: offsets: SmallVec<[usize; 8]>
  }
```

dial9-tokio-telemetry/src/unwind.rs
```diff
@@ -23,51 +34,55 @@ unsafe extern "C" {
...
+        frame_ips.iter().position(|&ip| {
+            let sym = unsafe { _Unwind_FindEnclosingFunction(ip as *mut c_void) };
+            std::ptr::eq(sym, root)
+        })
+    });
+
+    match (leaf_start, root_end) {
+        (Some(start), Some(end)) if start + 1 < end => &frame_ips[start + 1..end],
+        (Some(start), None) if start + 1 < frame_ips.len() => &frame_ips[start + 1..],
+        _ => frame_ips,
```
jlizen: i would expect to return an empty slice if we can't find boundaries, this seems like it would add noise?

dial9-tokio-telemetry/src/task_dumped.rs
```diff
@@ -207,16 +222,42 @@ impl FrameBuf {
...
+                let root_addr = chunk[1];
+                let leaf_addr = chunk[2];
+
+                // Determine the end of this chain's IPs: either the start of
+                // the next chain or the end of the ips buffer.
+                let ip_end = self
+                    .offsets
+                    .chunks_exact(3)
+                    .map(|c| c[0])
+                    .find(|&s| s > ip_start)
```
jlizen: nit: These are ordered, sowe should be able to instead just get `chunk[3]` instead of scanning from the beginning right? Anyway this is cold path, so if it's more readable this way, np.

dial9-tokio-telemetry/src/task_dumped.rs
```diff
@@ -207,16 +222,42 @@ impl FrameBuf {
...
+                    .unwrap_or(self.ips.len());
+
+                let raw = &self.ips[ip_start..ip_end];
+                let root = if root_addr == 0 {
```
jlizen: handle `leaf_address` similarly?

dial9-viewer/ui/demo-trace.bin
jlizen, line 1: why so much larger?

jlizen (approved):
Great refactor, looks good.

I agree with the choice to have the extra poll rather than risking lost wakes.

Looks like you have some red CI to muck with still.

dial9-viewer/ui/demo-trace.bin
rcoh: task dumps

PR #376, opened 2026-05-08
feat: add symposium keyword to dial9-tokio-telemetry

jlizen:
We're still figuring out the "discover crates with symposium skills that you don't already have" story, but probably v1 will involve crates.io keywords.

I imagine we want dial9 to pop up on the discovery list as Niko is getting eyes on Symposium, so adding our tag.

### DIFF
--- a/dial9-tokio-telemetry/Cargo.toml
+++ b/dial9-tokio-telemetry/Cargo.toml
@@ -7,7 +7,7 @@ repository.workspace = true
 publish = true
 readme = "README.md"
 description = "Low-overhead runtime telemetry for Tokio with poll timing, wake events, and CPU profiling"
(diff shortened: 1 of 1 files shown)


dial9-tokio-telemetry/Cargo.toml
```diff
@@ -7,7 +7,7 @@ repository.workspace = true
 publish = true
 readme = "README.md"
 description = "Low-overhead runtime telemetry for Tokio with poll timing, wake events, and CPU profiling"
-keywords = ["tokio", "telemetry", "profiling", "async", "tracing"]
```
jlizen, line 10: max of 5

rcoh (approved):

PR #377, opened 2026-05-08
feat: Taskdump viewer & expand inline frames in flamegraphs

rcoh:
I am pretty stoked about this it came out really nice.

when the idle time has a crosshatch, it means we have a task dump for that period
Screenshot 2026-05-08 at 2 07 29 PM

when you click it, we render it as a flamegraph (there can be multiple futures that you are waiting on)

Screenshot 2026-05-08 at 2 07 39 PM

You can also see a task-flamegraph for the entire task

### DIFF
--- a/dial9-viewer/skills/dial9-toolkit/scripts/analyze.js
+++ b/dial9-viewer/skills/dial9-toolkit/scripts/analyze.js
@@ -677,6 +677,7 @@ async function parseWorkerMain(traceFile, cachePath) {
     callframeSymbols: mapToEntries(trace.callframeSymbols),
     threadNames: mapToEntries(trace.threadNames),
     runtimeWorkers: mapToEntries(trace.runtimeWorkers),
+    taskDumps: mapToEntries(trace.taskDumps),
     clockSyncAnchors: trace.clockSyncAnchors, clockOffsetNs: trace.clockOffsetNs,
   }});
   for (const e of trace.events) writeLine({ t: 'e', d: e });
@@ -703,7 +704,7 @@ function loadCacheFile(cachePath) {
     const rec = JSON.parse(line);
     switch (rec.t) {
       case 'm': raw = rec.d;
-        for (const k of ['spawnLocations','taskSpawnLocs','taskSpawnTimes','taskTerminateTimes','callframeSymbols','threadNames','runtimeWorkers'])
+        for (const k of ['spawnLocations','taskSpawnLocs','taskSpawnTimes','taskTerminateTimes','callframeSymbols','threadNames','runtimeWorkers','taskDumps'])
           if (raw[k]) raw[k] = new Map(raw[k]);
         break;
       case 'e': events.push(rec.d); break;
--- a/dial9-viewer/ui/trace_analysis.js
+++ b/dial9-viewer/ui/trace_analysis.js
@@ -431,22 +431,26 @@
       node.count++;
       for (const addr of chain) {
         const entry = callframeSymbols.get(addr);
-        const resolved = Array.isArray(entry) ? entry[0] : entry;
-        const key = resolved ? resolved.symbol : addr || "??";
-        const formatted = formatFrame(addr, callframeSymbols);
-        if (!node.children.has(key)) {
-          node.children.set(key, {
-            name: formatted.text,
-            fullName: key,
-            location: resolved ? resolved.location : null,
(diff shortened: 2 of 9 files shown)


dial9-viewer/skills/dial9-toolkit/scripts/analyze.js
```diff
@@ -677,6 +677,7 @@ async function parseWorkerMain(traceFile, cachePath) {
     callframeSymbols: mapToEntries(trace.callframeSymbols),
     threadNames: mapToEntries(trace.threadNames),
     runtimeWorkers: mapToEntries(trace.runtimeWorkers),
+    taskDumps: mapToEntries(trace.taskDumps),
```
jlizen, line 680: Bit surprised not to see any changes to steering, i would have expected the schema to change at least?

Np if planned as a follow up.

dial9-viewer/ui/trace_analysis.js
```diff
@@ -431,22 +431,26 @@
...
-            name: formatted.text,
-            fullName: key,
-            location: resolved ? resolved.location : null,
-            docsUrl: formatted.docsUrl,
-            children: new Map(),
-            count: 0,
-            self: 0,
-          });
+        // Expand inlined frames: an array entry means multiple frames at one address
+        const frames = Array.isArray(entry) ? entry : [entry];
```
jlizen, line 441: Is it just me or is this reversing the ordering of the flamegraphs, to put the outermost deepest? That seems unintuitive if so.

dial9-viewer/ui/trace_analysis.js
```diff
@@ -431,22 +431,26 @@
...
+        for (let fi = frames.length - 1; fi >= 0; fi--) {
+          const resolved = frames[fi];
+          const key = resolved ? resolved.symbol : addr || "??";
+          const formatted = resolved ? formatFrame(resolved) : formatFrame(addr, callframeSymbols);
```
jlizen, line 446: This seems more fragile than the old way since frames might be `[null]`.

We could do something like:
```
Array.isArray(entry) ? entry[0] : entry.
```

dial9-viewer/ui/viewer.html
```diff
@@ -4356,6 +4412,27 @@ <h3>⌨ Keyboard</h3>
...
+                    if (mx >= r.x1 && mx <= r.x2 && my >= r.y1 && my <= r.y2) {
+                        selectedTaskId = r.wakerTaskId;
+                        renderAll();
+                        return;
+                    }
+                }
+                // Check if clicking an idle region with task dumps
+                for (const r of taskDetailHitRegions) {
+                    if (mx >= r.x1 && mx <= r.x2 && my >= r.y1 && my <= r.y2 && r.taskDumps) {
+                        showTaskDumpStack(r.taskDumps);
```
jlizen: probably should be in a try/catch similar to `showIdleTimeFlamegraph`?

dial9-viewer/ui/viewer.html
```diff
@@ -4412,6 +4489,114 @@ <h3>⌨ Keyboard</h3>
             const fgContainer = document.getElementById("fg-container");
             const fgInstance = FlamegraphRenderer.createFlamegraph(fgContainer);
 
+            function showTaskDumpStack(dumps) {
```
jlizen: there is a lot of redundant sidebar code between `showTaskDumpStack` and `showIdleTimeFlamegraph`, should we consolidate?

Also consider extracting the dump UI logic into a separate module.

dial9-viewer/ui/viewer.html
```diff
@@ -2796,9 +2816,39 @@ <h3>⌨ Keyboard</h3>
...
+                        gapDumps.push(dumps[di]);
+                    }
+                    const hasDump = gapDumps.length > 0;
+
+                    ctx.fillStyle = hasDump ? "#2a2a5a" : "#2a2a4a";
                     ctx.fillRect(x1, bandTop, w, bandH);
-                    ctx.strokeStyle = "#444";
+
+                    // Cross-hatch pattern for idle periods with task dumps
+                    if (hasDump) {
```
jlizen, line 2856: feels like we'll want this in a helper sooner or later

jlizen (approved):
New interface looks great, just some small things.

Consider adding unit tests for `TaskDumpEvent` parsing and the frame expansion logic.

PR #382, opened 2026-05-08
 fix(test): stitch together all demo segments

jlizen:
Fix for:
```
taskDumps:
✓ trace.taskDumps is a Map with 84 task IDs
✓ All taskDumps arrays are sorted by timestamp
✓ TaskDumps have expected {timestamp, callchain} shape with hex-string addresses
✗ task 13 has taskDumps but is not in taskSpawnTimes
Error: Process completed with exit code 1.
0s
```
https://github.com/dial9-rs/dial9-tokio-telemetry/actions/runs/25584069384/job/75109003829

Looks like there are some race conditions where a task Id might be from a "pre-trace" task (due to i/o driver, etc). Allowing a small ratio of missing spawns.


dial9-viewer/ui/test_trace_analysis.js
```diff
@@ -467,13 +467,17 @@ async function main() {
   }
 
   function testTaskDumpsTaskIdsKnown() {
-    // Every task ID that has a dump should be a known spawned task (no orphans).
+    // Most task IDs with dumps should be known spawned tasks.
+    // Some may be pre-existing tasks spawned before tracing started.
```
rcoh: Hmmm this seems like a bug. in the demo trace everything should be instrumented

rcoh (approved):

dial9-viewer/ui/test_trace_analysis.js
jlizen: I'll timebox chasing it a bit

jlizen:
Ok the real issue is that we rotate based on wall-clock time, which occasionally will force early rotation, and split a task across multiple traces.

For now, I"m fixing our test to concatenate all segments before analyzing. I'll also cut a backlog to re-examine the wall clock timing choice

jlizen:
Reproduced the failure running in docker locally when we split across wallclock boundaries. The fix stitches things together, and I had a successful run across a similar wallclock boundary.

PR #385, opened 2026-05-09
fix: rate_limit log when drain is failing

rcoh:
when `drained` returns an io error, we would call it in a hot-ish (5ms delay) loop and create a lot of logspam.

There is a more nuanced consideration to how we should handle when the drain command is erroring that probably depends on the root cause of the individual error but this fixes the logspam.


jlizen (approved):
Thanks for adding the new test infra!

Issue #279, opened 2026-04-23
full in memory pipeline 


rcoh:
For users who have plenty of ram but want to avoid disk writes, we should allow the writer to push data into memory instead of writing to disk, then hand the data directly to the background queue instead of going via a file.

If we really pushed this direction, we could even incrementally do the background work on the file (e.g. compress it as segments come in) in an attempt for avoid the few hundred milliseconds of background work that can disrupt other work on the host.

jv1i:
completed in #435, only remaining follow up is support for it on the dial9 macro (in progress)

PR #389, opened 2026-05-11
docs(design): in-memory pipeline

jv1i:
#279

Adds a design document for the implementation of a full in-memory pipeline that does not rely on writing to disk. 
The last section (# 7 open questions) includes some of the items that I don't have any strong preference on.


dial9-tokio-telemetry/design/in-memory-pipeline.md
```diff
@@ -0,0 +1,426 @@
...
+
+**Invariant: exactly one consumer.** `Notify` wakes one waiter per `notify_one`; a second worker would race on `notified()` and silently steal half the wakeups. If parallel processing is ever wanted (e.g. multiple S3 uploads in flight), it lives **inside** the sink (one worker, one sink, sink fans out internally), not as multiple workers reading the same queue.
+
+### Channel-driven worker
+
+The worker's existing poll-the-directory loop becomes one variant of a
+`SegmentSource`:
+
+```rust
+pub(crate) enum SegmentSource {
```
rcoh: I think my instinct here is that we extract an Fs abstraction instead of a Writer abstraction 

I would like to get rid of the TraceWriter trait. 

This design will simultaneously support in memory and make it easier to test RotatingWriter

dial9-tokio-telemetry/design/in-memory-pipeline.md
```diff
@@ -66,71 +60,84 @@ record_event() ──┐
...
+Object-safe so `Arc<dyn Fs>` works.
 
 ```rust
-trait Fs: Send + Sync + 'static {
-    /// Open a fresh active segment for writing. The writer wraps the
-    /// returned handle in a `BufWriter`. Subsequent rotations seal it
-    /// via [`Fs::seal`].
+pub trait Fs: Send + Sync + 'static {
+    // --- Write side ---
     fn create(&self, path: &Path) -> io::Result<Box<dyn Write + Send>>;
```
rcoh: probably fine but we should benchmark. I would not be surprised if you get significant wins from being able to optimize on the specific type in use.

dial9-tokio-telemetry/design/in-memory-pipeline.md
```diff
@@ -66,71 +60,84 @@ record_event() ──┐
...
+    /// Wait for new segments to potentially be available. Disk: sleeps
+    /// `poll_interval`. Memory: awaits the channel `Notify`.
+    fn wait_for_more<'a>(&'a self, stop: &'a CancellationToken, poll_interval: Duration) -> Pin<Box<dyn Future<Output = ()> + Send + 'a>>;
+    /// Writer signaled it will produce no more segments. Memory: channel flag. Disk: always false (stop token handles shutdown).
+    fn writer_done(&self) -> bool { false }
+    /// Writer finalize hook. Memory: sets writer_done + notify.
+    fn mark_writer_done(&self) {}
+    /// `RotatingWriter::evict_oldest` hook. Disk: bumps the
+    /// dropped-segments counter surfaced via `TakenFiles`.
+    fn record_writer_eviction(&self) {}
```
rcoh: seems like this could potentially be connected w/ remove_sealed?

rcoh (approved):
seems pretty much good, worth a prototype

dial9-tokio-telemetry/design/in-memory-pipeline.md
jv1i: yup, done

PR #392, opened 2026-05-11
feat: instrumented joinsets and other custom spawns

### LINKED ISSUE #301: make Traced future public

noxware:
## Current solution

Exposes a `handle.spawn_with(fut, |f: TracedFuture| { ... })` to allow spawning instrumented/traced futures through custom spawn APIs (like `JoinSet::spawn`).

## Example

```rust
let mut set: JoinSet<()> = JoinSet::new();
handle.spawn_with(fut, |f| set.spawn(f));
```

## For reviewers

Most important changes are at `recorder/mod.rs` and `traced.rs`.

## Other
Closes #301

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
+++ b/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
@@ -33,9 +33,9 @@ crate::primitives::thread_local! {
     /// cleared in `on_thread_stop`. Enables [`TelemetryHandle::current`].
     static CURRENT_HANDLE: RefCell<Option<TelemetryHandle>> = const { RefCell::new(None) };
 
-    /// Set by `TelemetryHandle::spawn()` before calling `tokio::spawn()`,
-    /// so the `on_task_spawn` hook can distinguish instrumented from raw spawns.
-    static INSTRUMENTED_SPAWN: Cell<bool> = const { Cell::new(false) };
+    /// Nest count for [`InstrumentedSpawnGuard`]. `on_task_spawn` treats
+    /// any value `> 0` as an instrumented spawn.
+    static INSTRUMENTED_SPAWN: Cell<u32> = const { Cell::new(0) };
 }
 
 // ---------------------------------------------------------------------------
@@ -204,7 +204,7 @@ fn register_hooks(
             s5.if_enabled(|buf| {
                 let task_id = TaskId::from(meta.id());
                 let location = meta.spawned_at();
-                let instrumented = INSTRUMENTED_SPAWN.with(|f| f.get());
+                let instrumented = INSTRUMENTED_SPAWN.with(|f| f.get()) > 0;
                 let timestamp_ns = crate::telemetry::events::clock_monotonic_ns();
                 buf.record_encodable_event(&runtime_context::TaskSpawn {
                     timestamp_ns,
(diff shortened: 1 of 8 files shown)


dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -506,42 +506,73 @@ impl TelemetryHandle {
...
+    /// # async fn work() {}
+    /// # fn demo(runtime: &Runtime) {
+    /// let handle = TelemetryHandle::current();
+    /// let mut set: JoinSet<()> = JoinSet::new();
+    /// handle.spawn_with(work(), |f| set.spawn_on(f, runtime.handle()));
+    /// # }
+    /// ```
+    ///
+    /// [`TracedFuture<F>`]: crate::telemetry::TracedFuture
+    pub fn spawn_with<F, S>(
```
noxware, line 559: This is the new API, the root part of the PR.

dial9-tokio-telemetry/src/traced.rs
```diff
@@ -21,30 +23,78 @@ impl std::fmt::Debug for TracedHandle {
...
+            #[pin]
+            inner: InstrumentedFuture<F>,
+        },
+        Empty,
+    }
+}
+
+pin_project! {
+    /// Future wrapper that captures wake events for a known Tokio task.
+    pub(crate) struct WakeTracked<F> {
```
noxware: This is what used to be `Traced` (kinda).

dial9-tokio-telemetry/src/traced.rs
```diff
@@ -21,30 +23,78 @@ impl std::fmt::Debug for TracedHandle {
...
 pin_project! {
-    /// Future wrapper that captures wake events (and later, task dumps).
-    pub struct Traced<F> {
+    /// Future wrapper produced by `spawn_with` for custom spawn APIs.
+    ///
+    /// On first poll, `TracedFuture<F>` resolves the surrounding Tokio task ID
+    /// and uses it for task instrumentation. If the future is polled outside a
+    /// Tokio task context, it runs as a transparent passthrough without wake
+    /// tracking or task dumps.
+    pub struct TracedFuture<F> {
```
noxware, line 41: This is the concreate type of the `spawn_with` closure. It composes `WakeTracked` and `TaskDumped` (if enabled), without dynamic dispatching.

Note: It can also be called `Traced<F>` as the name is now free. Although we also have other explicit "traced" types like `TracedRuntime`. Opinons?

dial9-tokio-telemetry/src/traced.rs
```diff
@@ -21,30 +23,78 @@ impl std::fmt::Debug for TracedHandle {
...
+impl<F> TracedFuture<F> {
+    pub(crate) fn new(inner: F, handle: Option<TracedHandle>) -> Self {
+        Self {
+            state: TracedFutureState::Init { inner, handle },
+        }
+    }
+}
+
+pin_project! {
+    #[project = TracedFutureStateProj]
```
noxware, line 62: An explicit name is needed to pattern match the enum.

dial9-tokio-telemetry/src/traced.rs
```diff
@@ -21,30 +23,78 @@ impl std::fmt::Debug for TracedHandle {
...
+        },
+        Passthrough {
+            #[pin]
+            inner: F,
+        },
+        Instrumented {
+            #[pin]
+            inner: InstrumentedFuture<F>,
+        },
+        Empty,
```
noxware, line 77: `Empty` is just used to transit the state safely.

dial9-tokio-telemetry/src/traced.rs
noxware, line 1: This file contains the auxiliar changes to make `spawn_with` work by exposing the `TracedFuture` type. It has the most important changes.

dial9-tokio-telemetry/src/traced.rs
```diff
@@ -21,30 +23,78 @@ impl std::fmt::Debug for TracedHandle {
...
 
+#[cfg(feature = "taskdump")]
+type MaybeTaskDumped<F> = crate::task_dumped::TaskDumped<F>;
+
+#[cfg(not(feature = "taskdump"))]
+type MaybeTaskDumped<F> = F;
+
+type InstrumentedFuture<F> = WakeTracked<MaybeTaskDumped<F>>;
+
 pin_project! {
```
noxware, line 34: We rely on pin projection to avoid writting unsafe code.

noxware (commented):
Added some comments to help with the code review :)

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -506,42 +506,73 @@ impl TelemetryHandle {
...
-    F: std::future::Future,
-{
-    future
+    /// Spawn an instrumented future through a user-supplied spawn function.
+    ///
+    /// `spawn_fn` is invoked synchronously with a [`TracedFuture<F>`] that
+    /// the caller is expected to spawn immediately. The closure's return
+    /// value is forwarded back to the caller, so you can keep the
+    /// [`tokio::task::JoinHandle`], [`tokio::task::AbortHandle`], or
+    /// whatever the spawn function returns.
```
jv1i: Should we spell out more explicitly the need to immediately spawn, and what "immediately" means?
something like
> spawn_fn must perform a real tokio::spawn (or equivalent) synchronously on this thread. Deferring the spawn or running the future via block_on will misattribute wake events and not properly flag as instrumented.

(iirc users have been confused with block_on usage)

dial9-tokio-telemetry/tests/spawn_with.rs
```diff
@@ -0,0 +1,202 @@
...
+    drop(runtime);
+    drop(guard);
+}
+
+/// `RuntimeTelemetryHandle::spawn_with` composes with `JoinSet::spawn_on`
+/// to target a specific runtime even when called from outside any
+/// runtime context, while still emitting wake events and marking the
+/// `TaskSpawn` instrumented.
+#[test]
+fn runtime_handle_spawn_with_targets_correct_runtime() {
```
jv1i, line 161: doc not fully accurate, test not currently asserting wether the task spawn was marked as instrumented, or if emits wake events, right? I think you can achieve that with `CapturingWriter`

dial9-tokio-telemetry/src/traced.rs
```diff
@@ -21,30 +23,78 @@ impl std::fmt::Debug for TracedHandle {
...
         #[pin]
         inner: F,
-        handle: TracedHandle,
-        task_id: TaskId,
         waker_data: Arc<TracedWakerData>, // reused across polls to avoid a per-poll Arc allocation
     }
 }
 
-impl<F> Traced<F> {
+impl<F> WakeTracked<F> {
```
jv1i: bikeshed: with everything else being called Something**Traced** (wether they're semantically tracing or tracking) I'd personally name this WakeTraced, also so it matches `TracedWakerData` and some of the new comments/docs. (not user-facing so not super important)

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -606,18 +637,60 @@ impl RuntimeTelemetryHandle {
...
     }
+
+    /// Spawn an instrumented future through a user-supplied spawn function.
+    ///
+    /// Mirrors [`TelemetryHandle::spawn_with`] but does not require an
+    /// ambient tokio runtime context — `spawn_fn` is invoked on the
+    /// calling thread, so it is the closure's responsibility to target
+    /// this handle's runtime (typically via
+    /// [`tokio::task::JoinSet::spawn_on`] passing the appropriate
+    /// [`tokio::runtime::Handle`]).
```
jv1i: I think language here might be a bit misleading, it implies `TelemetryHandle::spawn_with` needs a tokio runtime context and this one doesn't. But both `spawn_with` are the same in that regard right? the difference is what handle type you have at hand to call it on

yulnr (approved):
Looks good to me! happy we could keep the instrumented flag an internal concern.
Leaving some minor comments

dial9-tokio-telemetry/tests/spawn_with.rs
noxware: Missed this, thanks!

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
noxware: Yeah, it's a bit misleading. There is no need for a global tokio runtime context, but the user's custom spawn executed inside the closure must be hooked by dial9, so the TL is read in `on_task_spawn`. `spawn_with` will do nothing for non-hooked tokio runtimes.

I changed this, check if it looks better now.

noxware:
@yulnr It should be done now :)

rcoh (approved):
Nice!

Issue #393, opened 2026-05-11
Android Support


nickrobinson:
I noticed a few places throughout the codebase where `#[cfg(target_os = "linux")]` was present to gate some of the profiling behavior. There may be some limitations on Android when it comes to capturing CPU cycles, but it seems like SchedStat support should work. I was curious if leaving Android off was an intentional choice here?

I'm happy to make and test this change, but wanted to make sure I wasn't missing any additional info first.

rcoh:
definitely not intentional, only my own lack of knowledge of the android ecosystem. Happy to merge any changes related to this. And if you know any decent way to get CI to eventually include these targets, let me know

nickrobinson:
I spent some time looking at CPU profiling as well. I think that is going to be much more involved. There are a few issues at the moment:

1. There are a few system calls that are being made that have not been added to android in libc (timer_create/timer_delete/timer_settime). There either needs to be an upstream change, or some conditional logic to make these sys calls directly on Android
2. ```perf_event_open``` is pretty locked down on android. It is gated based on the status of ```perf_event_paranoid```. Some checking would need to be done up front and a warning provided back to users if it is inaccesible

rcoh:
I think disabling the ctimer fallback on android is reasonable. From my extremely short research it seems like an Android specific `simpleperf` backend is probably what we would need?

nickrobinson:
FYI: I've put up a [PR to libc](https://github.com/rust-lang/libc/pull/5108) to add the missing android libc functions that the cpu profiling feature is using. I've played around with this locally with a Pixel 6a device and it seems like once these changes are in place in libc all that is needed to get CPU Profiling working is to update the target_os macros like what was done on #395

nickrobinson:
PR to add CPU Profiling Support: https://github.com/dial9-rs/dial9/pull/685

PR #395, opened 2026-05-12
feat: add android schedstat support

nickrobinson:
This PR updates the ```target_os``` flags to enable schedstat support for android. Android has the same ```/proc/self/task/<tid>/schedstat``` file descriptor as typical linux systems. The running process is able to access it without running into permissions issues.

Here is a screenshot of a trace pulled from a Pixel 7 emulator running locally:

Screenshot 2026-05-12 at 5 49 13 AM

Related: #393


rcoh (approved):
Thanks!

rcoh:
Looks like a couple of CI failures

PR #396, opened 2026-05-12
add: expose public Unwinder::capture API

rcoh:
This series of changes exposes a public framepointer unwinding API which will be used from #362

### DIFF
--- a/perf-self-profile/src/sys/linux/fp_profiler/unwind.rs
+++ b/perf-self-profile/src/sys/linux/fp_profiler/unwind.rs
@@ -144,7 +144,10 @@ pub unsafe fn unwind(pc: usize, mut fp: usize, sp: usize, out: &mut [u64]) -> Ca
 ///
 /// # Safety
 /// `ucontext` must be the pointer the kernel passed to a SA_SIGINFO handler.
-pub unsafe fn unwind_from_ucontext(ucontext: *mut libc::c_void, out: &mut [u64]) -> CaptureResult {
+pub(crate) unsafe fn unwind_from_ucontext(
+    ucontext: *mut libc::c_void,
+    out: &mut [u64],
+) -> CaptureResult {
     let (pc, fp, sp) = unsafe { read_pc_fp_sp(ucontext) };
     unsafe { unwind(pc, fp, sp, out) }
 }
--- a/perf-self-profile/src/sys/linux/fp_profiler/unwind.rs
+++ b/perf-self-profile/src/sys/linux/fp_profiler/unwind.rs
@@ -49,14 +49,21 @@ fn strip_pac(addr: usize) -> usize {
 /// Walk the frame-pointer chain starting from the given (pc, fp, sp) triple,
 /// usually obtained from a signal handler's ucontext.
 ///
+/// Returns `(frames_written, truncated)`. `truncated` is `true` if the walk
+/// stopped because the output buffer (or [`MAX_FRAMES`]) was full *and* at
+/// least one additional frame would have been valid. A natural stop (end of
+/// chain, faulty load, implausible pointer) produces `truncated = false`.
+///
 /// # Safety
 /// - `install_handler` must have been called.
 /// - Should generally be called from a signal handler where the target thread
 ///   is stopped; walking a running thread's stack races with mutations.
-pub unsafe fn unwind(pc: usize, mut fp: usize, sp: usize, out: &mut [u64]) -> usize {
+pub unsafe fn unwind(pc: usize, mut fp: usize, sp: usize, out: &mut [u64]) -> (usize, bool) {
     let limit = out.len().min(MAX_FRAMES);
     if limit == 0 {
-        return 0;
+        // No room even for the interrupted PC. The walk would have produced
+        // at least one frame, so this is truncation.
+        return (0, true);
     }
 
     out[0] = pc as u64;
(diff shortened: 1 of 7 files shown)


perf-self-profile/src/sys/linux/fp_profiler/unwind.rs
```diff
@@ -65,47 +72,54 @@ pub unsafe fn unwind(pc: usize, mut fp: usize, sp: usize, out: &mut [u64]) -> us
...
 }
 
 /// Unwind from inside a signal handler given the raw ucontext.
 ///
+/// Returns `(frames_written, truncated)`; see [`unwind`] for details.
+///
 /// # Safety
 /// `ucontext` must be the pointer the kernel passed to a SA_SIGINFO handler.
-pub unsafe fn unwind_from_ucontext(ucontext: *mut libc::c_void, out: &mut [u64]) -> usize {
+pub unsafe fn unwind_from_ucontext(ucontext: *mut libc::c_void, out: &mut [u64]) -> (usize, bool) {
```
Fluzko: complete NIT, but a bit hard to remember what a `(usize, bool)` means without reading docs, might be worth creating a type?

Fluzko (approved):

perf-self-profile/src/unwinder.rs
```diff
@@ -0,0 +1,531 @@
+//! Framepointer-based stack unwinding
+//!
+//! This implementation uses a SIGSEGV fault handler to allow safe walking of stacks. Without
+//! this, there is no way to perform framepointer unwinding without risking segfaults when walking
+//! stacks where framepointers are not enabled.
+//!
+//! Because of this, the unwinder must be "[`install`ed](Unwinder::install)" before
+//! you can use it to [`capture`](Unwinder::capture) a stack.
+//!
+//! The unwinded stacks are only addresses. You must use a symbolizer separately to
```
jlizen: unwound

perf-self-profile/src/unwinder.rs
```diff
@@ -0,0 +1,531 @@
...
+    /// Called from [`Unwinder::capture`] (which is `#[inline(never)]`), so the
+    /// "current frame" observed here is `Unwinder::capture`'s frame. This
+    /// function is `#[inline(always)]` specifically so it does *not* introduce
+    /// another frame — see the `frame_zero_points_into_caller_of_capture`
+    /// test.
+    ///
+    /// # Safety
+    /// Same obligations as [`Unwinder::capture`]: handler installed,
+    /// not inside a SIGSEGV handler, frame pointers enabled.
+    #[inline(always)]
```
jlizen, line 150: This (and read_caller_regs()) are a bit fragile since it is only a (strong) hint to inline it...
We could perhaps do something more resilient with a capture!() macro used by the caller that does an asm!() read? Though ergonomics would be pretty gross,  likely a bad idea.

I see we do have a runtime test with frame_zero_points_into_caller_of_capture which helps for our own usage, though it doesn't extend to callers that might have totally different optimization settings, LTO, smaller codegen units, etc. I wonder if there is any sort of smoke test that we can offer them too. I guess the problem then is, whatever test is run, is still being run in debug mode with different compilation settings?

This is probably pedantics since these functions are so small that rustc is unlikely to skip inlining, probably ignore me.

perf-self-profile/src/sys/linux/fp_profiler/unwind.rs
```diff
@@ -36,27 +36,34 @@ const MAX_FRAME_SIZE: usize = 0x40000;
...
+/// stopped because the output buffer (or [`MAX_FRAMES`]) was full *and* at
+/// least one additional frame would have been valid. A natural stop (end of
+/// chain, faulty load, implausible pointer) produces `truncated = false`.
+///
 /// # Safety
 /// - `install_handler` must have been called.
 /// - Should generally be called from a signal handler where the target thread
 ///   is stopped; walking a running thread's stack races with mutations.
-pub unsafe fn unwind(pc: usize, mut fp: usize, sp: usize, out: &mut [u64]) -> usize {
+pub unsafe fn unwind(pc: usize, mut fp: usize, sp: usize, out: &mut [u64]) -> (usize, bool) {
```
jlizen: I'd suggest a named struct return here as well

perf-self-profile/src/sys/linux/fp_profiler/unwind.rs
jlizen: related nit, can we call this pub(crate) if it is not re-exported? It's a bit hard to tell currently which is exported, I see that `capture()` is currently.

perf-self-profile/src/unwinder.rs
```diff
@@ -0,0 +1,531 @@
...
+    /// Another library or the runtime may install its own SIGSEGV handler
+    /// after [`install`](Self::install) is called. If that handler does not
+    /// chain to ours, [`capture`](Self::capture) may crash on a bad frame
+    /// pointer instead of aborting the walk safely. Callers who need
+    /// defence against this can call `verify_handler` periodically or
+    /// before safety-critical captures.
+    ///
+    /// Performs one `sigaction` syscall. Not suitable for per-sample hot
+    /// paths.
+    pub fn verify_handler(&self) -> bool {
```
jlizen, line 68: might be worth a comment that this requires &self to ensure that `Unwinder::install()` has been called, even though it doesn't require any &self access internally

jlizen (approved):

perf-self-profile/src/unwinder.rs
rcoh: I think this is probably not worth it. The cost of it being not inlined is just an extra stack frame so I'm not worried

PR #397, opened 2026-05-13
refactor(dial9-viewer): move skills to .symposium/skills

jlizen:
Moving from `dial9-viewer/skills/**` to `dial9-viewer/.symposium/skills/**` to align with [the new symposium default](https://github.com/symposium-dev/symposium/pull/215).

If we'd rather, we can keep the old location and instead have a specific non-default passed through in https://github.com/symposium-dev/recommendations .

Let me know if you prefer that.


rcoh:
I don't have a strong opinion but this seems like a really strong coupling to symposium

jlizen:
Fair, i'll just change it in the recommendations repo, good point.

PR #399, opened 2026-05-13
fix: recover from missing .active file during rotation (#386)

### LINKED ISSUE #386: failure handling: `drained` failure
right now, if `drained()` on `TraceWriter` fails, the flush loop will call it basically infinitely. If this failure is due to some sort of state violation then it may be permanently stuck. We should consider what failure modes can cause this and how dial9 should respond.

The one failure mode I have seen is if the file is deleted from disk but there are likely others.

rcoh:
If the active trace file or its parent directory is removed externally (operator, log rotation, container teardown), `RotatingWriter::rotate()` fails at `fs::rename(.active, .bin)` or `File::create(new_path)`. Before this fix the error propagated out of `drained()` without advancing `next_drain_time` / `next_rotation_time`, so `should_drain()` kept firing on every 5ms flush-loop tick — busy loop.

Recovery:
- Advance both timers up front in rotate() so any failure below cannot cause busy-spinning.
- Tolerate NotFound from the rename: log a rate-limited warning, skip pushing to closed_files, and start a fresh segment.
- Tolerate NotFound from File::create by retrying create_dir_all once.
- On any unrecoverable error, mark the writer Finished so it stops cleanly instead of retrying every cycle.
- Apply the same NotFound tolerance to finalize().

Fixes #386

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/writer.rs
+++ b/dial9-tokio-telemetry/src/telemetry/writer.rs
@@ -432,24 +432,83 @@ impl RotatingWriter {
         let WriterState::Active { writer: raw, .. } = &mut self.state else {
             return Ok(());
         };
-        raw.flush()?;
-        // Seal the current segment: snapshot size and rename .active → .bin
-        let closed_size = raw.bytes_written();
-        let sealed = Self::file_path(&self.base_path, self.next_index - 1);
-        fs::rename(&self.active_path, &sealed)?;
-        self.closed_files.push_back((sealed, closed_size));
 
-        let new_path = Self::active_path(&self.base_path, self.next_index);
-        self.next_index += 1;
-        let file = File::create(&new_path)?;
-        let writer = BufWriter::new(file);
-        self.state = Self::prepare_segment(writer)?;
-        self.active_path = new_path;
-        self.did_rotate = true;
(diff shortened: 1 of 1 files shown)


dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -432,24 +432,83 @@ impl RotatingWriter {
...
+                }
+            }
+            Err(e) => {
+                self.state = WriterState::Finished;
+                return Err(e);
+            }
+        };
         let writer = BufWriter::new(file);
-        self.state = Self::prepare_segment(writer)?;
+        self.state = match Self::prepare_segment(writer) {
```
Fluzko, line 502: with a fail on this we would be leaving an orphan file created [@479](https://github.com/dial9-rs/dial9/pull/399/changes#diff-d8c06bb8e8616ef91eb3f9f3c5fe34150ec700174d01c602da55415cdb4c38f3R479) and an inconsistent state. No big deal, but probably worth to take into account

jlizen: Agreed, it would be good to remove the file inside our error handling block.

jlizen (approved):

PR #401, opened 2026-05-13
fix: bring back old API on core telemetry builder

rcoh:
A previous PR removed `.s3_config` as an option on TelemetryCoreBuilder, this brings it back

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
+++ b/dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
@@ -1516,8 +1516,26 @@ pub struct TelemetryCore;
 impl TelemetryCore {
     /// Build a telemetry session. Recording starts disabled; call
     /// [`TelemetryGuard::enable`] to begin recording.
-    #[builder]
+    #[builder(state_mod = telemetry_core_builder)]
     pub fn new(
+        /// The pipeline of [`SegmentProcessor`](crate::background_task::SegmentProcessor)s
+        /// to run on each sealed segment. When empty the background worker
+        /// is not spawned.
+        #[builder(field)]
+        processors: Vec<Box<dyn crate::background_task::SegmentProcessor>>,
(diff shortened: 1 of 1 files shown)


dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -1659,6 +1694,38 @@ impl TelemetryCore {
     }
 }
 
+// Custom methods on the generated builder.
+impl<W: TraceWriter, S: telemetry_core_builder::State> TelemetryCoreBuilder<W, S> {
```
quinnwerks, line 1698: Thank you!

quinnwerks (approved):
Just out of curiosity: how do you think we can avoid this?

skanjalkar (approved):

rcoh:
I am doing two things:
- improving the agentic review guidance to do a better job looking for breaking changes like this. I suspect a breaking changes specific pass would have caught it. This breakage was pretty obvious but it was in a complex PR and it got missed 
- a weekly job that will use code search to ensure we have tests covering all existing usage patterns 

I did the second one which caught this miss. I'll cut a PR for the usage patterns as well

jlizen (approved):

jlizen:
Shouldn't cargo-semver-checks have caught this too @rcoh ?

rcoh:
I don't think we are running it on PRs and running it on releases I think
is reliant on the maintainers machine? We should set that up too.

I suspect the quantity of generics may have it a path that CSC doesn't
cover though.

jlizen:
Yeah even if we are only running it on releases, i would have expected it to flag this as a breaking change and force a breaking version bump. Maybe it got swept up by a release that was already breaking?

It would be good to run it as a non-blocking step on PRs too though. It isn't reliant on the maintainers machine, no, it just builds docs in CI and analyzes them.

Anyway certainly possible that CSC also got confused.

Issue #351, opened 2026-05-04
using dial9 as only a profiler


rcoh:
dial9 is actually a useful tool even if you only use it to get profiling data off the host (without even pulling Tokio events).

This would need a new public API potentially, then in the viewer if you load a trace with only profiling info it should jump to a flamegraph

PR #406, opened 2026-05-15
feat: config from env

### LINKED ISSUE #302: add from_env that configures dial9 from a standard set of environment variables
This can allow for standardized CDK and zero-configuration normal applications.

A set of variables and defaults 

```
DIAL9_ENABLED=false
DIAL9_ENABLED_FILE=/tmp/enable-dial9 # start disabled and turn on dial9 when file is present 
DIAL9_ROTATION_SECS=15
DIAL9_S3_BUCKET=
DIAL9_CPU_PROFILE_ENABLED=true
DIAL9_CPU_SAMPLE_HZ=99
DIAL9_SCHEDULE_PROFILE_ENABLED=true
DIAL9_TRACE_DIR=/tmp/dial9-traces
DIAL9_MAX_DISK_USAGE_MB=1024

```

When configuration is invalid, debug builds panic and release builds fallback to disabled

noxware:
## Description

This PR gathers inspiration from the proposal in its [issue](https://github.com/dial9-rs/dial9-tokio-telemetry/issues/302), from the `production_use.rs` example, from [this](https://github.com/rustfs/rustfs/blob/3ced40f221aa007372a4b9c30fe295195912b6e2/crates/config/src/constants/runtime.rs#L31) real world usage and from the programmatic builder itself, to define a set of standard env vars and a function that reads them to launch dial9 with zero code.

## Out of scope

- `DIAL9_ENABLED_FILE` requires a special setup for the watcher.
- Mixing config from env with programmatic builder settings.

## Closes

#302


rcoh (commented):
I think the one piece I'd like to cover here from day-0 is a way to have profiling on but Tokio events off. (#351 )

the other features can probably be saved for later

dial9-tokio-telemetry/src/config.rs
```diff
@@ -190,12 +190,479 @@ type RuntimeConfigurator = Box<
...
+const ENV_DIAL9_S3_BUCKET: &str = "DIAL9_S3_BUCKET";
+const ENV_DIAL9_SERVICE_NAME: &str = "DIAL9_SERVICE_NAME";
+const ENV_DIAL9_S3_PREFIX: &str = "DIAL9_S3_PREFIX";
+const ENV_DIAL9_CPU_PROFILE_ENABLED: &str = "DIAL9_CPU_PROFILE_ENABLED";
+const ENV_DIAL9_CPU_SAMPLE_HZ: &str = "DIAL9_CPU_SAMPLE_HZ";
+const ENV_DIAL9_SCHEDULE_PROFILE_ENABLED: &str = "DIAL9_SCHEDULE_PROFILE_ENABLED";
+const ENV_DIAL9_TASK_DUMP_ENABLED: &str = "DIAL9_TASK_DUMP_ENABLED";
+const ENV_DIAL9_TASK_DUMP_IDLE_THRESHOLD_MS: &str = "DIAL9_TASK_DUMP_IDLE_THRESHOLD_MS";
+
+const DEFAULT_ENABLED: bool = false;
```
noxware, line 209: Note: When configuring the library manually, from Rust code, we default to enabled, but, by looking at the issue and `production_use.rs` I decided to default to disabled when using `from_env`.

dial9-tokio-telemetry/src/config.rs
```diff
@@ -190,12 +190,479 @@ type RuntimeConfigurator = Box<
...
+const DEFAULT_TRACE_DIR: &str = "/tmp/dial9-traces";
+const DEFAULT_MAX_DISK_USAGE_MB: u64 = 1024;
+const DEFAULT_TASK_TRACKING_ENABLED: bool = true;
+const DEFAULT_CPU_PROFILE_ENABLED: bool = cfg!(all(target_os = "linux", feature = "cpu-profiling"));
+const DEFAULT_SCHEDULE_PROFILE_ENABLED: bool =
+    cfg!(all(target_os = "linux", feature = "cpu-profiling"));
+const DEFAULT_TASK_DUMP_ENABLED: bool = false;
```
noxware, line 217: Defaults to `false` as it seems to be expensive.

dial9-tokio-telemetry/src/config.rs
```diff
@@ -190,12 +190,479 @@ type RuntimeConfigurator = Box<
...
+    prefix: Option<String>,
+}
+
+#[derive(Debug)]
+struct ResolvedEnvConfig {
+    enabled: bool,
+    trace_dir: PathBuf,
+
+    // None means the underlying RotatingWriter builder owns the default.
+    rotation_period: Option<Duration>,
```
noxware, line 279: Note: The issue showcases `15` as the proposed default, but in my implementation, I'm simply letting the builder use its existing default (which seems to be `60`).

In general, I tried to avoid having different defaults to the ones that are already in-place. The only exception was `DIAL9_ENABLED` as mentioned in other comment.

dial9-tokio-telemetry/src/config.rs
```diff
@@ -190,12 +190,479 @@ type RuntimeConfigurator = Box<
...
+}
+
+impl<S: EnvSource + ?Sized> EnvSource for &S {
+    fn get(&self, name: &str) -> Result<String, std::env::VarError> {
+        (*self).get(name)
+    }
+}
+
+#[derive(Debug)]
+struct ParsedEnvConfig {
```
noxware, line 241: Split "parsed" and "resolved" mainly to get better observability in tests and to reduce the responsibility of the `parse_env_config` function but I know it adds some duplication.

dial9-tokio-telemetry/src/config.rs
```diff
@@ -190,12 +190,479 @@ type RuntimeConfigurator = Box<
...
+        && let Some(stem) = path.file_stem().and_then(|s| s.to_str())
+        && !stem.trim().is_empty()
+    {
+        return stem.to_string();
+    }
+
+    "unknown-service".to_string()
+}
+
+fn warn(message: fmt::Arguments<'_>) {
```
noxware, line 495: Note: I added this function to mimic the error logging logic from `build_or_disabled` but I would personally just do `tracing::warn!()`, wdyt?

dial9-tokio-telemetry/src/config.rs
```diff
@@ -190,12 +190,479 @@ type RuntimeConfigurator = Box<
...
+    /// | `DIAL9_TASK_DUMP_ENABLED` | `false` | Capture async task dumps at idle yield points. |
+    /// | `DIAL9_TASK_DUMP_IDLE_THRESHOLD_MS` | `10` | Mean idle duration for task dump sampling. |
+    ///
+    /// Missing variables use defaults. Blank, invalid, or non-Unicode values
+    /// emit a warning and are treated as missing. Some numeric defaults come
+    /// from the underlying config builders and are listed here as the current
+    /// `from_env()` behavior. The returned config is built with
+    /// [`Dial9ConfigBuilder::build_or_disabled`], so writer setup failures are
+    /// logged and downgraded to a plain Tokio runtime.
+    pub fn from_env() -> Self {
```
noxware, line 624: We could also have an additional `try_from_env` that returns a `Result` to handle in case something goes wrong.

noxware: Related to this, the issue currently says

> When configuration is invalid, debug builds panic and release builds fallback to disabled

But I don't know if there is much value on having a different behavior in debug for this function, so I skipped that. Opinion?

dial9-tokio-telemetry/src/config.rs
```diff
@@ -190,12 +190,479 @@ type RuntimeConfigurator = Box<
...
+    /// | `DIAL9_TASK_TRACKING_ENABLED` | `true` | Track tasks spawned through dial9 handles. |
+    /// | `DIAL9_RUNTIME_NAME` | unset | Human-readable runtime name in trace metadata. |
+    ///
+    /// Supported S3 variables (`worker-s3` feature required):
+    ///
+    /// | Variable | Default | Meaning |
+    /// | --- | --- | --- |
+    /// | `DIAL9_S3_BUCKET` | unset | Upload sealed trace segments to this bucket. |
+    /// | `DIAL9_SERVICE_NAME` | binary name | Service name used in S3 keys and metadata. |
+    /// | `DIAL9_S3_PREFIX` | unset | Optional S3 object key prefix. |
```
rcoh: lets default this to `dial9-traces`

dial9-tokio-telemetry/src/config.rs
```diff
@@ -190,12 +190,479 @@ type RuntimeConfigurator = Box<
...
+    ///
+    /// Supported CPU profiling variables (`cpu-profiling` feature required):
+    ///
+    /// | Variable | Default | Meaning |
+    /// | --- | --- | --- |
+    /// | `DIAL9_CPU_PROFILE_ENABLED` | `true` on Linux with `cpu-profiling`, `false` otherwise | Enable CPU stack sampling. |
+    /// | `DIAL9_CPU_SAMPLE_HZ` | `99` | CPU sampling frequency in Hz. |
+    /// | `DIAL9_SCHEDULE_PROFILE_ENABLED` | `true` on Linux with `cpu-profiling`, `false` otherwise | Enable per-worker scheduler event capture. |
```
rcoh: link to docs explaining the extra settings required for this to work

rcoh (commented):
look good! two small comments

dial9-tokio-telemetry/src/config.rs
noxware: I did the change. Please note the following implications:
- The default is handled here, so it's different than the builder's one. I suppose that's good for backwards compatibility, so this is not important. 
- My implementation reports all blank strings as invalid (to avoid hard to debug issues caused by "invisible" strings in configuration), so, with this change, there is no way to have "no prefix" through env vars. This is probably important. WDYT?

noxware:
An unrelated CI job seems to be failing due to a missing `README_TELEMETRY.md`, but the important ones are green.

noxware:
@rcoh I noticed https://github.com/dial9-rs/dial9/issues/350, which seems to propose a better default for `max_file_size` than the one I took from `examples/production_use.rs`. I think we should match that proposal here before merge.

Then, when that issue gets addressed in the builder API, we can just change `ResolvedEnvConfig` to hold `None` for `max_file_size`, delegating the default value handling to the programmatic API, keeping a single default source of truth. WDYT?

rcoh (approved):

PR #408, opened 2026-05-15
Extract Source trait for flush-thread data sources

rcoh:
Introduce a `Source` trait that abstracts "thing the flush thread drains into the central collector." CpuProfiler and SchedProfiler now implement Source, and the recorder holds them as `Vec<Box<dyn Source>>` in SharedState instead of separate typed fields.

This is a pure refactor with no behavior change:
- The flush loop iterates sources and calls source.flush(&ctx)
- Worker thread start/stop hooks iterate sources for on_worker_thread_start/stop
- The trait is pub(crate) and gated behind cfg(feature = "cpu-profiling")

Future data sources (e.g. memory profiler) can implement Source without touching the recorder wiring.


jv1i (approved):
Great idea!

nit: this is adding some dead-code warnings (Source trait, FlushContext + fields, methods) when building without cpu-profiling (the trait has no impls without cpu profiling). Can we suppress those somehow? e.g. `#[cfg_attr(not(feature="cpu-profiling"), allow(dead_code))]` (for now)?

PR #409, opened 2026-05-15
Rework release CI & add semver checks

rcoh:
This reworks CI slightly:
1. Releases are not automatic anymore. They _MUST_ be triggered manually. So the flow is now: merge release PR -> cut release. The doc now covers this. This is a little bit more secure and it also makes releasing a bit cleaner.

2. Added tags for all github actions.


CONTRIBUTING.md
```diff
@@ -72,24 +72,35 @@ For other tests, `cargo nextest run` will run all of the normal tests.
...
 
-```
-cargo install release-plz --locked
-git checkout main && release-plz update
-# before committing, make sure that CHANGELOG.md contains an appropriate changelog
-git commit -a
-```
+1. Merge your PRs to `main` using conventional commit messages (e.g. `feat:`, `fix:`, `feat!:` for breaking changes).
+2. The release PR will update automatically. Review the changelog and version bumps.
+3. If you need to adjust versions (e.g. force a major bump), edit `Cargo.toml` versions in the release PR before merging.
```
jlizen, line 92: Might also be worth mentioning changelog improvements / adding code samples is encouraged?

CONTRIBUTING.md
```diff
@@ -72,24 +72,35 @@ For other tests, `cargo nextest run` will run all of the normal tests.
...
+4. Merge the release PR.
+5. Go to Actions → "Publish release" → Run workflow → confirm.
```
jlizen, line 94: We did it this way to keep the token narrower?

jlizen (approved):

CONTRIBUTING.md
rcoh: it also means we can decouple changing versions in cargo.toml and clicking release

jlizen: Wouldn't we get that anyway if we only released from the release-plz pr?

But anyway this is fine

PR #414, opened 2026-05-17
fix: enforce RotatingWriter retention across restarts

dxiao:
`RotatingWriter` enforces `max_total_size` from its in-memory `closed_files` queue. After a process restart that queue starts empty, so existing `trace.N.bin` / `trace.N.bin.gz` artifacts are not counted toward the disk budget and the writer can start again at `trace.0.bin.active`.

That means repeated daemon restarts can grow the trace directory past the configured cap.

This makes startup discover retained artifacts for the same base path, seed `closed_files` oldest-first, and continue writing at the next segment index. Eviction now removes the whole artifact family for a segment (`.bin`, `.bin.gz`, and future write-back suffixes), so background-processed files stay under the same budget.

Stale `.active` files from dead writers are discarded on startup because the worker cannot process them.

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/writer.rs
+++ b/dial9-tokio-telemetry/src/telemetry/writer.rs
@@ -4,7 +4,7 @@ use crate::rate_limit::rate_limited;
 use crate::telemetry::collector::Batch;
 use crate::telemetry::events::clock_pair;
 use crate::telemetry::format::{ClockSyncEvent, SegmentMetadataEvent};
-use std::collections::VecDeque;
+use std::collections::{BTreeMap, VecDeque};
 use std::fs::{self, File};
 use std::io::BufWriter;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -406,6 +425,93 @@ impl RotatingWriter {
...
+        {
+            return Some(SegmentArtifact::Active);
+        }
+
+        let (index, suffix) = rest.split_once(".bin")?;
+        if !suffix.is_empty() && !suffix.starts_with('.') {
+            return None;
+        }
+        Some(SegmentArtifact::Retained {
+            index: index.parse().ok()?,
```
rcoh, line 511: this won't happen currently but because of how the code works if we ever started producing something like `01` as an index, that could never be deleted.

jv1i (approved):

jv1i:
LGTM. I'll follow up with @rcoh comments

PR #417, opened 2026-05-18
move some docs around and add docs/agents

rcoh:


jv1i:
seems this was packed in https://github.com/dial9-rs/dial9/pull/408, we can close this one right?

rcoh:
no this has a separate set of changes

PR #418, opened 2026-05-18
refactor: extract sampling primitives into shared module

rcoh:
Move SplitMix64 and draw_exponential from task_dumped.rs into a new pub(crate) sampling module. Rename draw_exponential_ns → draw_exponential since the math is unit-agnostic (the caller decides whether 'mean' is nanoseconds, bytes, or anything else).

This prepares for the memory profiler (commit 2+), which will reuse the same geometric/Poisson sampling primitive to sample on bytes allocated rather than nanoseconds idle.

Pure refactor — no behavior change. Existing task-dump tests pass unmodified (only the function name changed).

### DIFF
--- a/dial9-tokio-telemetry/src/sampling.rs
+++ b/dial9-tokio-telemetry/src/sampling.rs
@@ -1,0 +1,90 @@
+//! Shared geometric/Poisson sampling primitives.
+//!
+//! Used by the task-dump idle sampler (sampling on nanoseconds) and by the
+//! memory profiler (sampling on bytes). The unit is opaque to the math —
+//! callers pass the mean and treat the returned i64 as a counter in their
+//! native unit.
+
+/// Minimal splitmix64 PRNG. Fast, no dependencies, good enough for sampling.
+pub(crate) struct SplitMix64(u64);
+
+impl SplitMix64 {
+    pub(crate) fn new(seed: u64) -> Self {
+        Self(seed)
+    }
+
+    pub(crate) fn next_u64(&mut self) -> u64 {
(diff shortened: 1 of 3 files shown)


dial9-tokio-telemetry/src/sampling.rs
```diff
@@ -0,0 +1,90 @@
...
+        for _ in 0..1000 {
+            let v = rng.draw_exponential(mean);
+            assert!(v >= 1);
+            if v > 1_000_000 {
+                saw_large = true;
+            }
+        }
+        assert!(saw_large, "expected some values much larger than 1");
+    }
+}
```
Fluzko, line 90: These are _duplicated,_ they still live at `task_dumped.rs`. They're not exactly the same, but pretty much they are, probably worth consolidating them in one place

dial9-tokio-telemetry/src/sampling.rs
```diff
@@ -0,0 +1,90 @@
...
+        z ^ (z >> 31)
+    }
+
+    /// Draw from an exponential distribution with the given mean.
+    /// Returns at least 1 to avoid immediate re-trigger when the
+    /// counter goes to zero.
+    ///
+    /// The unit is whatever the caller treats `mean` as (nanoseconds,
+    /// bytes, etc.).
+    pub(crate) fn draw_exponential(&mut self, mean: u64) -> i64 {
```
Fluzko:
```suggestion
    pub(crate) fn draw_exponential(&mut self, mean: u64) -> u64 {
```
this is guaranteed to return `>1` so I would change it to a narrower and _more correct_ type

Fluzko (approved):

PR #419, opened 2026-05-18
Design updates for memory profiling

rcoh:
Several design doc updates for memory profiling discovered during prototyping. The biggest change is a switch to using a ring buffer.

I also cleaned up the doc because a lot of the pre-requisite work is already done.

### DIFF
--- a/docs/design/memory-profiling.md
+++ b/docs/design/memory-profiling.md
@@ -5,25 +5,28 @@
 Add sampled allocation profiling to dial9. A `Dial9Allocator<A>` wraps any
 `GlobalAlloc` (default `System`) and emits `AllocEvent` / `FreeEvent` events
 into the trace on sampled allocations. Stacks are captured via the frame
-pointer unwinder we already use for CPU profiling. The viewer and analysis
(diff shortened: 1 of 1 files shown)


docs/design/memory-profiling.md
```diff
@@ -854,502 +569,264 @@ fn alloc(&self, layout: Layout) -> *mut u8 {
...
+   profiling this bias is systematic — allocations inside functions
+   with unusually large stack frames (large `Box::pin(future)` state
+   machines, large `[u8; N]` locals) will consistently have truncated
+   stacks. Plan: raise to 1 MiB with empirical validation of
+   false-frame rate. Longer term, a variable-size ring buffer would
+   allow capturing full stacks without a fixed frame cap.
+
+2. **SIGSEGV handler ordering.** Applications that install their own
+   SIGSEGV handler after dial9 initialization may break the unwinder's
+   fault tolerance.
```
jv1i, line 827: good catch! idk how common but might be worth testing and maybe documenting later for users to be aware of

jv1i (approved):

PR #420, opened 2026-05-18
feat: add AllocEvent and FreeEvent to trace format

rcoh:
Adds wire-format definitions for sampled allocation and matching free events that the upcoming memory profiler will emit. Per design §3:

- `AllocEvent { timestamp_ns, tid, size, addr, callchain }`
- `FreeEvent { timestamp_ns, tid, addr, size, alloc_timestamp_ns }`

`size` and `alloc_timestamp_ns` are denormalized onto `FreeEvent` so leak analysis stays useful when the matching `AllocEvent` has been evicted by trace rotation.

This is the schema only — no producer (allocator hook) and no UI flamegraph yet. Both will land in follow-up PRs (ring-buffer scaffold, `Dial9Allocator` wrapper, `MemoryProfiler::install`, hook wiring) per the rollout in design §15.

## API shape

Both events are gated on `#[cfg_attr(not(feature = "unstable-events"), non_exhaustive)]`. By default (no feature) they cannot be constructed by downstream crates, so adding fields later is non-breaking. Crates that need to construct them directly (e.g. for tests or custom event injection) opt in to `unstable-events` and accept that new fields may break them.

## Backwards compatibility

Per AGENTS.md trace-format rules: this PR only **adds** new event types. Old traces don't contain `AllocEvent` / `FreeEvent`, the JS decoder gracefully no-ops on missing schemas, and we don't need old Rust decoders to read new traces.

## Demo trace

Not regenerated — there is no producer of these events yet.

## Verification

- `cargo fmt --check`: clean
- `cargo clippy --all-targets --all-features`: clean
- `cargo nextest run`: 594/594 passed
- `cargo nextest run --stress-duration 20s`: 2 iterations × 594/594 passed

### DIFF
--- a/dial9-tokio-telemetry/Cargo.toml
+++ b/dial9-tokio-telemetry/Cargo.toml
@@ -85,6 +85,9 @@ worker-s3 = [
     "dep:aws-config",
     "dep:time",
 ]
+## Removes #[non_exhaustive] from event structs so downstream crates can
+## construct them directly. Unstable: no semver guarantees on new fields.
+unstable-events = []
 ## All features except platform-specific ones (taskdump). Used in CI for non-Linux targets.
 __nonlinux_all_features = [
     "analysis",
@@ -97,6 +100,7 @@ __nonlinux_all_features = [
 dial9-tokio-telemetry = { path = ".", features = [
     "analysis",
     "tracing-layer",
+    "unstable-events",
     "worker-s3",
 ] }
 assert2 = { workspace = true }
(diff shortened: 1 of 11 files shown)


jv1i (approved):
Looks good!

Side note: I wonder if we eventually want to make `TelemetryEvent` `#[non_exhaustive]`. Though I guess often new additions here mean breaking changes either way, and perhaps for the analysis tool you would like to be aware of new events rather than skip them.

rcoh:
yeah we definitely want to make that `#[non_exhaustive]`. What we _actually_ want is a TelemetryEvent that is much more like how we parse it in JS where its open ended and based only on field names

PR #428, opened 2026-05-19
feat: add ring-buffer scaffold for memory profiling

rcoh:
Adds `pub(crate) mod memory_profiling` (gated on the new `memory-profiling` cargo feature) with:
- RawAlloc<MAX_FRAMES> and RawFree fixed-size POD records.
- RingBuffers<MAX_FRAMES> wrapping two `crossbeam_queue::ArrayQueue`s — alloc queue (default 4096 slots × 1056 B = ~4 MiB at 128 frames) and free queue (default 32K slots × 32 B = ~1 MiB; 8× the alloc slot count per design §9 because frees are pushed unconditionally and allocs only when sampled).
- MemoryProfileSource<MAX_FRAMES> implementing the existing Source trait, draining both queues each flush cycle and emitting AllocEvent / FreeEvent into the trace via the consolidator's ThreadLocalEncoder.

Drain is timestamp-ordered: at each step we emit whichever queue head has the older timestamp. Naive 'all allocs, then all frees' would corrupt the liveset under address reuse within a single flush cycle (alloc → free → alloc-with-same-addr); see the `address_reuse_within_flush_cycle_preserves_liveset` test.

Liveset tracking (HashMap<u64, LivesetEntry>) is wired in but only activates when MemoryProfileSource is constructed with `MemoryProfileSource::new(rings, true)`.

No allocator hook yet (later commit). Tests push synthetic RawAllocs and RawFrees directly into the queues and verify AllocEvents and FreeEvents come out the other end.

### DIFF
--- a/dial9-tokio-telemetry/src/memory_profiling/source.rs
+++ b/dial9-tokio-telemetry/src/memory_profiling/source.rs
@@ -451,4 +451,52 @@ mod tests {
         assert_eq!(entry.size, 512);
         assert_eq!(entry.timestamp_ns, 300);
     }
+
+    /// Demonstrates that `poll_start_ts_or_now` produces strictly ordered
+    /// timestamps even for events that would otherwise share a clock tick —
+    /// the scenario that occurs during a realloc (free old + alloc new at
+    /// same address).
+    #[test]
+    fn monotonic_ts_solves_realloc_ordering() {
+        use crate::telemetry::recorder::poll_start_ts_monotonic;
+
+        // Simulate a realloc: alloc, free, alloc — all at the "same instant".
+        // poll_start_ts_or_now guarantees each gets a distinct, increasing timestamp.
+        let t1 = poll_start_ts_monotonic();
+        let t2 = poll_start_ts_monotonic();
+        let t3 = poll_start_ts_monotonic();
+        assert!(t1 < t2 && t2 < t3, "timestamps must be strictly ordered");
+
+        let rings = rings(16, 16);
+        rings.alloc_queue.push(make_raw_alloc(0x6000, 256, t1)).ok();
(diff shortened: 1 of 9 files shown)


dial9-tokio-telemetry/src/memory_profiling/source.rs
```diff
@@ -0,0 +1,454 @@
...
+                }
+                (Some(a), Some(f)) => {
+                    if a.ts_ns <= f.ts_ns {
+                        let a = next_alloc.take().expect("checked Some above");
+                        self.handle_alloc(a, ctx);
+                        next_alloc = self.rings.alloc_queue.pop();
+                    } else {
+                        let f = next_free.take().expect("checked Some above");
+                        self.handle_free(f, ctx);
+                        next_free = self.rings.free_queue.pop();
```
jv1i, line 140: (I might be missing context here) with `ReusePollStart` as the default (per design doc), every alloc/free in a poll shares one `ts_ns`, so a same-address realloc will emit alloc/free/alloc with no ordering between them right?
If so this tiebreak here can't separate them and the `free` would end up evicting the still-live realloc'd entry. (`address_reuse..` test only feeds increasing timestamps, don't think it catches this)
So do we want a seq field on RawAlloc/RawFree or maybe a doc note that liveset needs `Precise`? (and/or if `track_liveset = true` then default to `Precise`?)

jv1i (approved):
Looks good to me! leaving a note about one thing I'm unsure about, don't need to address it on this PR.

dial9-tokio-telemetry/src/memory_profiling/source.rs
rcoh: that's a good callout! I changed the pollstart timestamp hook so that it is monotonic

PR #433, opened 2026-05-20
refactor: split recorder/mod.rs into focused modules

rcoh:
Split the 3500-line god module into four focused files:

- **flush_loop.rs**: `DrainState`, `FlushStats`, `flush_once`, `run_flush_loop`
- **handle.rs**: `TelemetryHandle`, `RuntimeTelemetryHandle`, `spawn`, `InstrumentedSpawnGuard`
- **guard.rs**: `TelemetryGuard`, `TraceRuntimeCoreBuilder`, `WorkerHandle`
- **builder.rs**: `TracedRuntimeBuilder` (typestate), `TelemetryCore`, `TracedRuntime`, `assemble_processors`, `PipelineConfig`, `TelemetryRuntimeError`

`mod.rs` retains only the glue: module declarations, re-exports, `register_hooks`, `attach_runtime`, `ControlCommand`, and tests.

**Motivation**: Understanding any one concept (e.g. "how does the flush loop work?") previously required scrolling through 2000+ lines of unrelated builder machinery. Each extracted module now has a narrow interface and concentrated locality.

**No behavioral changes.** All 332 tests pass, including stress runs.

PR #434, opened 2026-05-20
refactor: inline EventWriter, delete the shallow wrapper

rcoh:
**Stacks on #433** (split recorder/mod.rs).

`EventWriter` was a pass-through layer wrapping `Box<dyn TraceWriter>` with a single counter. Its 10-method interface was pure delegation — it fails the deletion test.

**Changes:**
- Delete `event_writer.rs` entirely (100 lines removed)
- Move `flush_sources` to `SharedState` (where the sources already live)
- Move `events_written` counter into the flush loop as a local variable
- `flush_once` and `run_flush_loop` now take `&mut dyn TraceWriter` / `Box<dyn TraceWriter>` directly

**No behavioral changes.** All 332 tests pass, including stress runs.


jv1i (approved):
Looks good!
nit: some comments removed on `flush_loop.rs`

PR #435, opened 2026-05-20
feat!: in-memory pipeline

jv1i:
https://github.com/dial9-rs/dial9/issues/279

# Overview

Adds an in-memory writer (`InMemoryWriter`) via a new Fs abstraction, that allows dial9 to be used without writing to disk.

## Public API

```rust
let writer = InMemoryWriter::new(max_total_size)?;

// or, with knobs
let writer = InMemoryWriter::builder()
    .max_total_size(16 * MB)
    .max_segment_size(1 * MB)      // optional, dial9 picks if absent
    .rotation_period(Duration::from_secs(60))
    .segment_metadata(vec![("svc".into(), "trace".into())])
    .build()?;
```

`InMemoryWriter::builder` requires `max_total_size >= 3 × max_segment_size` so it can bound writer + non-cpu pipeline RAM (1 active + 1 in-flight + 1 ring slot minimum (see Memory Usage below).

**Breaking:** `RotatingWriter` renamed to `DiskWriter`. The writer is now generic over backend (`SegmentWriter<Mode>`) with `DiskWriter` / `InMemoryWriter` as the public types; memory constructors are `InMemoryWriter::new` / `::builder` (were `new_in_memory` / `in_memory_builder`).

**Breaking:** `SegmentData::segment()` returns `&SegmentRef` (`Disk` / `Memory`) instead of `&SealedSegment`. External processors that called `.path()` need to match on the enum.

### Compile-time gating
  
`.write_back()` paired with a memory writer is a compile error. `PipelineBuilder<Mode>::write_back` only exists on `<Disk>`, and `build_and_start` unifies the builder's `Mode` with the writer arg, so the closure body and the writer have to agree.

Not yet catching:`WriteBackProcessor` added via `.pipe(WriteBackProcessor)` instead of `.write_back()`. Bypasses the gated method, compiles, fails at runtime.

Non-goal (on its own, might happen in follow-up as part of pipeline changes): memory writer + no delivery processor (no S3, no custom uploader). Compiles, ring drains and data is silently discarded. Surfacing either would need a Delivery typestate or marker-trait scheme. IMO friction outweighs the value on what I have sketched. Folding into the broader pipeline/processor pass listed in follow-ups.

## Memory usage

`max_total_size` is a hard cap on the writer + pipeline RAM (non-cpu mode). Internally three pools share it:

- Active buffer: `1 × max_segment_size`. Writer fills one at a time before sealing.
- In-flight: `1 × max_segment_size`. One segment held by the worker. Pipeline is serial.
- Ring: `max_total_size - 2 × max_segment_size`. Sealed segments waiting for the worker. Drop-oldest on overflow.

Validation requires `max_total_size >= 3 × max_segment_size` so the ring has room for at least one sealed segment.

`in_flight_bytes` tracks the post-stage payload size as it mutates through processors. It does not capture intra-stage transients (e.g. gzip's input + growing output overlap), which fit inside the in-flight reserve.

> ⚠ **`cpu-profiling` adds memory that `max_total_size` does not cover.** Perf sampler ring buffers (e.g. ~8 MiB at 4 CPUs) + blazesym symbol cache (~20-30 MiB once warm) live outside any writer config (pre-existing behavior). Additionally, the symbolize stage appends a symbol-table chunk to the payload whose size scales with unique stack frames, currently uncapped. We could fold both into `max_total_size` (IMHO better), or be very transparent about `max_total_size` not meaning "dial9 will use at max X amount of RAM"

### Bench results

Ran a workload that mirrors `examples/long_workload.rs`: TCP echo server + 16 chatty clients (varied delays) + 3 CPU-burst tasks. 10s, 4 worker threads, 128 KiB segments, 3s rotation. Ring/in-flight numbers from `WorkerCycleMetrics` via `with_worker_metrics_sink`.

> These rows are from a local variant of the committed `pipeline_heap_bench`: adds the `mem-backpressure` / `mem-evicting` modes and the `WorkerCycleMetrics` columns, and runs the TCP-echo workload at 128 KiB segments. The committed bench uses a simpler workload at 512 KiB segments.

| mode | cap | workload Δ | ring peak | in-flight peak | dispatched | evicted |
|---|---|---|---|---|---|---|
| disk               | —       | 12.21 MiB | —        | —        | 8  | 0 |
| mem                | 16 MiB  | 10.32 MiB | 0 KiB    | 233 KiB  | 8  | 0 |
| mem-backpressure   | 16 MiB  | 10.91 MiB | 478 KiB  | 248 KiB  | 8  | 0 |
| mem-evicting       | 384 KiB | 10.67 MiB | 0 KiB    | 253 KiB  | 5  | **4** |
| disk-cpu-profiling | —       | 45.81 MiB | —        | —        | 9  | 0 |
| mem-cpu-profiling  | 16 MiB  | 46.44 MiB | 228 KiB  | 291 KiB  | 9  | 0 |

`mem-backpressure` and `mem-evicting` attach a `SlowSink` after gzip that sleeps 2s per segment, standing in for S3 PUT under transient regional slowdown. `mem-evicting` uses the minimum-allowed cap (`3 × max_segment_size`) to force production past drain.

Reading:

- **`mem` vs `disk`**: gzip+noop drains every cycle, ring stays empty, heap delta ≈ 0.
- **`mem-backpressure`**: slow drain queues ~480 KiB (~4 segments) in the ring, 16 MiB cap was generous enough that nothing evicts.
- **`mem-evicting`**: slow drain against the minimum cap. Producer outruns drain, eviction fires (4 of 9 segments evicted). Peak heap stays bounded by the cap.
- **cpu-profiling** modes are dominated by perf ring buffers + blazesym cache (~30-35 MiB), independent of writer cap.

#### Sustained-load comparison via dial9's own memory profiler

5 min run, 16 TCP clients + CPU bursts,`max_total_size = 64 MiB`, `max_segment_size = 2 MiB`, `rotation_period = 60s`. dial9's allocator hook installed with `track_liveset = true`, sample rate 4 KiB. Symbolize ran inline: blazesym/gimli/dwarf and allocator-infra frames filtered out so each sample bins under the first dial9, tokio, or workload frame.

Numbers are approximate (4 KiB sampling under-counts the small-allocation tail, so retained heap is a lower bound):

| | disk | mem | Δ |
|---|---:|---:|---:|
| churn | ~1,770 MiB | ~1,690 MiB | **−80 MiB** |
| retained at workload end | ~73 MiB | ~79 MiB | **+6 MiB** |

- **Churn:** mem ~80 MiB less. Disk re-reads sealed segments from disk (`std::fs::read`) and runs the flush loop; mem skips both, segments arrive in-process via the ring.
- **Retained:** mem ~6 MiB more, ring handoff vs disk re-reading into a fresh `Vec`. No drift over the 5 min window (end-of-run ≈ 1-min warmup, so no leak).

Both retained figures are dial9 internals sitting **outside** `max_total_size`, dominated by payload copies (scale with `max_segment_size`) and per-thread drain state (scale with `worker_threads`). **Sizing:** at this config, budget ~75-80 MiB for dial9 internals on top of `max_total_size` (plus cpu-profiling overhead from above, if enabled).

## Writer changes

New `pub(crate) Fs` enum (`Disk` / `Mem`) covers the writer + worker segment lifecycle (create / seal / remove / take_files / wait_for_more / writer_done). Disk path unchanged, stays default. Memory routes encoded `Bytes` through an in-process ring channel to the existing `SegmentProcessor` pipeline. Drop-oldest eviction on a byte budget.

New metrics: `WorkerCycleMetrics` per `take_files` cycle, symmetric across backends.

### Retry semantics

On `retryable: true`, the worker re-enqueues the original seal'd bytes (Arc-clone snapshot taken at dispense, before any processor mutates the payload). Symmetric with disk re-reading the file. Bounded by `MEMORY_RETRY_BUDGET = 3` because each re-enqueue pins a ring slot  (disk has no equivalent cost so its retry stays unbounded).

## Tests

- unit + e2e: retry within budget recovers, budget exhaustion drops, end-to-end through real `S3PipelineUploader` + `CircuitBreaker` against a fake S3 with injected errors
- multi-thread race on prod `run()` shutdown / `writer_done` handoff (lost-wakeup safe)
- fs.rs unit + shuttle (atomics + queue handoff)
- accounting delta across pipeline stages (`in_flight_bytes` follows payload mutation)

## Follow-ups

- Update dial9 macro to support memory configuration
- Potentially another pass at the public API with broader breaking changes to pipeline setup / processor API
- cpu-profling mem usage mentioned above
- perhaps gate PRs on mem usage? (with the added heap bench), currently just reports to bencher


dial9-tokio-telemetry/src/background_task/fs/mem.rs
```diff
@@ -0,0 +1,431 @@
...
+        Ok(data.len())
+    }
+    fn flush(&mut self) -> io::Result<()> {
+        Ok(())
+    }
+}
+
+struct MemSealedSegment {
+    index: u32,
+    size: u64,
```
rcoh: do we need `size`? presumably we have `bytes.len()`

dial9-tokio-telemetry/src/background_task/fs/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+pub(crate) enum Fs {
+    Disk(DiskFs),
+    Mem(MemFs),
+}
+
+impl Fs {
+    /// Create a new active-segment write handle.
+    pub(crate) fn create(&self, path: &Path) -> io::Result<ActiveHandle> {
+        match self {
+            Fs::Disk(d) => d.create_handle(path),
```
rcoh: naming these methodsthe same as the parent is helpful. `create_segment`?

dial9-tokio-telemetry/src/background_task/fs/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+            Fs::Mem(m) => m.create_handle(path),
+        }
+    }
+
+    pub(crate) fn disk(base_path: &Path) -> Arc<Self> {
```
rcoh: new_disk?

dial9-tokio-telemetry/src/background_task/fs/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+        Arc::new(Fs::Disk(DiskFs::from_base_path(base_path)))
+    }
+
+    pub(crate) fn memory(max_segment_size: u64, max_total_size: u64) -> io::Result<Arc<Self>> {
```
rcoh: new_in_memory?

dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -346,9 +381,59 @@ impl RotatingWriter {
...
+    /// Create an in-memory writer. Encoded bytes flow through an in-process
+    /// ring instead of disk.
+    ///
+    /// `max_segment_size` bounds each segment. The ring holds up to
+    /// `max_total_size / max_segment_size` segments; oldest is dropped on
+    /// overflow.
+    ///
+    /// Returns `Err(InvalidInput)` if `max_segment_size` is 0 or larger
+    /// than `max_total_size`.
+    pub fn in_memory(max_segment_size: u64, max_total_size: u64) -> std::io::Result<Self> {
```
rcoh: recommend making this a builder. I suspect we will really just want to have a max_total_size and leave it up to dial9 to pick a reasonable segment size

dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -346,9 +381,59 @@ impl RotatingWriter {
...
+            drain_interval,
+            next_drain_time: Self::next_boundary(now, drain_interval),
+            fs,
+            tracks_closed_files: false,
+            _mode: PhantomData,
+        })
+    }
+}
+
+impl<M: WriterMode> RotatingWriter<M> {
```
rcoh: we could potentially alias `pub type InMemoryWriter = RotatingWriter<Memory>` which may improve the public API

dial9-tokio-telemetry/src/metrics.rs
```diff
@@ -36,6 +37,29 @@ pub(crate) struct FlushMetrics {
...
 
+/// Metrics emitted once per worker cycle
+#[metrics(rename_all = "PascalCase")]
+#[derive(Debug)]
+pub(crate) struct WorkerCycleMetrics {
+    pub operation: Operation,
+    /// Memory ring depth: segments waiting in the in-process channel,
+    /// not popped this cycle. `None` on disk (disk backpressure
+    /// surfaces as `in_flight_*` + `dropped_segments`).
+    pub ring_depth: Option<u64>,
```
rcoh: confusing name...perhaps `in_flight_segments`? how is this different than in_flight_count?

dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -2398,4 +2301,182 @@ mod tests {
...
+
+        for _ in 0..EVENTS {
+            writer.write_encoded_batch(&test_batch()).unwrap();
+        }
+        // Seals the active segment onto the ring and signals writer_done.
+        writer.finalize().unwrap();
+
+        let captured = Arc::new(Mutex::new(Vec::<u8>::new()));
+
+        struct CaptureProcessor(Arc<Mutex<Vec<u8>>>);
```
rcoh: move to testutil?

rcoh:
should we rename RotatingWriter to `RotatingDiskWriter` or something like that?

rcoh (approved):
LGTM, awesome work. We might as well merge #470 and cut one more release before we merge this and start work towards 4.0

jv1i:
> should we rename RotatingWriter to RotatingDiskWriter or something like that?

@rcoh we already had a `DiskWriter` alias here for `RotatingWriter` (with `RotatingWriter` defaulting to disk mode), so to make things more explicit and symmetrical, what if we rename `RotatingWriter` more generically to `SegmentWriter` and have users use the alises for both (not just memory)? did that now on [aef1a68](https://github.com/dial9-rs/dial9/pull/435/commits/aef1a6843749f68106e82105fe119718b626149b)
  
with that I also renamed to `InMemoryWriter::new` and `DiskWriter::new` instead of having `RotatingWriter::new_in_memory`

I think that's a bit more clear than having `RotatingDiskWriter/DiskWriter` that also has a memory mode, WDYT?

rcoh:
Yeah that makes sense

PR #436, opened 2026-05-20
fix: detect block_in_place gaps and correct CPU sample worker attribution

### LINKED ISSUE #427: utilize tid in workerpark/unpark to attribute CPU samples
Right now, CPU samples have a worker id. However, when block_in_place is used this doesn't work and the trace files have nonsense attribution of CPU samples

The analysis code + viewer should locate the worker the unparked with the correct tid.

We probably need to maintain some sort of data structure that tracks worker tid, or read through the events linearly by timestamp and properly attribute worker id to the data

rcoh:
Fixes #427. I attempted to provide some specific block-in-place insights in the viewer but the main fix here is that we properly keep track of worker tids.

During `block_in_place`, the producer-side `worker_id` on CPU samples is wrong because the tid→worker mapping doesn't update.


dial9-viewer/ui/trace_parser.js
```diff
@@ -111,6 +111,135 @@
...
+    // Build per-worker gap lists for sample rewriting.
+    // We index by worker because gaps belong to a worker, but we suppress
+    // samples by tid: any sample whose tid matches the gap's fromTid OR
+    // toTid, falling inside the gap window, is unattributable.
+    const gapsByWorker = new Map();
+    for (const g of gaps) {
+      let arr = gapsByWorker.get(g.workerId);
+      if (!arr) { arr = []; gapsByWorker.set(g.workerId, arr); }
+      arr.push(g);
+    }
```
jv1i, line 203: I don't think this `gapsByWorker` is being used anywhere right?

jv1i (approved):

Issue #437, opened 2026-05-20
Support Custom Events in Viewer


nickrobinson:
The idea here is to be able to view custom events alongside tracing span data in dial9-viewer. This would be helpful for visually correlating custom events with interesting long polls/span data.

PR #438, opened 2026-05-20
feat(viewer): add custom events view

nickrobinson:
This PR adds a custom event viewer section. This allows custom events to be displayed next to tracing span data in the UI.

Screenshot 2026-05-20 at 11 50 24 AM

Addresses #437

### DIFF
--- a/dial9-viewer/ui/viewer.html
+++ b/dial9-viewer/ui/viewer.html
@@ -428,6 +428,28 @@
                 height: 100%;
                 display: block;
             }
+            #custom-events-panel {
+                display: none;
+                height: 40px;
+                background: #1a1a2e;
+                border-top: 1px solid #333;
+                flex-shrink: 0;
+                position: relative;
+                overflow: hidden;
+            }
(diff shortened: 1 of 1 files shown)


rcoh:
awesome! I'm fine merging this as is. From your usage (I can see in your screenshot you also have spans in the trace). Do you think it would be good to show the events in the spans panel somehow or did you find it helpful to have them split out?

nickrobinson:
> awesome! I'm fine merging this as is. From your usage (I can see in your screenshot you also have spans in the trace). Do you think it would be good to show the events in the spans panel somehow or did you find it helpful to have them split out?

I could go either way there. I do think a user would likely want some clear visual distinction around these if they appear in the span view to be able to differentiate. Ours is so noisy at the moment it made it easier to break it out into its own section. Let me know what you think, I'm happy to try it the other way and see how it looks. There isn't a ton of additional real estate to work with here :)

rcoh:
lets merge this as is and we can always iterate

rcoh (approved):

PR #441, opened 2026-05-20
Add connection-established / closed events to the demo trace

### LINKED ISSUE #440: add custom events into the demo trace
the viewer now supports custom events, ensure that the demo app generates a few

rcoh:
Emit a DdbFlushWire custom event (with interned metric_name) after each individual DDB put in the metrics-service flush loop. This exercises the custom events feature in the demo trace so the viewer showcases it.

Closes #440


rcoh:
I'm wondering now if new connections opening or something like that would be a more interesting event

examples/metrics-service/src/buffer.rs
```diff
@@ -69,13 +103,26 @@ impl MetricsBuffer {
...
                 .instrument(span)
-                .await
-            {
+                .await;
+            let latency_us = start.elapsed().as_micros() as u64;
+            let success = result.is_ok();
+            record_event(
+                DdbFlush {
+                    timestamp_ns: clock_monotonic_ns(),
+                    metric_name: name.clone(),
```
Fluzko: I think this `clone` can be avoided, since later is consumed directly by `record_event` and reborrowed at `intern_string` call. Pretty sure we can hold a reference instead.

examples/metrics-service/src/buffer.rs
```diff
@@ -1,8 +1,42 @@
...
+impl Encodable for DdbFlush {
+    fn encode(&self, enc: &mut ThreadLocalEncoder<'_>) {
+        let metric_name = enc.intern_string(&self.metric_name);
+        enc.encode(&DdbFlushWire {
+            timestamp_ns: self.timestamp_ns,
+            metric_name,
+            latency_us: self.latency_us,
+            success: self.success,
+        });
+    }
```
Fluzko: I think this will create an event with the name of the struct, which feels unexpected. Probably worth just a name change. deferring it to you, just wanted to surface it

Fluzko (approved):

Fluzko (approved):

Issue #287, opened 2026-04-24
Add heap profiling / leak detection


rcoh:
All the knobs are there to do efficient memory profiling:

- [ ] Land #222 which will allow space-efficient storage of stack frames (which are extremely repetitive)
- [ ] Make the framepointer unwinding accessible as a library function (or use the libunwind solution from #213)
- [ ] Add a custom allocator that wraps another allocator and emits events at a defined sampling frequency`Alloc { addr, stackframe }`, `Free { addr, stackframe }` (probably a small amount of additional metadata is also required). 
- [ ] Update the viewer and analysis scripts to be able to spot memory leaks and show allocation profiles

PR #442, opened 2026-05-20
feat: add `Dial9Allocator`, a profiling allocator that feeds events into dial9 traces

rcoh:
This implements the dial9 core side #287 

Remaining work:
- [ ] README updates
- [x] integration into example app
- [x] update SKILL
- [x] update viewer

### DIFF
--- a/dial9-tokio-telemetry/src/memory_profiling/hook.rs
+++ b/dial9-tokio-telemetry/src/memory_profiling/hook.rs
@@ -1,3 +1,4 @@
+#![deny(clippy::arithmetic_side_effects)]
 //! Allocator hook — the hot path called from `Dial9Allocator::alloc`/etc.
 //!
 //! # Soundness contract
@@ -33,7 +34,6 @@ use crate::memory_profiling::profiler::MemoryProfilerInner;
 use crate::memory_profiling::ring::{DEFAULT_MAX_FRAMES, RawAlloc, RawFree};
 use crate::sampling::SplitMix64;
 use crate::telemetry::events::current_tid;
-use std::sync::atomic::Ordering;
 
 /// Per-thread sampling state. Held in TLS so each thread reads/writes
 /// its own counter and PRNG without synchronization (design §1).
@@ -86,7 +86,8 @@ fn ensure_initialized(state: &mut SamplingState, inner: &MemoryProfilerInner) {
         }
     };
     state.rng = SplitMix64::new(seed);
-    state.next_sample_bytes = state.rng.draw_exponential(inner.sample_rate_bytes) as i64;
+    state.next_sample_bytes =
+        i64::try_from(state.rng.draw_exponential(inner.sample_rate_bytes)).unwrap_or(i64::MAX);
     state.initialized = true;
 }
 
@@ -117,7 +118,15 @@ pub(crate) fn on_alloc(inner: &MemoryProfilerInner, ptr: *mut u8, size: usize) {
         };
         ensure_initialized(&mut state, inner);
 
-        let remaining = state.next_sample_bytes - size as i64;
+        // Saturate at i64::MAX — allocations this large are unrealistic but we
+        // handle them defensively rather than wrapping.
+        let size_i64 = i64::try_from(size).unwrap_or(i64::MAX);
+
+        // OK: intentionally allowed to go negative — the design uses signed arithmetic
+        // so that a large allocation exceeding the remaining budget produces a negative value,
+        // which triggers sampling on the next line.
+        #[allow(clippy::arithmetic_side_effects)]
+        let remaining = state.next_sample_bytes - size_i64;
         if remaining > 0 {
             state.next_sample_bytes = remaining;
             return;
@@ -128,7 +137,10 @@ pub(crate) fn on_alloc(inner: &MemoryProfilerInner, ptr: *mut u8, size: usize) {
         // by more than one draw's worth (design §1).
         let mut next = remaining;
         while next <= 0 {
-            next = next.saturating_add(state.rng.draw_exponential(inner.sample_rate_bytes) as i64);
+            next = next.saturating_add(
+                i64::try_from(state.rng.draw_exponential(inner.sample_rate_bytes))
+                    .unwrap_or(i64::MAX),
+            );
         }
         state.next_sample_bytes = next;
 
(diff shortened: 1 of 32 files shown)


dial9-tokio-telemetry/src/memory_profiling/source.rs
```diff
@@ -147,20 +151,22 @@ impl<const MAX_FRAMES: usize> Source for MemoryProfileSource<MAX_FRAMES> {
     fn name(&self) -> &'static str {
         "memory"
     }
+
+    fn as_any(&self) -> &dyn std::any::Any {
+        self
```
Fluzko: This downcasting is for tests only, right? I would move it to `#[cfg(test)]` to reduce noise

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -65,6 +65,12 @@ pub(crate) fn poll_start_ts_monotonic() -> u64 {
     })
 }
 
+/// Alias for `poll_start_ts_monotonic` used by the memory profiler hook.
+#[cfg(feature = "memory-profiling")]
+pub(crate) fn poll_start_ts_or_now() -> u64 {
```
Fluzko: This doesn't match its documented fallback behavior. `TimestampMode::ReusePollStart` docs say "_Falls back to clock_monotonic_ns() when called outside a task poll._" But the implementation is a pure alias for `poll_start_ts_monotonic()` with no fallback.

rcoh: Poll_start_ts_monotonic falls back I believe 

I think we should unify these time primitives into one module though

dial9-tokio-telemetry/src/memory_profiling/source.rs
rcoh: yeah good catch, I removed this

dial9-tokio-telemetry/src/memory_profiling/hook.rs
```diff
@@ -0,0 +1,600 @@
...
+}
+
+#[inline]
+fn stamp(mode: TimestampMode) -> u64 {
+    match mode {
+        TimestampMode::ReusePollStart => crate::telemetry::recorder::poll_start_ts_or_now(),
+        TimestampMode::None => 0,
+        TimestampMode::Precise => crate::telemetry::events::clock_monotonic_ns(),
+    }
+}
```
jv1i, line 121: Wondering if `None + track_liveset=true` is safe, with every event at `ts=0`, the merge in source can't order free/alloc on the same addr, so in-place realloc would evict the new entry from the liveset. Should we forbid that combo at install, or am I missing something? related to https://github.com/dial9-rs/dial9/pull/428#discussion_r3268063666, saw the fix for ReusePollStart, but I think that doesn't cover this mode.

dial9-tokio-telemetry/src/memory_profiling/ring.rs
```diff
@@ -1,37 +1,46 @@
...
 pub(crate) const DEFAULT_ALLOC_QUEUE_CAPACITY: usize = 4096;
 
-/// Default number of `RawFree` slots. 8× the alloc queue (design §9).
-#[expect(dead_code, reason = "wired up by allocator hook in a later commit")]
+/// Default number of `RawFree` slots. 8× the alloc queue.
+#[expect(
+    dead_code,
+    reason = "documents the sizing rationale; config uses ring_capacity * 8"
+)]
 pub(crate) const DEFAULT_FREE_QUEUE_CAPACITY: usize = DEFAULT_ALLOC_QUEUE_CAPACITY * 8;
```
jv1i: very nit: neither of these const are being actually used right? I understand `DEFAULT_FREE_QUEUE_CAPACITY` is just for documenting, but reading this I'd expect `DEFAULT_ALLOC_QUEUE_CAPACITY` to be maybe wired somewhere?

dial9-tokio-telemetry/tests/memory_profiling_hook_realloc.rs
```diff
@@ -0,0 +1,99 @@
...
+        .collect();
+
+    assert!(
+        !allocs.is_empty(),
+        "expected at least one AllocEvent from realloc path"
+    );
+    assert!(
+        !frees.is_empty(),
+        "expected at least one FreeEvent when liveset tracking is on"
+    );
```
jv1i, line 98: nit: could assert each `FreeEvent` matches a previously-recorded `AllocEvent` on (addr, size, alloc_timestamp_nanos).

dial9-tokio-telemetry/src/telemetry/recorder/flush_loop.rs
```diff
@@ -159,14 +159,22 @@ pub(super) fn run_flush_loop(
...
         let runtime_entries: Vec<(String, String)> =
             contexts.iter().filter_map(|c| c.metadata_entry()).collect();
-        if !runtime_entries.is_empty() {
+        let source_entries: Vec<(String, String)> = shared
+            .sources
+            .lock()
+            .unwrap()
+            .iter()
+            .flat_map(|s| s.segment_metadata())
+            .collect();
```
jv1i, line 173: should we cache this in `new()` so we don't rebuild the Strings every flush cycle? might be minor since the merge in writer short-circuits anyway.

dial9-tokio-telemetry/src/memory_profiling/hook.rs
```diff
@@ -0,0 +1,600 @@
...
+        let sample = RawAlloc {
+            tid: current_tid(),
+            size: size as u64,
+            addr: ptr as u64,
+            ts_ns: timestamp_ns,
+            frames,
+            frame_count: result.frames_written.min(DEFAULT_MAX_FRAMES) as u8,
+        };
+
+        inner.rings.push_alloc(sample);
```
jv1i, line 196: some drift from doc, `addr` here is always set, doc says it should be 0 if liveset is off. should we match that or is it ok to always include? I'm unsure about the intent

jv1i (approved):
Looks great! leaving some minor comments.

dial9-tokio-telemetry/src/memory_profiling/hook.rs
rcoh: great catch

PR #443, opened 2026-05-22
feat: Analysis toolkit for memory profiling [stacked]

rcoh:
This adds memory profiling to the metrics service and analysis instructions. This can produce, among many other things, allocations by spawn location:

```
  ┌──────────────────────────────────────────┬─────────┬────────────┐
  │ Spawn Location                           │ Samples │ Est. Bytes │
  ├──────────────────────────────────────────┼─────────┼────────────┤
  │ axum_traced.rs:173:32 (request handlers) │ 277     │ ~147 MB    │
  ├──────────────────────────────────────────┼─────────┼────────────┤
  │ hyper-util/.../tokio.rs:115:9 (hyper IO) │ 40      │ ~21 MB     │
  ├──────────────────────────────────────────┼─────────┼────────────┤
  │ main.rs:253:24                           │ 3       │ ~1.6 MB    │
  ├──────────────────────────────────────────┼─────────┼────────────┤
  │ main.rs:296:24                           │ 1       │ ~1.2 MB    │
  ├──────────────────────────────────────────┼─────────┼────────────┤
  │ main.rs:231:14                           │ 1       │ ~525 KB    │
  ├──────────────────────────────────────────┼─────────┼────────────┤
  │ (non-worker threads)                     │ 30      │ ~23 MB     │
  ├──────────────────────────────────────────┼─────────┼────────────┤
  │ Total                                    │ 352     │ ~195 MB    │
```


dial9-viewer/ui/test_trace_analysis.js
```diff
@@ -887,35 +888,6 @@ async function main() {
...
-    if (actives.length !== 1) {
-      fail(`Expected 1 active span (gap-crossing suppressed), got ${actives.length}: ${JSON.stringify(actives)}`);
-      return;
-    }
-    if (actives[0].start !== 60 || actives[0].end !== 70) {
-      fail(`Expected active [60,70), got [${actives[0].start},${actives[0].end})`);
-      return;
-    }
-    pass("Active span crossing block-in-place gap is suppressed; clean span preserved");
-  }
```
jv1i, line 917: is removing this test intentional?

dial9-viewer/ui/trace_analysis.js
```diff
@@ -855,6 +855,140 @@
...
+      site.count++;
+      site.estimatedBytes += allocWeight(a.size);
+    }
+    const topSites = [...siteMap.values()].sort((a, b) => b.estimatedBytes - a.estimatedBytes).slice(0, 10);
+
+    // Leaks: allocs with no matching free
+    const leaks = [];
+    let leakedBytes = 0;
+    for (const a of allocEvents) {
+      if (!freedAddrs.has(a.addr)) {
```
jv1i: didn't test this so I might not be tracking this right, but should we key by both addr and alloc timestamp to prevent false-positives on address reuse?

jv1i (approved):

dial9-viewer/ui/test_trace_analysis.js
rcoh: Not intentional — accidentally deleted when the analyzeAllocations tests were added in the same region. Restored in eabaa83.

dial9-viewer/ui/trace_analysis.js
rcoh: Good catch — you're right. Fixed by keying with `addr:allocTimestampNs` instead of just `addr`. Added a test that exercises address reuse. Fixed in eabaa83.

Issue #446, opened 2026-05-22
design: Serde-based Trace Deserialization


rcoh:
# PRD: Serde-based Trace Deserialization

## Problem Statement

The Rust-side trace decoding is built around a closed `TelemetryEvent` enum that must enumerate every possible event type. This is a fundamental impedance mismatch with the wire format, which is self-describing and open-ended — different crates can define different event types via `#[derive(TraceEvent)]`. Adding a new event type requires modifying the central enum, adding `TelemetryEventRef` variants, `decode_ref` match arms, and `to_owned_event` conversions. Downstream consumers cannot decode events without recompiling against the latest enum. The JS parser already works the right way (schema-driven, field-name-based), but the Rust side forces a closed-world assumption.

## Solution

Replace the closed `TelemetryEvent` enum with a serde `Deserializer` that lets consumers bring their own `#[derive(Deserialize)]` types. Users define an enum with exactly the event types they care about, and unknown events are handled via `#[serde(other)]`. The wire format's self-describing schema maps naturally to serde's internally-tagged enum model.

```rust
#[derive(Deserialize)]
#[serde(tag = "event")]
enum MyEvents {
    #[serde(rename = "PollStartEvent")]
    PollStart(PollStartEvent),
    #[serde(rename = "WakeEventEvent")]
    Wake(WakeEventEvent),
    #[serde(other)]
    Unknown,
}

for result in decoder.deserialize_events::<MyEvents>() {
    match result {
        Ok(event) => { /* typed access */ }
        Err(raw) => { /* inspect raw.name, raw.fields, raw.error */ }
    }
}
```

## User Stories

1. As a trace consumer, I want to define my own subset of events to deserialize, so that I don't need to handle event types I don't care about.
2. As a trace consumer, I want unknown event types to be gracefully skipped via `#[serde(other)]`, so that my code works with traces from newer producers.
3. As a trace consumer, I want `PooledString` fields to transparently resolve to `String` (or `&str` with a lifetime), so that I don't need to manually interact with the string pool.
4. As a trace consumer, I want `PooledStackFrames` fields to transparently resolve to `Vec<u64>` (or `&[u64]`), so that callchains are immediately usable.
5. As a trace consumer, I want the frame-header timestamp injected as a `timestamp_ns` field, so that my struct can include it without special handling.
6. As a trace consumer, I want an owned iterator API that yields `Result<E, OwnedRawEvent>`, so that I can process events in a loop without callbacks.
[...]

PR #447, opened 2026-05-22
add Deserialize impl for dial9-trace-format

rcoh:
This is the first in a series of changes en route to #446 where we will fully rip out telemetry event and replace it with something much more akin to what we currently do on the JS side.

This makes _deserializing_ actually function as a serde Format. I ported one test to use it. We will port the remaining tests to use it in a future PR.

After that, we will fully delete telemetry event in route to 0.4

### DIFF
--- a/dial9-trace-format/src/de.rs
+++ b/dial9-trace-format/src/de.rs
@@ -71,7 +71,7 @@ pub struct DeserError(String);
 
 impl DeserError {
     /// Construct a new error with the given message.
-    pub fn new(msg: impl Into<String>) -> Self {
+    pub(crate) fn new(msg: impl Into<String>) -> Self {
         Self(msg.into())
     }
 
@@ -233,17 +233,12 @@ impl<'de, 'a, 'f> MapAccess<'de> for RawEventMapAccess<'a, 'f> {
         }
 
         let synthetic = self.synthetic_offset();
-        let key: &str;
-        let pending: PendingValue<'a, 'f>;
-        match self.index {
-            0 => {
-                key = "event";
-                pending = PendingValue::Name(self.name);
-            }
-            1 if self.timestamp_ns.is_some() => {
-                key = "timestamp_ns";
-                pending = PendingValue::Timestamp(self.timestamp_ns.unwrap());
-            }
+        let (key, pending) = match self.index {
+            0 => ("event", PendingValue::Name(self.name)),
+            1 if self.timestamp_ns.is_some() => (
+                "timestamp_ns",
+                PendingValue::Timestamp(self.timestamp_ns.unwrap()),
+            ),
             i => {
                 let field_idx = i - synthetic;
                 let field_def = self.schema.fields().get(field_idx).ok_or_else(|| {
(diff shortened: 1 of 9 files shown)


dial9-trace-format/src/de.rs
```diff
@@ -0,0 +1,700 @@
...
+/// Error returned when deserializing a trace event into a typed value fails.
+///
+/// This wraps a message produced by serde or by the deserializer itself
+/// (e.g. for missing fields or pool lookup failures).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DeserError(String);
+
+impl DeserError {
+    /// Construct a new error with the given message.
+    pub fn new(msg: impl Into<String>) -> Self {
```
jlizen: Are we sure we want to expose the public constructor? Do we expect callers to use it?

dial9-trace-format/src/de.rs
```diff
@@ -0,0 +1,700 @@
...
+    }
+
+    fn next_value_seed<V: DeserializeSeed<'de>>(
+        &mut self,
+        seed: V,
+    ) -> Result<V::Value, Self::Error> {
+        let v = self
+            .pending_value
+            .take()
+            .expect("next_value without next_key");
```
jlizen: return a proper `DeserError` even if probably not reachable?

dial9-trace-format/src/de.rs
```diff
@@ -0,0 +1,700 @@
...
+
+    fn next_key_seed<K: DeserializeSeed<'de>>(
+        &mut self,
+        seed: K,
+    ) -> Result<Option<K::Value>, Self::Error> {
+        if self.index >= self.total_entries() {
+            return Ok(None);
+        }
+
+        let synthetic = self.synthetic_offset();
```
jlizen, line 235: you can declare this in the `i => {..}` match arm only

dial9-trace-format/tests/serde_deserialize.rs
```diff
@@ -0,0 +1,893 @@
...
+        Mixed {
+            timestamp_ns: 13_000,
+            pi: std::f64::consts::PI,
+            flag: true,
+            delta: -9_999,
+        }
+    );
+}
+
+// ── Test 11: Option<T> field entirely absent from wire schema ────────────────
```
jlizen, line 581: there are two 11s

dial9-trace-format/tests/serde_deserialize.rs
```diff
@@ -0,0 +1,893 @@
+//! Tests for the [`dial9_trace_format::de`] module — serde Deserializer for
```
jlizen, line 1: We should have a test covering a failed type coercion (eg requesting a `bool` from a wire `varint`)

dial9-trace-format/tests/serde_deserialize.rs
```diff
@@ -0,0 +1,893 @@
...
+//! `RawEvent`.
+//!
+//! These tests cover the full matrix of behaviors described in
+//! `prd-serde-deserializer.md`:
```
jlizen: does this file exist?

jlizen (approved):
Pretty conventional impl, seems good.

PR #451, opened 2026-05-26
feat(tokio-telemetry): add Deserialize to built-in event structs

rcoh:
Context: #446 

Adds decode-side companion structs for all built-in trace events in `telemetry::analysis_events`. These structs derive/implement `serde::Deserialize` so they can be used as decode targets with the serde-based trace iterator.

## Tests

- `builtin_event_serde.rs`: end-to-end test with a real TracedRuntime
- `synthetic_trace_serde.rs`: deterministic round-trip encoding all 14 event types with exact field assertions (no runtime needed)

### DIFF
--- a/dial9-tokio-telemetry/Cargo.toml
+++ b/dial9-tokio-telemetry/Cargo.toml
@@ -42,7 +42,7 @@ serde_json = "1"
 smallvec = "1"
 bytes = "1"
 dial9-perf-self-profile = { workspace = true, optional = true }
-dial9-trace-format = { workspace = true, features = ["serde"] }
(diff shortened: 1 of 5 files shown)


jv1i (approved):
Looks good! only thing that caught my attention but unsure if worth addressing is that using serde, events with schema mismatch (but known names) will error on decoding, whereas in the existing decoder they fall into `TelemetryEvent::Custom`. For serde, only mismatching name fall into `Other` (with no name/felds), known name + mismatching schema does not fall there (e.g. old trace prior to `tid` on `WorkerParkEvent`)

PR #453, opened 2026-05-26
Address misc UI usability issues (#450)

### LINKED ISSUE #450: misc UI usability issues
- [ ] Color coding poll lengths are confusing. Switch to a heatmap, possibly log scale?
- [ ] Move the legend for the number of spawned tasks to the bottom where that panel is (and generally rethink the legend)
- [ ] When you select the dropdown for "kernel scheduling delay", long polls, etc. it should automatically jump to the first one (instead of requiring you to click next)
- [ ] Open the flamegraph sidebar wider by default

rcoh:
Closes #450.

Four small fixes to the viewer's timeline UX, in roughly increasing order of size.

### 3. Auto-jump to first POI on dropdown / sort change
Changing the POI filter (`Kernel Scheduling Delays`, `Long Polls (>1ms)`, …) or the "Worst first" toggle now jumps the view to POI #1 of the new list, so users can immediately see what matched. Initial trace load still leaves the user at the overview.

### 4. Wider flamegraph sidebar by default
`#stack-sidebar` opens at `min(640px, 50vw)` instead of `400px`. The drag-resize handle still works (min 200, max 70vw unchanged).

### 2. Rework the legend
The top toolbar legend was mixing items that visualize on the worker lanes with one item (Active Tasks) that only appears in the queue-chart panel at the bottom. Now:

- The queue-chart panel has its own inline legend covering Global queue, Max local queue, and Active tasks.
- "Active Tasks" is removed from the top toolbar legend.
- "Local Q" stays in the top legend because it is also overlaid on each worker lane (yellow line).

### 1. Log-scale poll color heatmap
Replaced the four hard-coded duration buckets with a continuous, log-scale heatmap. The function lives in `trace_analysis.js` so it can be unit-tested in node:

| duration | color |
|----------|-------|
| ≤100ns   | dim navy `#2a5a7a` (floor) |
| 10µs     | cyan `#4fc3f7` |
| 100µs    | orange `#ff8a65` |
| 1ms      | bright red `#ff4444` |
| ≥1s      | pure red `#ff0000` (ceiling) |

Stops are linearly interpolated in RGB on the log10(duration) axis. The quantized variant (24 bins) is what `pollColor` actually calls, so adjacent polls of "approximately equal" duration still share a color string and the existing LOD merging in the worker-lane renderer keeps zoomed-out drawing fast.

The toolbar legend is now a gradient strip with tick labels at 100ns/10µs/1ms/1s aligned with the actual stops.

### Testing
- New unit tests in `dial9-viewer/ui/test_poll_color.js` (TDD: hex format, monotonic redness, floor/ceiling clamping, anchor points match legend swatches).
- All existing `dial9-viewer/ui/test_*.js` tests pass.
- `cargo build -p dial9-viewer` clean.

JS/HTML-only change — no `.rs` files touched, no trace format changes, so per `AGENTS.md` no `cargo nextest` / stress run required.

### DIFF
--- a/dial9-viewer/ui/test_poll_color.js
+++ b/dial9-viewer/ui/test_poll_color.js
@@ -1,0 +1,135 @@
+#!/usr/bin/env node
+"use strict";
+
+// Unit tests for pollHeatmapColor — a continuous, log-scale heatmap mapping
+// poll duration (in nanoseconds) to a hex color string.
+//
+// Run with: node test_poll_color.js
+// Used by viewer.html for issue #450 (item 1: poll color heatmap).
+
+const { pollHeatmapColor } = require("./trace_analysis.js");
+
+let failures = 0;
+function fail(msg) {
+  console.log(`✗ ${msg}`);
+  failures++;
+}
+function pass(msg) {
+  console.log(`✓ ${msg}`);
+}
+
+function isHex(s) {
+  return typeof s === "string" && /^#[0-9a-f]{6}$/i.test(s);
+}
+
+function hexToRgb(hex) {
+  return [
+    parseInt(hex.slice(1, 3), 16),
+    parseInt(hex.slice(3, 5), 16),
+    parseInt(hex.slice(5, 7), 16),
+  ];
+}
+
+// 1. Returns valid hex strings for any input duration
+function testHexFormat() {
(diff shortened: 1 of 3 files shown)


jv1i (approved):
> Log-scale poll color heatmap

Nice!

PR #454, opened 2026-05-26
feat: cpu profiler only mode

### LINKED ISSUE #351: using dial9 as only a profiler

noxware:
## Changes
- Adds a `.with_tokio_instrumentation(...)` method and a `DIAL9_TOKIO_INSTRUMENTATION_ENABLED` env var, that avoids installing tokio hooks (by not calling `attach_runtime`) when set to `false` (default `true`).
- Modify the web viewer to open the flamegraph if the trace file contains cpu profiling info only. 

## Other
Closes #351

### DIFF
--- a/dial9-tokio-telemetry/README.md
+++ b/dial9-tokio-telemetry/README.md
@@ -91,6 +91,7 @@ Runtime knobs:
 | Name | Default | Meaning |
 | --- | --- | --- |
 | `DIAL9_TASK_TRACKING_ENABLED` | `true` | Track tasks spawned through dial9 handles. |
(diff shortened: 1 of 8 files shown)


dial9-tokio-telemetry/README.md
```diff
@@ -214,6 +215,30 @@ Dial9Config::builder()
     // ...
 ```
 
+To use dial9 as a CPU profiler without installing Tokio runtime hooks, keep
```
rcoh, line 218: probably worth noting you can also get this via CoreTelemetryBuilder?

rcoh (approved):

PR #458, opened 2026-05-27
docs(README): add memory profiling section

rcoh:
#287

### DIFF
--- a/dial9-tokio-telemetry/README.md
+++ b/dial9-tokio-telemetry/README.md
@@ -132,6 +132,7 @@ dial9 is fundamentally a central buffer that can collect data from different sou
 
 - [Tokio Events](#tokio-events): dial9 can capture poll, wake, and worker events from Tokio
 - [CPU profiling](#cpu-profiling-linux-only): dial9 can capture linux performance counters and events to produce flamegraphs
+- [Memory profiling](#memory-profiling): dial9 can sample heap allocations to produce allocation flamegraphs and detect leaks
 - [Tracing spans](#tracing-span-events-opt-in): dial9 can capture tracing spans to bring tracing context into your trace files
 - [Task dumps](#task-dumps-linux-only): dial9 can capture a task dump (a backtrace when your future goes idle) to determine what it is waiting for when idle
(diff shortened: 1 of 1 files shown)


dial9-tokio-telemetry/README.md
```diff
@@ -229,6 +230,67 @@ Dial9Config::builder()
   sudo sysctl kernel.kptr_restrict=0
   ```
 
+### Memory profiling
+
+dial9 can sample heap allocations using Poisson sampling and capture stack traces for each sample. This produces allocation flamegraphs showing where memory is being allocated. With liveset tracking enabled, you can also detect memory leaks by seeing which allocations are never freed.
```
jlizen: would be nice to have a link to background info on poisson sampling

jlizen: Should we mention that the agent skills has a module focused on memory? (Or have we not built that yet?)

dial9-tokio-telemetry/README.md
```diff
@@ -229,6 +230,67 @@ Dial9Config::builder()
...
+# fn main() {}
+```
+
+The `sample_rate_bytes` controls how frequently allocations are sampled. At the default of 512 KiB, a service allocating 1 GB/s produces ~2000 samples/sec. Set to `1` to sample every allocation (useful for tests, not production).
+
+#### Liveset tracking and leak detection
+
+When `track_liveset(true)` is set, dial9 records every deallocation so it can determine which sampled allocations are still live at any point in the trace. This is how you find memory leaks: allocations that appear in the liveset and grow over time without being freed.
+
+> **Caveat:** At very high deallocation rates the free queue can overflow. When a free event is dropped, the corresponding allocation remains in the liveset even if it was actually freed — appearing as a false-positive "leak". If you see suspicious liveset growth in a high-throughput service, check the `dropped_frees` counter before concluding you have a real leak.
```
jlizen: where would they see the `dropped_free` counter?

jlizen (approved):

dial9-tokio-telemetry/README.md
rcoh: fixed, #459 emits events when this happens

PR #459, opened 2026-05-27
feat: emit MemoryProfileOverflowEvent on ring buffer overflow

rcoh:
Emit a trace event when the memory profiler's ring buffers drop samples during flush. JS parser/analysis surfaces overflow counts and warns about false leaks from dropped frees.

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/format.rs
+++ b/dial9-tokio-telemetry/src/telemetry/format.rs
@@ -1,6 +1,7 @@
 use crate::telemetry::events::CpuSampleSource;
 #[cfg(any(feature = "analysis", test))]
 use crate::telemetry::events::TelemetryEvent;
+
 use crate::telemetry::task_metadata::TaskId;
 #[cfg(any(feature = "analysis", test))]
 use dial9_trace_format::decoder::{StackPool, StringPool};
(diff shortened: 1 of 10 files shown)


dial9-viewer/skills/dial9-toolkit/scripts/analyze.js
```diff
@@ -217,6 +218,7 @@ function accumulateTrace(acc, trace) {
   // Memory events
   if (trace.allocEvents) for (const e of trace.allocEvents) acc.allocEvents.push(e);
   if (trace.freeEvents) for (const e of trace.freeEvents) acc.freeEvents.push(e);
+  if (trace.memoryOverflows) for (const e of trace.memoryOverflows) acc.memoryOverflows.push(e);
```
jv1i, line 222: very nit: perhaps name it `memoryOverflowEvents` to match the others?

dial9-tokio-telemetry/src/telemetry/format.rs
jv1i, line 414: Should we add an arm here for `MemoryProfileOverflowEvent`? (and also add this to `TelemetryEvent`)
or do we want this event just on the JS side?

If we add it, should we have a unit test in `source.rs` for the overflow path? e.g. preloading `rings.dropped_*` and asserting that a `MemoryProfileOverflowEvent` is emitted with the correct deltas across flushes

jv1i (approved):
Just leaving a question on the analysis side

dial9-tokio-telemetry/src/telemetry/format.rs
rcoh: This is deleted on 0.4 so I figured I wouldn't bother

PR #463, opened 2026-05-27
Add dial9.version

### LINKED ISSUE #423: Encode the exact version of dial9 into the segment metadata
We should encode the crates.io version of dial9 (e.g. 0.3.7) into the segment metadata

Fluzko:
closes #423 

## Summary
Adds `dial9.version` to the segment processor, taking the value from `CARGO_PKG_VERSION` env var

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/writer.rs
+++ b/dial9-tokio-telemetry/src/telemetry/writer.rs
@@ -86,10 +86,10 @@ impl<W: TraceWriter + ?Sized> TraceWriter for Box<W> {
     }
 }
 
-/// Reserved segment-metadata key carrying the crates.io version of
-/// `dial9-tokio-telemetry`. User-supplied entries with this key are silently
-/// dropped, so the producer version must reflect the actual binary.
-const DIAL9_VERSION_KEY: &str = "dial9.version";
+/// Segment-metadata key carrying the crates.io version of
+/// `dial9-tokio-telemetry`. Populated by default, any user-supplied entry with
+/// this key take precedence.
+const DIAL9_VERSION_KEY: &str = "dial9.dial9-tokio-telemetry.version";
 
 /// Compile-time value for `DIAL9_VERSION_KEY`.
 const DIAL9_VERSION_VALUE: &str = env!("CARGO_PKG_VERSION");
@@ -111,15 +111,11 @@ impl Default for SegmentMetadata {
 }
 
 impl SegmentMetadata {
-    /// Build segment metadata from user-supplied entries on top of the built-in
-    /// `dial9.version` key. User entries with the reserved key are dropped.
+    /// Build segment metadata from user-supplied entries on top of the default
+    /// `dial9.dial9-tokio-telemetry.version` key. User entries with the same key override the default.
     fn new(user_entries: Vec<(String, String)>) -> Self {
         let mut s = Self::default();
-        s.merge(
-            user_entries
-                .into_iter()
-                .filter(|(k, _)| k != DIAL9_VERSION_KEY),
-        );
+        s.merge(user_entries.into_iter());
         s
     }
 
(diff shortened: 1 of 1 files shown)


dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -86,14 +86,41 @@ impl<W: TraceWriter + ?Sized> TraceWriter for Box<W> {
...
 }
 
-#[derive(Default, Clone)]
+/// Reserved segment-metadata key carrying the crates.io version of
+/// `dial9-tokio-telemetry`. User-supplied entries with this key are silently
+/// dropped, so the producer version must reflect the actual binary.
+const DIAL9_VERSION_KEY: &str = "dial9.version";
+
+/// Compile-time value for `DIAL9_VERSION_KEY`.
+const DIAL9_VERSION_VALUE: &str = env!("CARGO_PKG_VERSION");
```
Fluzko: I've decided to only add this one, but we could potentially also add `dial9_cli_binary` and `dial9_trace_format`. They're not related at all, but if we want full traceability, we could add them.

dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -86,14 +86,41 @@ impl<W: TraceWriter + ?Sized> TraceWriter for Box<W> {
...
+    /// Build segment metadata from user-supplied entries on top of the built-in
+    /// `dial9.version` key. User entries with the reserved key are dropped.
+    fn new(user_entries: Vec<(String, String)>) -> Self {
+        let mut s = Self::default();
+        s.merge(
+            user_entries
+                .into_iter()
+                .filter(|(k, _)| k != DIAL9_VERSION_KEY),
+        );
+        s
```
rcoh: seems unnecessary, if someone overrides it on purpose it seems worth honoring

dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -2279,6 +2332,80 @@ mod tests {
...
+                env!("CARGO_PKG_VERSION").to_string()
+            )),
+            "expected dial9.version entry matching CARGO_PKG_VERSION"
+        );
+    }
+
+    /// The built-in `dial9.version` entry is authoritative — attempts to
+    /// override it via the builder or `update_segment_metadata` are dropped.
+    #[test]
+    fn test_dial9_version_cannot_be_overridden() {
```
rcoh: if you override it on purpose you probably have a good reason...if we wanted to do this I would define a private segment metadata namespace. seems unnecessary

dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -86,14 +86,41 @@ impl<W: TraceWriter + ?Sized> TraceWriter for Box<W> {
     }
 }
 
-#[derive(Default, Clone)]
+/// Reserved segment-metadata key carrying the crates.io version of
+/// `dial9-tokio-telemetry`. User-supplied entries with this key are silently
+/// dropped, so the producer version must reflect the actual binary.
+const DIAL9_VERSION_KEY: &str = "dial9.version";
```
rcoh: this should probably be longer so we can put other stuff here as well, e.g. in 0.4 we may want the core version, the tokio telemetry version, etc.

`dial9.dial9-tokio-telemetry.version` e.g.

then the UI can show dial9 specific metadata since we know everything under the dial9 prefix is ours

dial9-tokio-telemetry/src/telemetry/writer.rs
Fluzko: Can you go further? There shouldn't be any reason for someone to change it, but only an error. I don't think giving users a knob to _change_ it would solve it either. It's some internal metadata that we add for debugging or related purposes. I would instead agree on a knob that disables this kind of metadata (say, we add more in the future). 
I would like to know more about why you think this should be allowed to be managed.

rcoh: more that we just shouldn't both with special case code to specifically _block_ setting it, its just not really worth it

dial9-tokio-telemetry/src/telemetry/writer.rs
Fluzko: Makes sense, didn't find any better alternative than yours, going with `dial9.dial9-tokio-telemetry.version`

rcoh (approved):

PR #465, opened 2026-05-28
Cache the symbolizer across segments (#462)

### LINKED ISSUE #462: cache the symbolizer in the pipeline
right now, the symbolizer is recreated for all segements

rcoh:
Closes #462.

## Problem

The pipeline's `SymbolizeProcessor` constructed a fresh `blazesym::Symbolizer` for every segment, paying the full ELF/DWARF parse cost (~600 ms) on every flush. `blazesym::Symbolizer` is `!Send` and `!Sync`, so it can't share the tokio blocking pool.

## Fix

Introduce `dial9_perf_self_profile::offline_symbolize::OfflineSymbolizer`: a long-lived symbolizer that owns one `blazesym::Symbolizer` on a dedicated `std::thread`. Callers submit work over an mpsc channel and block for the response; blazesym's per-ELF DWARF cache stays warm for the lifetime of the worker.

`SymbolizeProcessor` now holds an \`Arc<OfflineSymbolizer>\` and forwards calls to it.

The thread is spawned **lazily on first \`symbolize()\` call** so it descends from \`dial9-worker\` (which has a tracked perf fd) rather than from the constructing thread; this keeps the symbolizer thread visible to dial9's own CPU profiler.

Per-request panics inside the symbolize thread are caught (with rate-limited error logging), so a transient blazesym error on one segment doesn't destroy the cache.

The legacy \`symbolize_trace_with_maps\` keeps its signature and constructs a one-shot \`Symbolizer\` internally — no public API break.

## Measurement

Reproduced on the \`metrics-service\` example with 2s rotation and 20s run, observed via per-segment \`Symbolize.Time\`:

| Segment | Before (main) | After (default config) | After (no memory profiling) |
|--------:|--------------:|-----------------------:|-----------------------------:|
| 0 | 632 ms | 595 ms | **385 ms** |
| 1 | 657 ms | 230 ms | **102 ms** (6.4× vs main) |
| 2 | 631 ms | 244 ms | **97 ms** (6.5× vs main) |
| 3 | 669 ms | 239 ms | **96 ms** |
| 4–8 | ~600 ms | 149–230 ms | 78–94 ms |
| 9 | 463 ms | 114 ms | 81 ms |

Steady-state went from ~600 ms to ~80–100 ms when the cache is hot.

## Why two columns?

After the cache fix, I flamegraphed the symbolize thread by mining CPU samples whose stacks pass through \`offline_symbolize\` from the trace itself. With memory profiling enabled, **about half of the symbolize thread's CPU was \`Dial9Allocator\` ring-buffer bookkeeping** (\`ring::RingBuffers::push_free\`) — the global allocator was hooking every alloc/free that blazesym, the trace decoder, and the proc-maps parser made.

That's a separate concern from this PR — it's the cost of opting into memory profiling, not a symbolize bug. Worth tracking as a follow-up: the symbolize thread doesn't need to be in the memory-profile sample path, since its allocations are tooling, not application work.

With memory profiling off, the steady-state symbolize-thread breakdown is:

- 37% \`Decoder::for_each_event\` — parsing trace bytes to collect addresses
- ~19% address dedup (HashMap insert + hash)
- 4% \`/proc/self/maps\` line parsing
- ~10% actual blazesym (DWARF, ELF, demangle, ksym)

So with the cache warm, blazesym is only ~10% of the residual cost — the rest is decode + bookkeeping. Future cleanups for another PR.

## Test strategy

Counter-based, parallelism-safe, no timing assertions:

\`\`\`rust
let symbolizer = OfflineSymbolizer::new();
assert_eq!(symbolizer.symbolizer_constructions(), 0); // lazy
for seg in &three_real_segments {
    symbolizer.symbolize(seg, &maps).unwrap();
}
assert_eq!(symbolizer.symbolizer_constructions(), 1); // exactly one Symbolizer
\`\`\`

If a future change re-introduces per-segment \`Symbolizer::new()\`, this test fails immediately rather than just regressing performance silently.

## Tradeoffs

- One extra \`std::thread\` per \`SymbolizeProcessor\` (named \`dial9-symbolizer\`). One processor per worker. Spawned lazily, so unused configs never pay for it.
- Per-request copy of \`input\` and \`maps\` to send across the channel. A few µs vs hundreds of ms of work.
- Maps are still re-read per segment (we want to pick up \`dlopen\`'d libraries). blazesym's per-ELF cache stays warm regardless.
- Stale ELF cache: if a path is unloaded and a different binary is loaded at the same path, blazesym may serve stale symbols. Acceptable for production use.

## Side changes

- \`examples/metrics-service\` accepts \`--worker-metrics-path PATH\` so dial9 worker pipeline metrics are observable. Without this the \`Symbolize.Time\` metric had nowhere to go.
- \`resolve_kernel_symbol_returns_name\` skips cleanly when \`kptr_restrict\` masks \`/proc/kallsyms\` instead of failing with a misleading assertion.

## Tested

- \`cargo fmt --check\` clean
- \`cargo clippy --all-targets --all-features\` clean
- \`cargo nextest run -p dial9-perf-self-profile --stress-duration 20s\`: 34 iterations × 83 tests
- \`cargo nextest run -p dial9-tokio-telemetry --features cpu-profiling,worker-s3 --lib --stress-duration 20s\`: 10 iterations × 278 tests
- 89 \`dial9-tokio-telemetry\` integration tests including \`background_symbolization\`
- 4 shuttle tests
- 213 \`dial9-trace-format\` / \`dial9-viewer\` / \`dial9\` tests


perf-self-profile/src/offline_symbolize.rs
```diff
@@ -94,6 +102,397 @@ fn collect_stack_frame_addresses(
...
+    fn worker_loop(
+        rx: mpsc::Receiver<Request>,
+        constructions: Arc<AtomicU64>,
+        state_constructions: Arc<AtomicU64>,
+    ) {
+        // Lazy: only construct the Symbolizer when we receive the first
+        // request, so the cost is paid on first segment, not on creation.
+        let mut symbolizer: Option<blazesym::symbolize::Symbolizer> = None;
+        let mut state = SymbolizeState::new();
+        state_constructions.fetch_add(1, Ordering::Relaxed);
```
jv1i, line 390: Should we call `register_current_thread()` at the top of `worker_loop` to also cover the ctimer fallback (no inherit there)?

jv1i (approved):
LGTM. Leaving a comment about cpu profiling this without perf

Issue #455, opened 2026-05-27
add source for system metrics


rcoh:
- [x] getrusage (rss etc.)
-  [x] add the ability to easily add custom metrics via a callback

PR #470, opened 2026-05-29
feat: Process resource usage source

noxware:
## Changes
- Implements a new `ProcessResourceUsageSource` based on the `getrusage` Unix system call.
- Exposed at the public API as a method in `TracedRuntimeBuilder`, `TelemetryCoreBuilder` and an env var in `from_env`.

## Other
Part of #455


dial9-tokio-telemetry/src/config.rs
```diff
@@ -216,6 +219,7 @@ const DEFAULT_CPU_PROFILE_ENABLED: bool = cfg!(all(target_os = "linux", feature
 const DEFAULT_SCHEDULE_PROFILE_ENABLED: bool =
     cfg!(all(target_os = "linux", feature = "cpu-profiling"));
 const DEFAULT_TASK_DUMP_ENABLED: bool = false;
+const DEFAULT_PROCESS_RESOURCE_USAGE_ENABLED: bool = cfg!(unix);
```
noxware, line 222: This mimics how we handle the default of "cpu profile" and "schedule profile", which are not enabled by default in the Rust builder API, but they are enabled by default based on the platform capabilities at `from_env`.

dial9-tokio-telemetry/src/telemetry/process_resource_usage.rs
```diff
@@ -0,0 +1,312 @@
+//! Process resource usage sampled from the operating system.
+
+use std::time::Duration;
+
+const DEFAULT_SAMPLE_INTERVAL: Duration = Duration::from_secs(1);
```
noxware: is 1 sec an apropiarte default?

dial9-tokio-telemetry/src/telemetry/format.rs
```diff
@@ -173,6 +173,33 @@ pub(crate) struct QueueSampleEvent {
     pub global_queue: u8,
 }
 
+/// Wire-format event for process resource usage sampled from `getrusage(RUSAGE_SELF)`.
+#[derive(Debug, TraceEvent)]
+#[cfg_attr(not(feature = "unstable-events"), non_exhaustive)]
+pub struct ProcessResourceUsageEvent {
+    /// Wall-clock timestamp in nanoseconds (monotonic).
+    #[traceevent(timestamp)]
+    pub timestamp_ns: u64,
```
jv1i: doc confusing: wall-clock (realtime) != monotonic (arbitrary start point), in this case the value comes from `clock_monotonic_ns()` right? so it is just monotonic.

dial9-tokio-telemetry/src/telemetry/process_resource_usage.rs
```diff
@@ -0,0 +1,312 @@
...
+
+            match read_process_resource_usage() {
+                Ok(snapshot) => {
+                    let event = snapshot.into_event(clock_monotonic_ns());
+                    record_encodable_event(&event, ctx.collector, ctx.drain_epoch);
+                }
+                Err(e) => rate_limited!(Duration::from_secs(60), {
+                    tracing::warn!("failed to read process resource usage via getrusage: {e}");
+                }),
+            }
```
jv1i, line 95: nit: could move `self.last_sample = Some(now)` to advance only on `Ok` so that on errors sampling doesn't get paused until the next full interval. though I guess this wouldn't error in practice.

jv1i (approved):
LGTM!

dial9-tokio-telemetry/src/telemetry/process_resource_usage.rs
rcoh: Depends on the cost of getting the stats. If it is cheap more frequent would be nice

rcoh (commented):
Did you re record the demo trace and validate data sanity?

dial9-tokio-telemetry/src/telemetry/format.rs
noxware: Makes sense, I will do the change here but note that wording is all over the place in `events.rs` :P

dial9-tokio-telemetry/src/telemetry/process_resource_usage.rs
noxware: Researched and experimented a bit. It's extremely fast. It's still a syscall but... It seems to simply return the values of some in-memory system counters so... no heavy blocking code, scheduler, I/O or anything similar at sight. So instead of being conservative, I reduced the interval to just `100ms`. WDYT?

noxware:
Regenerated and committed the demo trace, thanks 🤙 To my understanding, it looks good if that's what you mean by "validate data sanity" @rcoh .

Screenshot 2026-05-29 at 5 06 41 PM

dial9-tokio-telemetry/src/telemetry/process_resource_usage.rs
noxware: I think skipping to the next full interval is a good thing. It acts as a "natural backoff" mechanism in case something goes wrong. Like... "if something goes wrong, retry in `<INTERVAL>`, instead of immediately in the next flush".

And yeah, in practice this should probably not fail :P the 2 bits of unsafe code are very simple and I'm using the most common configuration for this syscall (according to my exploration). So, unless something very terrible is happening, this should not fail 🙏 

Note: I reduced the interval value, so by the default configuration, this "retry" will come faster. Tiny semi-related detail.

rcoh (approved):

PR #485, opened 2026-06-02
feat!: replace `TelemetryEvent` and Rust decode side with new serde-based approach

rcoh:
… to serde decode

Squashed from the megamerge-serde-series branch, minus the owned event iterator (which can land separately).

Changes:
- TraceEvent trait trimmed to encode-only (Ref/decode removed)
- TraceField trait trimmed to encode-only (Ref/decode_ref removed)
- All integration tests migrated from CapturingWriter+TelemetryEvent pattern matching to BytesCapturingWriter+serde Deserialize structs
- Examples migrated to use serde decode path
- SegmentMetadata.entries changed from Vec<(String, String)> to HashMap
- test_parser.js event name mappings updated to Event suffix
- validation.rs removed (tests now validate inline)


dial9-tokio-telemetry/src/telemetry/analysis.rs
```diff
@@ -37,104 +33,76 @@ impl TraceReader {
...
-                        (name.to_string(), owned)
-                    })
-                    .collect();
-                events.push(TelemetryEvent::Custom {
-                    timestamp_nanos: ev.timestamp_ns,
-                    name: ev.name.to_string(),
-                    fields,
-                });
+                Dial9Event::Other | Dial9Event::ProcessResourceUsageEvent(_) => {
+                    // Unknown event: deserialize as CustomEvent to get fields.
```
jv1i, line 69: Is there a reason for `ProcessResourceUsageEvent` to be handled like `Other`? or should it fall into `_ => {}`?

jv1i (approved):
LGTM

Issue #390, opened 2026-05-11
Allow setting the wire id directly in TraceEvent


rcoh:
Right now, the WireId is always assigned at runtime, which means we do a hashmap lookup on every write.

There are a few ways to work around around this. I think most of them involve storing the epoch for the writer in TraceEvent so you know if this has been written yet.

But in any case, it would be reasonable to start the dynamic ids at 256 and allow 0-256 to be statically assigned to trace events that we control.

jv1i:
Completed in https://github.com/dial9-rs/dial9/pull/487

PR #487, opened 2026-06-02
perf: Improve Wire ID Resolution

jv1i:
#390

## Overview

`Encoder::write::<T>` did an `FxHashMap::get` per event to resolve the wire id. 

Two mechanisms cut that out, with the first one being the most important:

1. Per-encoder slot cache: each `#[derive(TraceEvent)]` type gets a lazily assigned dense slot. The encoder keeps a `Vec<u32>` keyed by slot, so the hot path is a bounds-checked array load instead of a hashmap lookup. Per-encoder, so per-thread `ThreadLocalBuffer` encoders don't contend. 
This change by itself gives a ~ -27% instruction count.

2. Static wire ids: `#[traceevent(wire_id = WireId::X)]` claims a fixed id in `0..256`. 
Annotated events skip the slot load and use the const id directly, dynamic ids start at 256. (all `dial9-tokio-telemetry` events are annotated)
This change provides another 4.5% improvement on top of the previous one for direct/writer and ~10% on rawcopy.

Bench: iai-callgrind instructions, 1M events / batches_100.

| bench | main | slot | + static |
|---|---:|---:|---:|
| codec encode | 427.2M | 356.2M | **349.2M (-18%)** |
| threadlocal direct | 25.3M | 18.6M | **17.8M (-30%)** |
| threadlocal rawcopy | 28.2M | 21.9M | **19.7M (-30%)** |
| writer encode | 25.6M | 18.8M | **18.0M (-30%)** |

### Static id uniqueness

Ids come from one `WireId` enum with explicit discriminants. A duplicate is a compile error, so the enum is the single source of truth. 
I thought of using the distributed slice from `linkme` to make this less manual while still having compile-time errors on duplicates, but that wouldn't work on some platforms like Android.

The derive appends `as u16` and asserts `id < STATIC_WIRE_ID_LIMIT` at compile time. Because uniqueness is compile-guaranteed, the encoder only needs a 32-byte`[u64; 4]` registered bitset, no per-id owner tracking.


jv1i:
Just realized I should've opened the PR from an upstream branch to get the bencher results here, but I shared the results of my local runs.

dial9-tokio-telemetry/src/telemetry/format.rs
```diff
@@ -104,8 +104,35 @@ impl TraceField for WorkerId {
...
+    Free = 15,
+    MemoryProfileOverflow = 16,
+    // Only referenced when the span-event layer is compiled.
+    #[cfg(feature = "tracing-layer")]
+    SpanClose = 20,
+}
+
 /// Wire-format event for a task poll start.
 #[derive(Debug, TraceEvent)]
+#[traceevent(wire_id = WireId::PollStart)]
```
rcoh: these don't actually need to be static, just static for the program. you could have a global counter that is incremented when a new trace event type is first initialized

rcoh (commented):
I believe we can merge the slot & setting the wireid.

You can do something like generate a lazy_static in the TraceEvent macro that does something like:

```
enum WireId { StaticId, Dynamic }
fn static_wire_id () {
  static id: Option<WireId>;
  if id.is_none() { let next_id = global_id.fetch_add(1) + 1;
  if id < MAX_STATIC_ID { id = Some(Some(id) 
}
```

obviously a very rough sketch but I think you get the idea

jv1i:
@rcoh updated with your idea, much cleaner to implement, only slightly worse than the fully static approach so I think it's worth the merge

| bench | main | merged (current) | const+slot (prev) |
|---|---:|---:|---:|
| codec encode | 427.2M | 358.2M (-16%) | 347.2M (-19%) |
| threadlocal direct | 25.3M | 18.8M (-26%) | 17.4M (-31%) |
| threadlocal rawcopy | 28.2M | 20.8M (-26%) | 19.4M (-31%) |
| writer encode | 25.6M | 19.0M (-26%) | 17.6M (-31%) |

dial9-trace-format-derive/src/lib.rs
```diff
@@ -139,6 +139,25 @@ fn derive_trace_event_impl(input: DeriveInput) -> proc_macro2::TokenStream {
             type Ref<'a> = #ref_name<'a>;
 
             fn event_name() -> &'static str { #name_str }
```
rcoh: Wdyt about making this opt in in the macro?

rcoh (approved):

Issue #483, opened 2026-06-01
dial9-trace-analysis skill: teach agents to read actual stack frames, not just spawn locations


rcoh:
## Background

When evaluating an agent built with the `dial9-html-report` skill against a baseline (no skill), we noticed both runs identified the same finding for the metrics-service demo trace:

> **Critical: Blocking file I/O on async worker at startup** — attributed to AWS config loading at `examples/metrics-service/src/main.rs:260:14`

But the actual stack frames in the off-CPU samples for that poll show this is **not** AWS config loading — it's `rustls_native_certs::load_native_certs` reading PEM cert files:

```
unix::FileDesc::read_buf unix.rs:178
pal::unix::cvt mod.rs:245
BufReader::fill_buf bufreader.rs:454
buffer::Buffer::fill_buf buffer.rs:148
borrowed_buf::BorrowedBuf::len borrowed_buf.rs:99
rustls_pki_types::pem::from_buf_inner pem.rs:248
rustls_pki_types::pem::read_until_newline pem.rs:525
rustls_native_certs::load_pem_certs lib.rs:315
ReadIter::next pem.rs:128
rustls_native_certs::load_pem_certs_from_dir lib.rs:303
[…]
rustls_native_certs::load_native_certs lib.rs:125
[…]
rustls_provider::build_connector::create_rustls_client_config rustls_provider.rs:141
rustls_provider::CryptoMode::provider rustls_provider.rs:27
rustls_provider::build_connector::wrap_connector rustls_provider.rs:155
http::HttpConnector::enforce_http http.rs:264
```

So the actual root cause is the rustls native-cert load — a one-time TLS setup cost that happens to occur inside the AWS-config-loading task. Two agents in a row missed this, both pattern-matching on the spawn location instead of reading the off-CPU stack.

## Why this is a skill problem

The `dial9-trace-analysis` skill (or whichever skill teaches agents to interpret `analyze.js` output) should explicitly instruct agents to:

1. **Always read at least one full off-CPU stack** for any long-poll finding before naming a root cause.
2. **Distinguish "where the future was spawned" from "what the future was doing."** The spawn location is where you started the task; the off-CPU stack is what blocked it.
3. **Recognize common library-level signatures** — e.g., `rustls_native_certs::load_*`, `aws_config::*`, `dns_lookup`, etc. — so agents can name them precisely.

[...]

PR #488, opened 2026-06-02
Add dial9-html-report skill: agent-authored HTML trace reports

rcoh:
## What

Adds a new `dial9-html-report` skill that teaches agents to compile their trace insights into a polished HTML report folder, with embedded flamegraphs and deep-links to the full viewer.

The skill is **primarily a teaching document** (`SKILL.md`), supported by small enabling pieces:

- **`dial9-trace-format/js/slice.js`** — a pure-JS trace slicer. Filters by time range, preserves symbol/metadata/clock-sync events so flamegraphs in slices still symbolize. Supports `--relative` (offsets from `minTs`, the natural mental model) and absolute monotonic ns. CLI + JS API.
- **`dial9 report serve <dir>`** — thin static-file server (tower-http `ServeDir`) so reports work in browsers without `file://` `fetch` issues. Same server lets the report's deep-links resolve `viewer.html` from the report folder.
- **`dial9-html-report/SKILL.md`** — the teaching content. Folder structure, copyable HTML skeleton, **flamegraph recipes** for precise sample filtering (per-task, per-poll-duration, off-CPU only, leaf-frame search, multi-trace union), viewer deep-link guidance with absolute-ns timestamps, and a complete worked example.

## Why

Without the skill, agents asked to "produce an HTML report from a trace" do a credible analysis but ship reports that:

- Fake flamegraphs with hand-built CSS bars (no real call-tree data).
- Invent viewer URL params like `?worker=` and `?task=` (don't exist).
- Copy multi-MB trace files verbatim into the report folder.
- Pattern-match findings on spawn locations rather than reading actual stack frames — e.g., attributing a blocking poll to "AWS config loading" when the off-CPU stack clearly shows `rustls_native_certs::load_pem_certs`.

Validated against the demo trace by running the same prompt against a baseline (no skill) and a skill-equipped agent. With the skill, the agent produces precisely-scoped flamegraphs (verified: 142/144 off-CPU samples in finding 1's flamegraph contain `rustls_native_certs::load_pem_certs`; 0 contain `hashmap::insert`), correct viewer deep-links with absolute monotonic ns, and minimal ceremony (5 file reads vs 0).

## What's tested

- `dial9-viewer/ui/test_slice.js` — 14 tests (round-trip, time-range correctness, gzip, `--relative` mode, symbol/metadata preservation).
- `dial9-viewer/ui/test_flamegraph_recipes.js` — 9 tests covering each recipe predicate and a type-contract check (timestamps must be Numbers, not BigInts — caught a real foot-gun).
- `dial9-viewer/tests/report_serve_test.rs` — 4 tests for the report-serve subcommand (HTML, binary, root-index, 404).
- `cargo nextest run -p dial9-viewer` 32/32 ✓
- `cargo fmt --check` ✓
- `cargo clippy -p dial9-viewer --all-targets --all-features` ✓

## Commits

1. Add JS trace slicer in `dial9-trace-format/js/slice.js`
2. Add embeddable viewer component in `dial9-viewer/ui/embed.html` (later removed)
3. Add `dial9-html-report` skill
4. Fix `slice.js` timestamp foot-gun: add `--relative` flag
5. Add `dial9 report serve <dir>` subcommand for HTML reports
6. Fix `slice.js` to preserve symbol table and metadata events
7. Teach agents to produce precisely-filtered flamegraphs
8. Remove `embed.html`, link reports directly to `viewer.html`
9. Fix BigInt foot-gun in skill: trace timestamps are Numbers

## Follow-ups (filed / planned)

- **Issue #483**: `dial9-trace-analysis` skill should explicitly teach agents to read off-CPU stack frames, not pattern-match spawn locations. (Linked above.)
- **Pool pruning in `slice.js`** — drop unreferenced strings/stacks/symbols so a 150ms slice goes from 2.4MB → ~200KB. Enables painless single-file inline reports.
- **Single-file inline-base64 report mode** — once slices are small enough, agents can inline the trace and viewer into one `report.html` for trivial sharing. Will require small `viewer.html` patch to read `window.dial9TraceData`.

## How to verify locally

```bash
git checkout pr1-html-report-skill
cargo build -p dial9-viewer

# Read the new skill
cargo run -p dial9-viewer --quiet -- agents skill dial9-html-report

# Build a report (use whatever agent you prefer; prompt them to build an HTML
# report from dial9-viewer/ui/demo-trace.bin and read the dial9-html-report skill)

# View it
cargo run -p dial9-viewer -- report serve path/to/report-folder
# → http://localhost:8000/report.html
```


dial9-viewer/ui/test_slice.js
```diff
@@ -0,0 +1,287 @@
...
+  }
+
+  let failures = 0;
+  function fail(msg) {
+    console.log(`✗ ${msg}`);
+    failures++;
+  }
+  function pass(msg) {
+    console.log(`✓ ${msg}`);
+  }
```
Fluzko: Pretty much a NIT, but we're declaring the `fail` and `pass` on every test file. What do you think about creating a test harness file with these abstractions and standarize it. I like the approach we've got at `test_panel_layout.js`

```js
function test(name, fn) {
  try {
    fn();
    console.log(`✓ ${name}`);
    passed++;
  } catch (e) {
    console.error(`✗ ${name}: ${e.message}`);
    failed++;
  }
}
// ...
test("1 is equal to 1", () => { assert.strictEqual(1, 1); }) // use node asserts
```

dial9-viewer/ui/test_slice.js
```diff
@@ -0,0 +1,287 @@
...
+      pass("Gzipped input decompressed and sliced to raw output");
+    } else {
+      fail(`Unexpected output header: 0x${sliced[0].toString(16)} 0x${sliced[1].toString(16)}`);
+    }
+  }
+
+  // ── Test 6: Relative slice with [0, durationFull] returns all events ──
+  // Note: relative mode anchors from the first event's timestamp (findMinTs),
+  // which may differ slightly from parseTrace's minTs (global minimum).
+  // Use a generous end to ensure we capture everything.
```
Fluzko: you say "first event's absolute timestamp" but it's actually the first timestamped event in stream order, not the global minimum. This differs from parseTrace().minTs

dial9-trace-format/js/slice.js
```diff
@@ -0,0 +1,489 @@
...
+  let outputTimestampBaseNs = 0n;
+
+  function copyBytes(start, end) {
+    out.push(buf.slice(start, end));
+  }
+
+  // Validate and copy header
+  if (buf.length < 5) throw new Error("Trace too short for header");
+  for (let i = 0; i < 4; i++) {
+    if (buf[i] !== MAGIC[i]) throw new Error("Invalid trace magic");
```
Fluzko, line 79: Shall we validate the vesion, which comes after the MAGIC, and avoid invalid ones?

dial9-trace-format/js/slice.js
```diff
@@ -0,0 +1,489 @@
...
+  # Absolute (event.ts values from parseTrace — typically 10-15 digit numbers):
+  node slice.js --input full.bin --output burst.bin --start 150439548276 --end 154676563153
+
+  # Relative (offsets from trace start — typically 9-10 digit numbers):
+  node slice.js --input full.bin --output burst.bin --relative --start 3900000000 --end 4050000000`);
+    process.exit(1);
+  }
+  const input = fs.readFileSync(inputPath);
+  const opts = {};
+  if (startNs != null || endNs != null) {
```
Fluzko, line 481: NIT `startNs > endNs`

Fluzko (approved):

PR #489, opened 2026-06-02
feat: callback as custom metrics source

noxware:
## Changes
- Exposes a minimal `.with_custom_events(...)` function to configure a callback that runs during flush cycles. The callback is stored as a `CustomEventsSource` internally.

## Notes
- Although many sources drain information from a producer, simpler ones may poll information immediately, for example `ProcessResourceUsageSource`. The callback supports setting a `minimum_interval` to throttle polling-style callbacks through `CustomEventsConfig`.
- `CustomEventsConfig` is not strictly needed if users throttle manually, but comparing timestamps by hand is error-prone and the config gives us room for small scheduling options later.
- An alternative would be `CustomEvents::builder().on_flush(|ctx| { ... }).some_unknown_config(...).build()` and using `with_custom_events` as a setter for that, but that is less ergonomic.
- I didn't regenerate the demo trace because this API does not produce a specific built-in event type.

## Other
Part of #455


rcoh:
Should this be called "with_custom_events"?

noxware:
Yes, `with_custom_events` is probably better than `with_custom_metrics`. I will do the change since I can't think a better name, but there is also the "global" `record_event` method for dispatching custom events. I hope this doesn't cause a confusion.

dial9-tokio-telemetry/src/telemetry/recorder/guard.rs
```diff
@@ -289,6 +291,28 @@ impl<'a> TraceRuntimeCoreBuilder<'a> {
         self
     }
 
+    /// Register a custom event callback.
+    ///
+    /// The callback runs during flush cycles while telemetry is enabled.
```
rcoh, line 296: docs should probably specify that it can be called multiple times

rcoh (approved):
lgtm, small nit

rcoh (approved):

noxware:
@rcoh nit addressed at the last commit. I will let you press the beautiful green button to keep you in control 🙂

PR #490, opened 2026-06-03
feat(config)!: in-memory support for dial9 main macro

jv1i:
# Overview

Follow-up to #435 

This adds an in-memory writer mode to `Dial9Config`. In #435 we landed in-memory support but the macro setup could only write traces to disk. 

The builder starts by picking a writer mode. `base_path(...)` selects disk and `in_memory()` selects memory. Each one hands back a builder that only carries the knobs that mode supports, so you cannot mix disk and memory settings by accident. That mismatch is a compile error now instead of something caught at runtime.

## Using it

Disk works the same as before, you just enter through `base_path`:

```rust
Dial9Config::builder()
    .base_path("trace.bin")
    .max_file_size(64 << 20)
    .max_total_size(1 << 30)
    .build()?
```

In-memory keeps segments in process memory, so you attach a delivery processor through `with_runtime` to ship them somewhere (S3, a custom uploader, and so on):

```rust
Dial9Config::builder()
    .in_memory()
    .max_total_size(16 << 20)
    .with_runtime(|r| r.with_custom_pipeline(|p| p.pipe(my_uploader)))
    .build()?
```

Pairing `write_back()` with the in-memory writer is a compile error, same as on the non-macro builder.

## Breaking changes

You now choose a writer mode up front, with `base_path(path)` for disk or `in_memory()` for memory _(could rename `base_path` to something like `on_disk` if we want more symetry)_.
`base_path` used to be a setter you could call anywhere in the chain. It is now the call that enters disk mode, so it has to come first. A setup that already calls `base_path` first keeps working unchanged, the other setters still chain after it in any order:

```rust
Dial9Config::builder().base_path(p).max_file_size(m).max_total_size(t).build()?
```

Two things change for the disabled path. `enabled` is now a setter on the chosen mode, so it comes after `base_path`, and turning telemetry off needs a mode picked first. Old code like this:

```rust
Dial9Config::builder().enabled(false).build()
```

becomes this:

```rust
Dial9Config::builder().base_path("trace.bin").enabled(false).build()?
```

The path is ignored when disabled, it is only there because you enter through a mode. `build()` on a disabled config now returns a `Result` like the enabled path, so it needs a `?` or an `expect` where the old infallible form did not.

## Rejected alternatives

- Adding memory to the existing flat builder and catching a disk-plus-memory mix at build time, works but only reports misuse at runtime.
- Letting the macro accept a fully built runtime through a `from_parts` helper, works but that makes you hand-assemble the whole thing and then wrap it, which is more work than not using the macro at all. The mode selector keeps the macro worth using and makes the bad combinations impossible to write.

## Follow-ups

`Dial9Config::from_env` still only builds a disk writer. In-memory from env is left for a follow-up: it could work with the S3 preset (`DIAL9_S3_BUCKET`), since that delivery path is expressible through env vars, and would need a writer selector like `DIAL9_WRITER=memory`, not sure you could express more custom pipelines with env vars.


dial9-tokio-telemetry/src/config.rs
```diff
@@ -732,152 +763,168 @@ impl Dial9Config {
...
+    /// `rotation_period` are optional.
+    pub fn base_path(self, base_path: impl Into<PathBuf>) -> DiskConfigBuilder {
+        Dial9Config::disk(base_path.into())
+    }
 
-        #[builder(field)] runtime_finalizer: Option<RuntimeConfigurator>,
+    /// Keep trace segments in process memory, with no filesystem usage.
+    ///
+    /// `max_total_size` is required.
+    pub fn in_memory(self) -> MemoryConfigBuilder {
```
rcoh: maybe `in_memory_buffer`, then we can name the other one `on_disk_buffer`?

rcoh (commented):
overall seems good. I'd like to think about how we can have as few different ways to configure dial9 as possible in 0.4.

right now we have `TracedRuntimeBuilder`, `TelemetryCore`, `Dial9Config` (and maybe others??)

its worth trying to get as close to 1 as possible

dial9-tokio-telemetry/src/config.rs
jv1i: Good idea, I was suggesting `in_memory()` and `on_disk()` above but perhaps adding `_buffer` is more clear (updated now)

Definitely agree that we have too much configuration surface and duplicated logic for setting up dial9, e.g. now we'll have `on_disk_buffer` here, but different naming on the non-macro config. If we're ok merging this PR I can separately do some prototyping on a general overhaul of the config in a way that is unified.

rcoh (approved):

Issue #469, opened 2026-05-29
add ability to trigger a dump on an application trigger 


rcoh:
If you are running dial9, it produces large quantities of data. But ideally, you could trigger traces to be uploaded only in cases of unexpected application events (e.g. Tokio idle ratio < .5 etc.).

Since dial9 already has a rotating buffer (on disk or in memory), we should add a mode where we _don't_ upload unless some tigger occurs.

Potentially this could be set up with a fairly low fixed budget on disk + a way to have the pipeline off by default?

A design PR would be a good place to start

cc @zz85 

Some trade offs to consider:
- what work happens in all cases vs. only when triggered
- how do the application triggers work? Are there default triggers we should ship?

zz85:
another idea that my coworkers raised - instead of a trigger to dump, if we are already writing to a circular buffer on disk, simply mark the traces for eligible for uploading to S3

PR #494, opened 2026-06-03
docs(design): Manual trigger for pipeline run

Fluzko:
related to #469 

## Summary
- Two layers, layer 1 is a generic `WorkerSchedule::OnTrigger` + `PipelineRun` (process_now / process_window) that lets any pipeline (custom or preset) run on demand. Layer 2 is an S3 sugar preset `with_s3_dump_on_trigger` + `DumpHandle` that adds dump semantics (id, name, prefix, completion sentinel) on top.
- Trigger gates when, not what. The same `SegmentProcessor` chain executes on demand. No new pipeline stages, just one schedule enum, one handle, one end-of-run hook.
- Two trigger flavors as peers. `dump_now` (snapshot of the ring) and `dump_window`(post) (snapshot plus post-trigger duration). Pre-trigger scope is bounded by the ring's existing max_total_size.


docs/design/on-trigger-pipeline-runs.md
```diff
@@ -0,0 +1,466 @@
...
+`build`, one end-of-run hook.
+
+## Layer 1: Generic on-trigger schedule
+
+### 1. Worker schedule
+
+A new enum exposed on the builder, orthogonal to pipeline selection:
+
+```rust
+pub enum WorkerSchedule {
```
rcoh: just a smal note that this needs to be non_exhasutive

docs/design/on-trigger-pipeline-runs.md
```diff
@@ -0,0 +1,466 @@
...
+```
+
+The schedule is independent of `with_custom_pipeline`,
+`with_s3_uploader`, or any future pipeline preset. A representative
+composition:
+
+```rust
+let (guard, run) = TracedRuntime::builder()
+    .with_custom_pipeline(|p| p.symbolize().gzip().write_back())
+    .with_worker_schedule(WorkerSchedule::OnTrigger)
```
rcoh: I would recommend passing the rx side of the trigger into `OnTrigger`. Otherwise we'll have really nasty typestate stuff

docs/design/on-trigger-pipeline-runs.md
```diff
@@ -0,0 +1,466 @@
...
+#[derive(Clone)]
+pub struct PipelineRun {
+    // sender into the worker's request channel
+}
+
+impl PipelineRun {
+    /// Snapshot run. Captures the set of segments currently sealed at
+    /// invocation, runs them through the configured pipeline, resolves
+    /// after the last segment finishes its pipeline.
+    pub async fn process_now(&self) -> Result<RunReceipt, RunError>;
```
rcoh: you might want to do a "timespan" request instead (e.g. caller basically says I want -5m -> +5m)

docs/design/on-trigger-pipeline-runs.md
```diff
@@ -0,0 +1,466 @@
...
+per-segment S3 objects today (`background_task/s3.rs` sets `service`,
+`boot-id`, `segment-index`, `start-time`, `host` as user-defined
+metadata on every uploaded object). The generic primitive in Layer 1
+stays neutral: `process_now`, `process_window`.
+
+### 1. Builder preset
+
+```rust
+let (guard, dump) = TracedRuntime::builder()
+    .with_s3_dump_on_trigger(s3_config)
```
rcoh: I'm not sure why this has anything to do with s3, seems like it is a generic concept of attaching metadata to files (which we already have various ways of doing.) I'd suggest we push metadata into the state, then the pipeline stages can decide how they want to handle it

rcoh (commented):
this seems in the right direction. I'd like to see another pass that is two part:
1. Framed explicitly around the customer experience. "write the docs" as it were. I think when you do this, the amount of complexity we're adding here will become apparent
2. Framed around implementation details

I think we're introducing a lot of different concepts here, but I bet we can find one simpler concept that will handle most of the cases.

I'd encourage you to think about it around an API like:

```

let (control, rx) = worker_schedule::control_channel();

Dial9Config().emit_trace_control(rx)....


// sometime later
control.dump_now(...)
control.dump_timerange(...)

```

I suspect that can unify things

docs/design/on-trigger-pipeline-runs.md
Fluzko: Well, this comes with some decisions we have to make... do we want to allow boundaries outside the ring buffer? If so, shall we put a limit on how wide that window can be?

Fluzko: left some alternatives to decide which one we one, based on their tradeoffs.

rcoh: I would definitely allow outside of the buffer; you might imagine that in an event you want to report, e.g. "the next 5 minutes of data"

docs/design/on-trigger-pipeline-runs.md
```diff
@@ -0,0 +1,429 @@
...
+    .with_trigger(rx)
+    .build()?;
+
+// Hand `control` to whatever subsystem decides when to dump
+// (an idle-ratio watcher, a panic hook, a `/dump` HTTP handler, ...).
+// `TriggerControl` is `Clone`; share it freely.
+
+let receipt = control
+    .dump_window(Some(Duration::from_secs(300)), Duration::from_secs(60))
+    .with_metadata([("reason".into(), "idle-ratio-drop".into())].into())
```
rcoh: this is a lot of `.into()`, maybe we can improve this API (e.g. by allowing repeated `with_metadata` and having each arg be `impl Into<String>`

docs/design/on-trigger-pipeline-runs.md
```diff
@@ -0,0 +1,429 @@
...
+);
+```
+
+That is the whole user-facing surface for the S3 case. No
+`WorkerSchedule` enum, no dump-specific builder preset, no end-of-run
+callback to plumb. The application registers a receiver, holds a
+sender, and awaits a receipt. Completion is signalled by the receipt
+resolving; applications that need a durable cross-process signal can
+publish it through whatever channel they already use (incidents-table
+row, Slack message, etc.).
```
rcoh: please clean up the design doc so it doesn't refer to old versions; will be confusing for readers

docs/design/on-trigger-pipeline-runs.md
```diff
@@ -0,0 +1,429 @@
+# On-Trigger Pipeline Runs
+
+## Overview
```
rcoh: for these design docs, I'd like to encourage that the first section of the doc is _literally_ what you plan to write as the user facing docs

docs/design/on-trigger-pipeline-runs.md
```diff
@@ -0,0 +1,429 @@
...
+```rust
+// Whole ring at trigger time; resolve as soon as those segments
+// finish the pipeline.
+control.dump_now().await?;
+
+// Look-back plus a post-trigger window. "-5 min .. +1 min": only
+// segments whose `end_epoch >= now - 300s` are included from the
+// ring, then keep capturing for another 60s before resolving.
+control
+    .dump_window(Some(Duration::from_secs(300)), Duration::from_secs(60))
```
rcoh: lets workshop the name of this API. perhaps

`control.dump_current_data()` (for dump now)

`control.dump_time_range(<lookback>, <look forward>)`

docs/design/on-trigger-pipeline-runs.md
```diff
@@ -0,0 +1,429 @@
...
+// Whole ring plus a post-trigger window. `pre = None` means take
+// everything the ring still has.
+control
+    .dump_window(None, Duration::from_secs(60))
+    .await?;
+```
+
+The pre filter is best-effort: the ring keeps only what
+`max_total_size` lets it keep, so the actual covered span (reported as
+`receipt.time_range`) may be shorter than the requested `pre`. The
```
rcoh: `pre`? did this use to refer to a variable?

docs/design/on-trigger-pipeline-runs.md
```diff
@@ -0,0 +1,429 @@
...
+sorted by time, and `ListObjectsV2 prefix={prefix}/dumps/{dump_id}/`
+returns one dump.
+
+Per-segment objects carry the same S3 user metadata as continuous-mode
+uploads (`service`, `boot-id`, `segment-index`, `start-time`, `host`;
+see `background_task/s3.rs`) plus `dump-id`. Callers that want
+additional correlation pairs (e.g. a human-readable reason, an
+incident id) pass them via `.with_metadata(...)` on the `DumpRun`;
+pipeline stages decide what to do with them (the S3 stage can surface
+them as additional user metadata, a custom redactor can read them, etc.).
```
rcoh: hmmm...I'm not positive I'd do it this way. I would probably put the files in the "normal place". If you want, you could upload a metadata file to `/dumps` that pointed to the files, but I would keep the actual trace files in the same place as they would be during normal operation

docs/design/on-trigger-pipeline-runs.md
```diff
@@ -0,0 +1,429 @@
...
+not subsumed and remains an alternative if "stay strictly within
+`max_total_size`, no matter what" is a hard requirement.
+
+Decision pending.
+
+## API reference
+
+```rust
+pub mod trigger {
+    pub fn channel() -> (TriggerControl, TriggerRx);
```
rcoh: `trigger::channel()` is a bit of an odd name. Do we already have a pipeline module? you could do `pipeline::trigger()`

docs/design/on-trigger-pipeline-runs.md
```diff
@@ -0,0 +1,429 @@
...
+    /// `None` takes whatever the ring still has. `post` is the
+    /// wall-clock window of post-trigger arrivals to include.
+    pub fn dump_window(&self, pre: Option<Duration>, post: Duration) -> DumpRun<'_>;
+}
+
+/// In-flight dump request. Resolves to `Result<DumpReceipt, DumpError>`
+/// when awaited (via `IntoFuture`). Chain `.with_metadata(...)` before
+/// awaiting to attach correlation pairs.
+#[must_use = "DumpRun does nothing unless awaited"]
+pub struct DumpRun<'a> { /* private */ }
```
rcoh: is this really must use? It would be _very_ odd if I had to await a dump run for it to actually dump. I would much rather that awaiting the dump run was optional.

docs/design/on-trigger-pipeline-runs.md
```diff
@@ -0,0 +1,429 @@
...
+}
+```
+
+`with_trigger(rx)` on `TracedRuntimeBuilder` writes a private
+`Option<TriggerRx>`. Absence (the default) keeps today's continuous
+behavior; presence flips the worker into triggered mode. `build()`
+returns the existing `TelemetryGuard` regardless; no new axis on the
+builder's phantom-state machinery.
+
+Implementation note on the S3 uploader: `object_key` in
```
rcoh, line 418: recommend making metadata files to capture the dump instead of putting the files in a different place.

Fluzko: So I deliberately removed the metadata files (used to call it `manifest`) since it would add complexity:
- two S3 fetches to retrieve the data, and logically join them to get the full context
- two files to upload
- the metadata insertion is already managed, so this would only add dump-id as the filter

Let me know if you feel more convinced with the matadata file, but I see it a bit too much and not giving any real value for a first iteration

rcoh: I don't see two fetches as problematic; its probably simpler than traversing the bucket anyway. I'm also not bothered by having two files. But if the files are in the "normal" place, we do still need some nice way to find them.

It seems harmless to write the file if only to give a quick way to find which files are relevant?

docs/design/on-trigger-pipeline-runs.md
```diff
@@ -0,0 +1,328 @@
...
+- `self.fs.writer_done` (existing, used to start drain-to-empty)
+- the new `trigger_rx` populated by `DumpControl`
+
+It does not call `take_files` between triggers. When a request arrives, it
+takes a snapshot of the segments currently in the ring (optionally filtered to
+`end_epoch >= now - lookback`), stamps the metadata, runs each captured
+segment through the same processor chain, and resolves the receipt when the
+last one finishes. Then it re-parks.
+
+Because a dump never claims future arrivals, there is no contention over the
```
rcoh: I don't think this is true? what does it mean "does not claim future arrivals"? I thought you could specify a time range that covered both look forward and look back. It seems like you deleted the dump_window API? was that intentional?

Fluzko: Yes, so I got confused (deleting old comment to avoid noise). We can do a best effort to upload segments from oldest to newer based on the window (the oldest are the ones that will be deleted first) and if the writes in the buffer are faster than the uploads then we upload what we find in when we move forward (best effort), so we avoid pinning.
Will rewrite with this in mind

rcoh (approved):

PR #495, opened 2026-06-03
memory profiling: Optimize memory profiling by tracking liveset on the allocator side

rcoh:
The current design that pushes every free into an array queue suffered from bad performance for a couple of reasons. This updates the design to use `scc` on the allocator side to track the liveset.

This is _generally_ good, but runs into some trickiness, especially around teardown. There is a race condition where the thread locals inside of SCC race with another thread local that is being freed.

I solved this with an OPT_OUT_TLS sentinel that seems to handle the problem. I also removed the optimization where we use a cached timestamp as it presented some cross-thread ordering issues with the liveset. I think we may _eventually_ move to a hybrid approach where, in general, we have a timer that uses clock monotonic as a base time, but within a short window, e.g. 1s we use a quanta-based approach

### DIFF
--- a/dial9-tokio-telemetry/src/memory_profiling/source.rs
+++ b/dial9-tokio-telemetry/src/memory_profiling/source.rs
@@ -140,10 +140,9 @@ impl Source for MemoryProfileSource {
     fn flush(&mut self, ctx: &FlushContext<'_>) {
         // Merge-sort drain by timestamp. This produces a best-effort
         // timestamp-ordered stream. Ordering is not guaranteed to be perfect:
-        // - Multiple producers push concurrently, so queue order may not
-        //   match timestamp order.
-        // - TimestampMode::ReusePollStart can produce stale timestamps.
-        // For profiling purposes, approximate ordering is sufficient.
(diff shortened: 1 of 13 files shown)


dial9-tokio-telemetry/src/memory_profiling/source.rs
```diff
@@ -89,32 +90,45 @@ impl MemoryProfileSource {
...
+        //   non-shutdown free will emit the correct event.
+        let (size, alloc_ts_ns) = if f.shutdown {
+            let Some((size, alloc_ts_ns)) = liveset.peek_with(&f.addr, |_, v| *v) else {
+                return; // already cleaned, or never sampled
+            };
+            if alloc_ts_ns >= f.ts_ns {
+                return; // address-reuse race or timestamp tie — see comment above
+            }
+            liveset.remove(&f.addr);
+            (size, alloc_ts_ns)
```
jv1i, line 119: do we want to use `entry()` here like in `on_dealloc`, to keep the peek + remove atomic?

jv1i (approved):
LGTM

PR #497, opened 2026-06-05
analysis improvements

rcoh:
A bunch of small improvements to `analyze.js` so that the output is more actionable & correct

summary:

- **Long poll CPU stacks: show full call chains**
  - Before: Only showed single leaf/trigger frame (e.g., `Store::shift_bins`)
  - After: Shows full app-level call chain with ellipsis for long stacks, matching the treatment scheduling samples already had

- **Long polls: show source file**
  - Before: `▸ 110.86ms on worker 3 at 89.01s`
  - After: `▸ 110.86ms on worker 3 at 89.01s [1780512780-585.bin]`

- **Scheduling delays: don't mislead when no tasks instrumented**
  - Before: `No delays > 1ms ✅` (implies everything is fine)
  - After: `No tasks instrumented — use dial9::spawn() to measure scheduling delays`

- **CPU hotspots: flamegraph-style hierarchical tree**
  - Before: Flat list of leaf frames (`leaf: __libc_recv / app frame: TcpStream::read`)
  - After: Full call tree with branching, sample counts at each level, enabling drill-down without custom scripts

- **Poll duration by spawn: sorted with outlier markers**
  - Before: Sorted by max, no visual indicators
  - After: Sorted by max descending, ⚠️ for max > 10ms, 🟡 for max > 1ms

- **Blocking calls: accurate messaging**
  - Before: `No off-CPU samples in trace`
  - After: `Schedule profiling not enabled` with enable instructions + shows kernel scheduling delays > 1ms if present

- **Peak active tasks: correct count**
  - Before: `Peak active tasks: 1` (wrong — didn't account for tasks alive at trace start)
  - After: `Peak active tasks: 151 (150 pre-existing at trace start)`

- **Trace summary: list files**
  - Before: Just showed file count in header
  - After: Lists all analyzed files at top of summary

- **frameKind: filter runtime noise from stacks**
  - Before: `BlockingTask::poll`, `FnOnce::call_once{{vtable.shim}}`, `Thread>::new::thread_start` appeared as "app" frames
  - After: Properly classified as runtime frames and excluded from app-level output

### DIFF
--- a/dial9-viewer/skills/dial9-toolkit/scripts/analyze.js
+++ b/dial9-viewer/skills/dial9-toolkit/scripts/analyze.js
@@ -40,8 +40,10 @@ function frameKind(sym) {
     if (/^(schedule|expand_fd|expand_files|alloc_fd|ep_poll|ep_send|futex_|sock_|inet_|tcp_|vfs_|filp_|fd_install|ksys_|fput|fget)/.test(sym)) return 'kernel';
   }
   // Tokio runtime internals
-  if (/^tokio::runtime::(scheduler|task|blocking|context|coop)/.test(sym)) return 'runtime';
+  if (/^<?tokio::runtime::(scheduler|task|blocking|context|coop)/.test(sym)) return 'runtime';
   if (/^std::(panicking|sys|thread)::/.test(sym)) return 'runtime';
+  if (/^<std::(sys|thread)::/.test(sym)) return 'runtime';
+  if (/^core::(ops::function|panicking)::/.test(sym)) return 'runtime';
   return 'app';
 }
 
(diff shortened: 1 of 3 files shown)


dial9-viewer/skills/dial9-toolkit/scripts/analyze.js
```diff
@@ -82,6 +84,7 @@ function fmtRel(ns, minTs) { return fmtDur(ns - minTs); }
 /** Create an empty accumulator for multi-trace analysis. */
 function createAccumulator() {
   return {
+    files: [],
```
jv1i: nit: AFAICT this `files: []` is never used, not populated because finalize drops it and returns a new literal, then result.files is set at the call sites instead, might want to either wire it through or could just drop it.

jv1i (approved):

PR #499, opened 2026-06-08
fix(test): flakyness around sched events

### LINKED ISSUE #492: flaky test
```
---- sched_events_sampling_reduces_count stdout ----

thread 'sched_events_sampling_reduces_count' (12675) panicked at dial9-tokio-telemetry/tests/sched_events.rs:127:5:
expected ~10x ratio, got 12.1x (n_all=241, n_sampled=20)
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/d595fce01043347bf7f80e85b76dcc41b59a3e6e/library/std/src/panicking.rs:678:5
   1: core::panicking::panic_fmt
             at /rustc/d595fce01043347bf7f80e85b76dcc41b59a3e6e/library/core/src/panicking.rs:80:14
   2: sched_events::sched_events_sampling_reduces_count
             at ./tests/sched_events.rs:127:5
   3: sched_events::sched_events_sampling_reduces_count::{closure#0}
             at ./tests/sched_events.rs:69:41
   4: <sched_events::sched_events_sampling_reduces_count::{closure#0} as core::ops::function::FnOnce<()>>::call_once
             at /rustc/d595fce01043347bf7f80e85b76dcc41b59a3e6e/library/core/src/ops/function.rs:250:5
   5: <fn() -> core::result::Result<(), alloc::string::String> as core::ops::function::FnOnce<()>>::call_once
             at /rustc/d595fce01043347bf7f80e85b76dcc41b59a3e6e/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Fluzko:
closes #492 
 
## Summary
Measure the ratio within a single run against an independent ground truth instead of across two runs. 
The kernel exposes each thread's context-switch count in `/proc/.../status` (`voluntary_ctxt_switches` + `nonvoluntary_ctxt_switches`), which is the same quantity perf's `SwContextSwitches` event counts.
Because recorded samples and true switches now come from the same threads in the same run, `records ~= total / 10` holds with only the +/- 1-record sampling boundary as error, regardless of how many switches occur. The real ~10x guarantee is kept, the cross-run variance is gone.

I've ran them in a privileged `rust:1.94.1-bookworm` container with `kernel.perf_event_paranoid=1`. Each ratio test looped 130 times with zero failures (30 baseline runs, plus 100 runs under CPU contention, 22 busy loops saturating all cores) to maximize involuntary context switches, the condition most likely to expose any residual boundary error.

### DIFF
--- a/dial9-tokio-telemetry/tests/sched_events.rs
+++ b/dial9-tokio-telemetry/tests/sched_events.rs
@@ -66,65 +66,91 @@ fn sched_events_capture_context_switches() {
     assert_eq!(cpu_profile_samples, 0, "should have no CpuProfile samples");
 }
 
+/// With `sampling_interval(10)`, perf records ~1/10 of the worker threads'
+/// context switches. Rather than compare two independent runs (whose total
+/// switch counts vary, making the ratio flaky), we run once and compare the
+/// emitted sample count against the kernel's own per-worker context-switch
+/// counter read from `/proc`. Numerator and denominator come from the same
+/// threads in the same run, so the ~10x relationship holds deterministically.
 #[test]
 fn sched_events_sampling_reduces_count() {
     use dial9_tokio_telemetry::telemetry::TracedRuntime;
     use dial9_tokio_telemetry::telemetry::cpu_profile::SchedEventConfig;
+    use std::collections::HashSet;
     use std::time::Duration;
 
-    let count_sched_samples = |interval: Option<u64>| -> usize {
-        let (writer, batches) = BytesCapturingWriter::new();
+    const PERIOD: u64 = 10;
+    let num_workers = 2u64;
 
-        let num_workers = 2;
-        let mut builder = tokio::runtime::Builder::new_multi_thread();
-        builder.worker_threads(num_workers).enable_all();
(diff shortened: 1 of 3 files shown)


dial9-tokio-telemetry/tests/sched_events.rs
```diff
@@ -66,65 +66,91 @@ fn sched_events_capture_context_switches() {
...
+                for _ in 0..5 {
+                    std::thread::sleep(Duration::from_millis(2));
+                    tokio::task::yield_now().await;
+                }
+            }));
+        }
+        for h in handles {
+            h.await.unwrap();
         }
+        tokio::time::sleep(Duration::from_millis(500)).await;
```
rcoh: instead of `time.sleep` use `guard.graceful_shutdown`

rcoh (approved):

dial9-tokio-telemetry/tests/sched_events.rs
Fluzko: Done!

rcoh (approved):

PR #500, opened 2026-06-08
refactor!: remove TraceWriter trait

### LINKED ISSUE #482: Remove `TraceWriter` trait

jv1i:
closes #482

# Overview

Removes the `TraceWriter` trait and its test implementations.

`SegmentWriter<Mode>` (the `DiskWriter` / `InMemoryWriter` aliases) is now the only writer. 

This is a breaking change, `TraceWriter` and `NullWriter` were public.

## Changes

**Writer**
 `SegmentWriter`'s methods are now inherent instead of trait methods (`flush` / `finalize` / `write_encoded_batch` / `update_segment_metadata` are `pub`, the flush-loop-only ones are `pub(crate)`). 
 `NullWriter` is replaced by `InMemoryWriter::discard()`, a fixed-size in-memory writer whose output is never
delivered. The flush loop and `TelemetryCore::new` now take a concrete `SegmentWriter<Mode>`.

**The build API now infers the writer mode** 
Previously the only way to reach in-memory mode on the builder was to call a pipeline method, so code that just
wanted to discard output had to write `.with_custom_pipeline(|p| p)` (or something like `with_s3...` purely to flip the mode marker, this is no longer necessary. 

On the no-pipeline state, `build` / `build_and_start` take the writer and infer `Mode` from it:

```rust
// disk
TracedRuntime::builder().build_and_start(tk, DiskWriter::single_file(path)?)?;
// in-memory, output discarded
TracedRuntime::builder().build_and_start(tk, InMemoryWriter::discard())?;
```

A configured pipeline pins the mode, so the custom and S3 pipeline states keep a `SegmentWriter<Mode>` argument that must match. Pairing a disk pipeline with an in-memory writer stays a compile error.

The `*_with_writer` build methods are removed: call `build` / `build_and_start`.

## Breaking changes

- `TraceWriter` and `NullWriter` are removed. Replace `NullWriter` with `InMemoryWriter::discard()`.
- `build_with_writer` / `build_and_start_with_writer` are removed. Use `build` / `build_and_start` (they now take the writer directly and infer the mode).
- Custom `TraceWriter` implementations are no longer possible. To transform or ship sealed segments, implement `SegmentProcessor` and add it with `with_custom_pipeline`.

## Testing

Tests that used to capture output through a custom `TraceWriter` now run a capturing `SegmentProcessor` over a real `InMemoryWriter` and read back the sealed segments after `graceful_shutdown`. The shuttle tests use a real `DiskWriter` and the `fail_all_writes` test hook.

## Note of reviewers

I intentionally didn't dig too much into improving the config API or removing some of the duplication the new mode inference requires. We'll probably rework that either way after crate structure v2.

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/recorder/shared_state.rs
+++ b/dial9-tokio-telemetry/src/telemetry/recorder/shared_state.rs
@@ -487,7 +487,7 @@ mod shuttle_tests {
     use crate::primitives::sync::atomic::{AtomicU64, Ordering};
     use crate::primitives::sync::{Arc, Mutex};
     use crate::telemetry::recorder::TelemetryCore;
-    use crate::telemetry::writer::DiskWriter;
+    use crate::telemetry::writer::{DiskWriter, InMemoryWriter};
     use dial9_trace_format::TraceEvent;
     use shuttle::rand::Rng;
     use std::collections::HashMap;
@@ -614,13 +614,14 @@ mod shuttle_tests {
         let num_threads = 3;
         let next_id = Arc::new(AtomicU64::new(0));
 
-        let dir = tempfile::tempdir().unwrap();
-        let writer = DiskWriter::builder()
-            .base_path(dir.path().join("trace"))
-            .max_file_size(256)
+        // Small segments force frequent rotation: the 100 MiB budget is far above the test's data,
+        // so the ring never evicts before we drain it.
+        let writer = InMemoryWriter::builder()
             .max_total_size(100 * 1024 * 1024)
+            .max_segment_size(256)
             .build()
             .unwrap();
+        let fs = writer.fs_handle().expect("in-memory writer exposes its fs");
 
         // Mock source: worker threads push events here, flush thread drains them.
         let source_pending: Arc<Mutex<Vec<ValidationEvent>>> = Arc::new(Mutex::new(Vec::new()));
@@ -672,17 +673,18 @@ mod shuttle_tests {
         }
         drop(guard);
 
-        // Decode per-segment from the sealed `.bin` files.
-        let mut files: Vec<std::path::PathBuf> = std::fs::read_dir(dir.path())
-            .unwrap()
-            .filter_map(|e| e.ok().map(|e| e.path()))
-            .filter(|p| p.to_string_lossy().contains(".bin"))
-            .collect();
-        files.sort();
-        let all_decoded: Vec<ValidationEvent> = files
-            .iter()
-            .flat_map(|p| decode_validation_events(&std::fs::read(p).unwrap()))
-            .collect();
+        // Drain the in-memory ring (memory pops one sealed segment per call).
+        let mut all_decoded: Vec<ValidationEvent> = Vec::new();
+        loop {
+            let taken = fs.take_files();
+            if taken.segments.is_empty() {
+                break;
+            }
+            for seg in taken.segments {
+                let (_seg_ref, payload, _accounting) = seg.load().unwrap();
+                all_decoded.extend(decode_validation_events(&payload.into_vec()));
+            }
+        }
         let expected = expected.lock().unwrap();
 
         // Run all invariants.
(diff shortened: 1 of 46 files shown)


dial9-tokio-telemetry/src/telemetry/recorder/shared_state.rs
```diff
@@ -658,15 +612,19 @@ mod shuttle_tests {
...
         let next_id = Arc::new(AtomicU64::new(0));
-        let segments: Arc<Mutex<Vec<Vec<u8>>>> = Arc::new(Mutex::new(Vec::new()));
+
+        let dir = tempfile::tempdir().unwrap();
+        let writer = DiskWriter::builder()
+            .base_path(dir.path().join("trace"))
+            .max_file_size(256)
+            .max_total_size(100 * 1024 * 1024)
+            .build()
+            .unwrap();
```
rcoh: do you think we want a disk writer or an in memory writer for the shuttle tests?

dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -265,8 +162,11 @@ pub struct SegmentWriter<Mode: WriterMode = Disk> {
     active_path: PathBuf,
     state: WriterState,
     next_index: u32,
-    /// Set after rotation; cleared by `take_rotated()`.
-    did_rotate: bool,
+    /// Test-only error injection: when set, the fallible methods return
+    /// `PermissionDenied` and `should_drain` always fires. Set via
+    /// [`fail_all_writes`](Self::fail_all_writes).
+    #[cfg(test)]
+    fail_writes: bool,
```
rcoh: or I guess this could be a different `WriterMode` or `_mode` could actually hold an object which would allow a `Disk` that would fail writes

dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -469,6 +371,21 @@ impl SegmentWriter<Memory> {
...
+    #[doc(hidden)]
+    pub fn discard() -> Self {
+        Self::create_in_memory(
+            16 * 1024 * 1024,
+            4 * 1024 * 1024,
+            DEFAULT_ROTATION_PERIOD,
+            SegmentMetadata::default(),
+        )
+        .expect("discard() sizes satisfy create_in_memory validation")
+    }
```
rcoh: this doesn't actually discard though right? The thing that would cause it to discard is not configuring a pipeline?

might be worth just deleting this and the tests that rely on it

dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -763,28 +680,47 @@ impl<M: WriterMode> SegmentWriter<M> {
...
 
-impl<M: WriterMode> TraceWriter<M> for SegmentWriter<M> {
-    #[allow(private_interfaces)]
-    fn fs_handle(&self) -> Option<Arc<Fs>> {
+impl<M: WriterMode> SegmentWriter<M> {
+    /// Test-only: make every fallible write return `PermissionDenied` and
+    /// force `should_drain` to fire. Used by the shuttle error-injection test.
+    #[cfg(test)]
+    #[cfg_attr(not(shuttle), allow(dead_code))]
+    pub(crate) fn fail_all_writes(mut self) -> Self {
```
rcoh: an alternative to consider: make a `primitives::fs` module where in shuttle, sometimes the disk APIs fail (configurable via a thread local or something)

rcoh (commented):
overall looks good. I think its worth spending a bit more time thinking about the shuttle/mocking story around the fs.

dial9-tokio-telemetry/src/telemetry/recorder/shared_state.rs
jv1i: should have been memory (this was migration artifact), we have better shuttle coverage (for now) on the memory side, updated now

dial9-tokio-telemetry/src/telemetry/writer.rs
jv1i: Good catch got his downgraded on iteration, it didn't discard on its own, I was just using it as a shorthand of a small mem writer to replace all the uses of NullWriter. 
I replaced it now with both just inlining `InMemoryWriter::new()` in some places and a test utility for it `small_mem_writer()`.

dial9-tokio-telemetry/src/telemetry/writer.rs
jv1i: Good idea! I went for this approach, only thing might read slightly confused is that we have `Fs` as an abstraction (memory or disk) and then `fs` primitives, but I think in context it's understandable (otherwise can rename on of them)

rcoh (approved):

PR #506, opened 2026-06-09
feat: socket accept queues source

### LINKED ISSUE #501: socket accept queues

noxware:
# Changes
Adds a source that reports socket accept queues stats by sampling from Source::flush with a specific interval. 

# Other
Closes #501


dial9-tokio-telemetry/src/telemetry/format.rs
```diff
@@ -177,6 +177,30 @@ pub struct ProcessResourceUsageEvent {
...
+    #[traceevent(timestamp)]
+    pub timestamp_ns: u64,
+    /// Linux socket inode reported by sock_diag.
+    pub socket_inode: u64,
+    /// IP version for `local_addr`: 4 or 6.
+    pub ip_version: u8,
+    /// IP protocol number. TCP is 6.
+    pub protocol: u8,
+    /// Local listener address.
+    pub local_addr: String,
```
noxware: I can probably make this some kind of binary format like `[u8; 16]` to avoid the human formatting of `String` and or the allocation of a `Vec<u8>`, but I would need to add support for that in the format, which I suppose is fragile to update without discussion.

dial9-viewer/ui/trace_parser.js
```diff
@@ -663,12 +663,24 @@
     }
     const hasTimeFilter = startTime > 0 || endTime < Infinity;
 
-    // Compute timestamp bounds from events (safe for large arrays)
-    let evMinTs = Infinity, evMaxTs = -Infinity;
-    for (let i = 0; i < events.length; i++) {
-      const t = events[i].timestamp;
-      if (t < evMinTs) evMinTs = t;
-      if (t > evMaxTs) evMaxTs = t;
+    // Bounds cover every parsed timestamp-bearing stream so a full-range slice
```
noxware: Basically, in my machine, there are custom events emitted after the last runtime event, and `maxTs` was considering only the runtime events not all. These JS changes accounts for that possibility. And the problem seems pre-existing in main. Does it makes sense?

dial9-tokio-telemetry/src/telemetry/socket_accept_queues.rs
```diff
@@ -0,0 +1,354 @@
+//! Socket accept queue snapshots sampled from Linux sock_diag.
+
+use std::time::Duration;
+
+const DEFAULT_SAMPLE_INTERVAL: Duration = Duration::from_secs(1);
```
rcoh: how was this chosen?

dial9-tokio-telemetry/src/telemetry/socket_accept_queues.rs
```diff
@@ -0,0 +1,354 @@
...
+        let mut socket = Socket::new(NETLINK_SOCK_DIAG)?;
+        let _local_addr = socket.bind_auto()?;
+        let kernel_addr = SocketAddr::new(0, 0);
+        socket.connect(&kernel_addr)?;
+
+        let mut snapshots = Vec::new();
+        dump_tcp_listeners(&socket, AF_INET, 1, &socket_inodes, &mut snapshots)?;
+        dump_tcp_listeners(&socket, AF_INET6, 2, &socket_inodes, &mut snapshots)?;
+        Ok(snapshots)
+    }
```
rcoh: can we keep this socket open instead of having to reopen it every time?

dial9-tokio-telemetry/src/telemetry/format.rs
```diff
@@ -177,6 +177,30 @@ pub struct ProcessResourceUsageEvent {
     pub involuntary_context_switches: u64,
 }
 
+/// Wire-format event for a TCP listener accept queue snapshot.
+#[derive(Debug, TraceEvent)]
+#[traceevent(wire_slot)]
+#[cfg_attr(not(feature = "unstable-events"), non_exhaustive)]
```
rcoh: this just shouldn't be public at all; we should make a separate version of the event to be parsed utilizing `derive(Deserialize)`

dial9-tokio-telemetry/src/telemetry/socket_accept_queues.rs
```diff
@@ -0,0 +1,354 @@
...
+
+            match collect_socket_accept_queues() {
+                Ok(snapshots) => {
+                    let timestamp_ns = clock_monotonic_ns();
+                    for snapshot in snapshots {
+                        record_encodable_event(
+                            &snapshot.into_event(timestamp_ns),
+                            ctx.collector,
+                            ctx.drain_epoch,
+                        );
```
rcoh, line 153: is this one per open socket? that seems like it could be a lot on a host with 5000 open connections?

rcoh:
I think there's actually some nuance in the design about what events make sense to emit that is worth discussing

dial9-tokio-telemetry/src/telemetry/socket_accept_queues.rs
noxware: Initially just being conservative based on a simple test I did a while ago but...

The overhead of this implementation is quite real. It scales with the scanning of `/proc/self/fd`, which means that the more listeners + accepted connections you have, the slower this gets.

Most servers have very few listeners, maybe just 1. In this case, the sample may just take `20-50us` in my machine. But, with a high volume of FDs, like 5000 (from my tests), this can get up to `4ms` in my machine which is quite a lot! (unexpected, honestly).

But the trick is... This FDs are not only from the listeners, but they may be from active accepted connections as well. And I think this is another problem with this implementation. Having thousands of active accepted connections is quite normal. And already accepted connections should not impact this measurement.

So... hmmm... I feel like we would need to deal with that to make this production ready.

dial9-tokio-telemetry/src/telemetry/socket_accept_queues.rs
noxware: I tried a small change that does this by leaking the socket and measured against the best and worst cases of my previously mentioned tests. On the best one, you see a tiny `2-5us` improvement (very little), and on the worst is basically unnoticeable. 

So the re-connection of these kind of sockets doesn't seem expensive. Also, the re-connection is one per interval, not one per socket query, with 1 sec interval this is.... negligible.

We could keep it open, just in case, but we would need to handle the cleanup on flush shutdown. As I mentioned, `Drop` would no be enough for this because `SharedState` can live after dropping the telemetry guard if the handle is held. To make this perfectly, it would be ideal to add a `Source::shutdown` function to the trait (with default empty impl).

dial9-tokio-telemetry/src/telemetry/socket_accept_queues.rs
noxware: Nah, it's one per listener. For each listener in your server process, in each interval, a sample is made for that single listener, and emitted. That single sample contains values like "how many TCP handshakes ended, without `accept()` being reached", which is just an aggregation/sum/count. 

Screenshot 2026-06-10 at 4 54 22 PM

noxware: > what events make sense to emit that is worth discussing

was that about this?

dial9-tokio-telemetry/src/telemetry/socket_accept_queues.rs
```diff
@@ -0,0 +1,354 @@
...
+                io::ErrorKind::WriteZero,
+                format!(
+                    "short netlink sock_diag request write: sent {sent} of {} bytes",
+                    request_bytes.len()
+                ),
+            ));
+        }
+
+        loop {
+            let (response_bytes, _sender) = socket.recv_from_full()?;
```
rcoh, line 245: we should set a timeout here

dial9-tokio-telemetry/src/telemetry/socket_accept_queues.rs
```diff
@@ -0,0 +1,354 @@
...
+                local_addr: self.local_addr.to_string(),
+                local_port: self.local_port,
+                pending_connections: self.pending_connections,
+                backlog_limit: self.backlog_limit,
+            }
+        }
+    }
+
+    fn collect_socket_accept_queues() -> io::Result<Vec<SocketAcceptQueueSnapshot>> {
+        let socket_inodes = current_process_socket_inodes()?;
```
rcoh: I think we could optimize this by only requerying the socket inodes when we actually find a new socket_inode after dumping the tcp listeners _and_ finding a new inode

dial9-tokio-telemetry/src/telemetry/format.rs
rcoh: when it comes time to optimize, we should support &'str or something like that to save allocation. not worth optimizing right now though I think

dial9-tokio-telemetry/src/telemetry/format.rs
```diff
@@ -177,6 +177,32 @@ pub struct ProcessResourceUsageEvent {
...
+    /// IP protocol number. TCP is 6.
+    pub protocol: u8,
+    /// Local listener address.
+    pub local_addr: String,
+    /// Local listener port.
+    pub local_port: u16,
+    /// Completed connections waiting to be accepted.
+    pub pending_connections: u32,
+    /// Effective accept backlog limit.
+    pub backlog_limit: u32,
```
rcoh: this all can be pub(crate)

dial9-tokio-telemetry/README.md
```diff
@@ -207,6 +215,31 @@ is enabled. To opt out, set:
 DIAL9_PROCESS_RESOURCE_USAGE_ENABLED=false
 ```
 
+### Socket accept queues (Linux only)
+
+With the `socket-accept-queues` feature, dial9 can sample TCP listener accept
```
rcoh: lets name the feature something more geneal like `linux-socket`; we will probably want to add other info here as well.

rcoh (commented):
overall this looks good. Lets _detach_ this from constructors; I'd like to start having a less centralized pattern where sources are individual items that you attach instead of all being centrally configured.

So you would setup your dial9 handle, then call `add_source(...)`

possibly we can wait for @yulnr 's pr

dial9-tokio-telemetry/src/telemetry/format.rs
noxware: Yeah, just mentioning this because we probably don't even need to transport the human readable formatting of IPs over the wire. IPv4 is 32bits, IPv6 is 128bits, and the current IPv8 proposal is 64bits. So `[u8; 16]` is enough for all, present and the maybe future. Or well, an enum with fixed size variants like [std::net::IpAddr](https://doc.rust-lang.org/std/net/enum.IpAddr.html), which does that. Then, the frontend can parse that into the human readable form.

But not worth touching the dial9 trace format here. Probably.

dial9-tokio-telemetry/src/telemetry/socket_accept_queues.rs
noxware: On my tests, in my machine, inside docker, I could see `<500us`, consistently, in the cold path (no change in the listeners list of the system).

Based on that observation, we could use a more aggressive interval like `200ms` to reduce the chance of missing spikes. However, this is just my machine. And there is still the hot path to be worried.

I reduced this interval to `400ms` as a conservative sub-second precision which has the chance to trigger twice in a second.

noxware:
I renamed the event to `TcpAcceptQueueEvent` and removed the `protocol` field. The event itself is pretty "TCP-specific", because it has an IP address, a port, and IP version.

There are other kind of useful sockets, like Unix Domain Sockets, used for IPC, which also has `accept()` semantics when configured as `SOCK_STREAM`, but it's not part of the IP stack. For example, it has a local `path` instead of an IP address.

We can later add `UnixAcceptQueueEvent` with a different set of specialized fields. Seems useful.

For the public configuration API, I'm leaving the current name, `with_socket_accept_queues`, as an umbrella configuration for accept queue events of any kind. Also leaving `SocketAcceptQueuesSource`, in case that becomes public to be used with an `add_source`. We can add later a setting to select socket types to watch (if really needed).

dial9-tokio-telemetry/src/telemetry/socket_accept_queues.rs
noxware: I can add this if you feel is needed for completeness, but for that I would really like to add `Source::shutdown` to handle the disconnection instead of relying on `Drop`. Would that conflict with @yulnr's work?

Or well... This is just a little improvement but nothing blocking for merge, can be added later as well.

noxware:
> So you would setup your dial9 handle, then call add_source(...)

Sounds like a good idea if that becomes available soon.

noxware:
A lot of CI failures are coming from `main`.

dial9-viewer/ui/trace_parser.js
```diff
@@ -664,12 +664,24 @@
     }
     const hasTimeFilter = startTime > 0 || endTime < Infinity;
 
-    // Compute timestamp bounds from events (safe for large arrays)
-    let evMinTs = Infinity, evMaxTs = -Infinity;
-    for (let i = 0; i < events.length; i++) {
```
rcoh, line 669: maybe we need a better way to compute this then? it doesn't seem great to need to hardcode all the subtypes here.

rcoh (approved):

dial9-viewer/ui/trace_parser.js
noxware: Erased the JS related commits and I did a clean re-implementation of that logic. It should avoid this hardcoding.

Note: I used the latest `upstream/main` for the rebase so don't worry about your missing merge commit. Your changes are here and the CI is fully green ✅

noxware:
Green CI. But I still need to regenerate the demo trace ⚠️

dial9-viewer/ui/test_flamegraph_recipes.js
```diff
@@ -86,24 +86,25 @@ async function main() {
     assert.ok(samples.length > 0, `expected >0 samples for long polls`);
   });
 
-  // ── Recipe 5: One specific poll instance (worst poll for a task) ──
-  await testAsync("Recipe worst poll for task 3", async () => {
-    const targetTaskId = 3;
+  // ── Recipe 5: One specific poll instance (worst sampled poll for a task) ──
+  await testAsync("Recipe worst sampled poll for a task", async () => {
```
noxware, line 90: This was a flaky test in my machine. As it was, I could also set `frequency_hz(999)` to address the flakiness in my machine, without changing the test, but its probably better to update it to not depend on a specific task id.

noxware:
Okey, this should be ready for a final review 🤙

rcoh (approved):

PR #508, opened 2026-06-10
fix(analysis): O(P²) sched-delay scan + idle-vs-blocking off-CPU split

marcbowes:
Two bugs in the dial9 analysis toolkit (`analyze.js` / `trace_analysis.js`), both surfaced while analyzing large applier traces from a storage server (a long-lived task with ~2.3M polls in one 60s / ~50 MB segment, CPU-profiling + sched-events enabled). Validated against the current source and fixed; line numbers in the original report were approximate.

## Bug 1 — `computeSchedulingDelays` is O(polls_per_task²); hangs on large segments

`trace_analysis.js`, `computeSchedulingDelays`. For each poll, the mid-poll-wake adjustment linearly rescanned **every** poll of the same task to find the one whose `[start,end]` interval contains the wake timestamp. For a task with P polls this is O(P²). Most tasks are small, but a single long-lived worker/event-loop task can have millions of polls in one segment → ~10¹¹–10¹² iterations. Symptom: `analyze` on a ~50 MB / ~9M-event segment ran >14 min and never completed, one core pinned at 100%. Small/idle segments finished in seconds, masking it.

`pollsByTask[taskId]` is already sorted by `start`, and a single task's polls never overlap (a task is never polled concurrently), so at most one poll's interval can contain the wake. Replaced the linear scan with a **binary search for the rightmost poll with `start <= wake.timestamp`**, preserving the original first-match semantics exactly.

**Verification:**
- Differential test on the demo trace: old vs new produce **byte-for-byte identical** output across **71,949** scheduling delays.
- 40k-poll microbenchmark: **1437ms → 24ms (60×)**, and the quadratic blowup is visible (20k→40k makes the old code ~4× slower; the new code ~2×).
- `test_directory_scale.js` full pipeline: **67.8s → 18.8s (3.6×)** on cached data, same `1,438,980` delays, same worst poll (`17.7ms`). Parse times unchanged (apples-to-apples).

## Bug 2 — "BLOCKING CALLS" conflated idle parking with real blocking

`analyze.js`, off-CPU sample grouping + the BLOCKING CALLS report. Off-CPU (scheduler / context-switch) samples were grouped with no filter for whether the sample occurred **inside a poll** (a task voluntarily blocked mid-poll = real blocking) vs **between polls** (a worker parked because it had no work = idle). On a Tokio multi-thread runtime an idle worker parks via a futex/condvar wait, so a mostly-idle runtime rendered as:

```
BLOCKING CALLS (scheduling/off-CPU samples)
  306669 off-CPU sample(s)
  306669 samples (100.0%) — leaf: __schedule
         blocked by: futex_wait_queue (kernel)
```

i.e. "100% blocked on a futex (lock contention)" when ~95% of those samples were idle worker parking. Actively misleading for root-cause analysis.

The signal already existed: `attachCpuSamples` knows whether a sample landed inside a poll. This PR records that as `sample.inPoll` and splits the bucket:
- in-poll → **real blocking**, attributed per task spawn location, with call stacks
- not-in-poll → **idle-park** (not blocking)

New report:
```
BLOCKING CALLS (scheduling/off-CPU samples)
  5356 off-CPU sample(s): 741 IN-POLL (real blocking), 4615 idle-park
  NOTE: idle-park = a worker with no work parked between polls (itself a futex/condvar
        wait); it is NOT blocking. Only IN-POLL samples are a task blocked mid-poll.

  IN-POLL blocking by task spawn location:
      580 (78.3%) — examples/metrics-service/src/axum_traced.rs:246:32
      ...
```

**Caveat (documented in code + report):** an async lock wait (`lock().await` returning `Pending`) ends the poll, so its subsequent off-CPU time is tagged between-polls (idle), not in-poll. The in-poll bucket captures synchronous blocking but not `.await`-style async lock waits, so the "idle-park" label shouldn't itself be over-trusted.

### Both analysis paths fixed (per maintainer note)

`analyzeTraces` has two paths that independently compute the off-CPU grouping: single-file → in-process `accumulateTrace`; directory → `analyzeWorkerMain` subprocess → `mergePartial`. Patching only one would leave the other silently wrong. Both now go through a **shared `groupOffCpuSamples` helper** and converge on the same `finalizeAccumulator` result shape; verified to produce **identical** output on the demo trace (single-file vs directory: same 741/4615 split, same per-location attribution).

New result fields (`inPollGroups`, `offCpuInPollCount`, `offCpuIdleCount`, `inPollByLoc`) are documented in the `dial9-trace-analysis` SKILL.md schema; `schedGroups` is kept unchanged for backward compatibility.

## Tests

- `test_trace_analysis.js`: **+4 checks** — `inPoll` flag matches poll attachment, off-CPU split is exhaustive, mid-poll-wake adjustment (drives the branch the demo trace doesn't), and gap-wake non-adjustment. 65 → 69 checks, all green.
- `test_all_skills_snippets.js`: 76 passed, 0 failed (the schema-contract check verifies the new fields are documented + correctly shaped).
- `test_directory_parse.js` (subprocess merge path) and `test_directory_scale.js`: green.

Note: a pre-existing, unrelated failure in `dial9-red-flags/scripts/red_flag_scan.js:159` reproduces identically on `main` and is out of scope for this PR.

### DIFF
--- a/dial9-viewer/skills/dial9-toolkit/scripts/analyze.js
+++ b/dial9-viewer/skills/dial9-toolkit/scripts/analyze.js
@@ -582,7 +582,7 @@ function reportAnalysis(a, label) {
     // bucket captures synchronous blocking but not .await-style async lock waits.
     console.log(`\n  Caveat: an async lock wait (lock().await) ends the poll, so its off-CPU time`);
     console.log(`          is counted as idle-park, not in-poll. The IN-POLL bucket captures`);
-    console.log(`          synchronous blocking but not .await-style async lock waits.`);
+    console.log(`          synchronous blocking but not .await-style async lock waits. To see async lock wait, consider enabling task dumps`);
   }
 
   // ── CPU hotspots ──
--- a/dial9-viewer/skills/dial9-toolkit/scripts/analyze.js
+++ b/dial9-viewer/skills/dial9-toolkit/scripts/analyze.js
@@ -70,6 +70,54 @@ function diagnoseStack(frames) {
   return { syscall, blocker, trigger, classified };
 }
 
+/**
+ * Group off-CPU (scheduling) samples, splitting real blocking from idle parking.
+ *
+ * An off-CPU sample taken *inside a poll* means a task voluntarily blocked
+ * mid-poll — that is real blocking. An off-CPU sample taken *between polls*
+ * means a worker parked because it had no work — that is idle, even though the
+ * park is itself a futex/condvar wait at the kernel level. Reporting both as
+ * "blocking calls" makes a mostly-idle runtime look 100% blocked on a futex.
+ *
+ * attachCpuSamples sets sample.inPoll (true when the sample landed within a
+ * poll). We split on that flag, deduplicate the in-poll stacks for the blocking
+ * report, and attribute them per spawn location.
+ *
+ * Caveat: an async lock wait (lock().await returning Pending) ends the poll, so
+ * its subsequent off-CPU time is tagged between-polls (idle), not in-poll. The
+ * in-poll bucket captures synchronous blocking but not .await-style async lock
+ * waits, so the "idle" bucket should not itself be over-trusted.
+ *
+ * Shared by both analysis paths (accumulateTrace and analyzeWorkerMain) so they
+ * cannot drift apart.
+ *
+ * @param {Array} offCpu - off-CPU samples (source === 1), each with .inPoll/.spawnLoc set
+ * @param {Map} callframeSymbols
+ * @returns {{ inPollGroups: Array, inPollCount: number, idleCount: number, inPollByLoc: Array<{loc: string, count: number}> }}
+ */
+function groupOffCpuSamples(offCpu, callframeSymbols) {
+  const inPoll = [];
+  let idleCount = 0;
+  const byLoc = new Map();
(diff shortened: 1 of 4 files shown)


dial9-viewer/skills/dial9-toolkit/scripts/analyze.js
```diff
@@ -479,15 +551,38 @@ function reportAnalysis(a, label) {
...
+        if (syscall) console.log(`         syscall:    ${syscall.display}`);
+        if (trigger) console.log(`         called from: ${trigger.display}`);
+      }
     }
+    // Caveat: an async lock wait (lock().await returning Pending) ends the poll,
+    // so its subsequent off-CPU time is tagged idle, not in-poll. The in-poll
+    // bucket captures synchronous blocking but not .await-style async lock waits.
+    console.log(`\n  Caveat: an async lock wait (lock().await) ends the poll, so its off-CPU time`);
+    console.log(`          is counted as idle-park, not in-poll. The IN-POLL bucket captures`);
+    console.log(`          synchronous blocking but not .await-style async lock waits.`);
```
rcoh:
```suggestion
    console.log(`          synchronous blocking but not .await-style async lock waits. To see async lock wait, consider enabling task dumps`);
```

rcoh (approved):

PR #515, opened 2026-06-10
dial9-viewer: build.rs: Generate relative skill includes

cemeyer:
Generated skill metadata used absolute paths for files emitted by the build script.  But this is not canonical and breaks in some build environments.  Generate `include_str!` references to canonical `concat!(env!(\"OUT_DIR\"), ...)` (or `CARGO_MANIFEST_DIR`) instead.

### DIFF
--- a/dial9-viewer/build.rs
+++ b/dial9-viewer/build.rs
@@ -45,7 +45,7 @@ fn main() {
             let body = strip_frontmatter(&skill_md);
 
             // Collect all files in the skill directory (recursively)
-            let mut files: Vec<(String, String)> = Vec::new(); // (relative_path, absolute_path)
+            let mut files: Vec<(String, RelReference)> = Vec::new(); // (relative_path, src_rel)
             collect_files(&dir_path, &dir_path, &mut files);
 
             skills.push(SkillInfo {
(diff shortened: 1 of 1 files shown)


dial9-viewer/build.rs
```diff
@@ -133,22 +137,54 @@ fn main() {
...
+            res.strip_prefix(&src_dir)
+                .unwrap()
+                .to_path_buf()
+                .to_string_lossy()
+                .to_string(),
+        )
+    } else {
+        RelReference::OutRel(
+            res.strip_prefix(&out_dir)
+                .unwrap()
```
rcoh, line 178: this will panic if `out_dir` contains a symlink; I think the right behavior is to canonicalize all three paths; (res, src_dir, and out_dir)

rcoh (commented):
thanks for the fix! One small comment inline

dial9-viewer/build.rs
cemeyer: Thanks, will fix. Do you prefer pushing fixup commits until review is done or should I force-push an amended commit?

rcoh: Usually fix commits are nicer. We squash on merge and the fix commits make it easy to review changes

But for a tiny or like this either is totally fine

cemeyer: Went ahead and force pushed the amended change, hope that's ok.

Edit: ah, sorry, github's UI didn't show me your comment.  Will keep that in mind next time :-).

rcoh (approved):

cemeyer:
I'm afraid this still isn't quite right; you may want to preemptively back it out.  Some of the canonicalized files are outside of either prefix.

Edit: or see #518.

PR #518, opened 2026-06-11
Simplify path handling in dial9-viewer build script

dtolnay:
**dial9-viewer/build.rs** is implemented in a roundabout way where it constructs absolute paths, runs `std::fs::canonicalize` on them, and afterward reverse engineers at runtime whether the resulting canonicalized paths are located inside the crate's source directory or cargo's generated code directory.

For example consider this use of `resolve_path`. This path is simply always going to be inside of OUT_DIR and is never going to be inside of CARGO_MANIFEST_DIR, except in the case that the environment has set CARGO_TARGET_DIR in such a way that it is a subdirectory of dial9-viewer's manifest directory, in which case things go wrong.

The whole thing can be replaced more simply by `RelReference::OutRel("dial9-setup-SKILL.md")`, which is the value that `resolve_path` _should_ always return on the original path expression after doing all the canonicalizing and starts_with-ing and strip_prefix-ing.

https://github.com/dial9-rs/dial9/blob/b9c2905de6068b3ae796f0e68d306cf2322916c9/dial9-viewer/build.rs#L66

https://github.com/dial9-rs/dial9/blob/b9c2905de6068b3ae796f0e68d306cf2322916c9/dial9-viewer/build.rs#L163-L184

For every use of `resolve_path`, the structure of the build script means that it is always known whether the argument path is located in the source directory or the generated code directory, without needing to inspect the filesystem further by canonicalizing the path.

This PR deletes `resolve_path` and switches to relative paths throughout the rest of the implementation, deferring concatenating them onto CARGO_MANIFEST_DIR or OUT_DIR only for the few operations which read from the filesystem using `std::fs`.

### DIFF
--- a/dial9-viewer/build.rs
+++ b/dial9-viewer/build.rs
@@ -1,6 +1,6 @@
 use std::env;
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 /// Generates files in OUT_DIR for the new Agent Skills directory structure:
 ///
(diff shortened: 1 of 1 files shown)


cemeyer:
No objection from me.

rcoh (approved):
Lgtm!

Issue #356, opened 2026-05-05
dial9 crate structure v2


rcoh:
## Problem Statement

`dial9-tokio-telemetry` is a monolithic crate that bundles the event bus (buffer, flush loop, writer), Tokio runtime hooks, CPU profiling orchestration, tracing layer, memory profiling, background pipeline, S3 upload, and analysis into a single package. This makes it impossible to use dial9's core recording infrastructure without Tokio, prevents `dial9-perf-self-profile` from operating as a standalone Source, and forces users to depend on a crate named after a specific runtime even when they only need the event recording API.

## Solution

Split the workspace into focused crates with clear responsibilities:

- **`dial9`** — the facade crate (library + CLI binary). The one dependency most users add. Re-exports everything, owns feature gates, wires sources to the pipeline.
- **`dial9-core`** — the event bus. Buffer, flush loop, writer, Handle, Source trait, pipeline traits. Runtime-agnostic.
- **`dial9-tokio-telemetry`** — Tokio integration only. Runtime hooks, TracedFuture, spawn, task tracking.
- **`dial9-perf-self-profile`** — Linux system profiling as a standalone Source + segment processor (symbolization).

Users switch from `dial9-tokio-telemetry = "0.3"` to `dial9 = "0.4"` and get a cleaner, more composable API.

## User Stories

1. As a library author, I want to record custom events into a dial9 trace without depending on Tokio, so that my library works with any async runtime.
2. As an application developer, I want to add `dial9` to my Cargo.toml and get everything working, so that I don't need to figure out which internal crate to depend on.
3. As an application developer, I want to write `#[dial9::main]` instead of `#[dial9_tokio_telemetry::main]`, so that my code uses the canonical crate name.
4. As an application developer, I want to call `dial9::spawn(future)` to get an instrumented task, so that I don't need to manage handles manually.
5. As an application developer, I want to call `dial9::record_event(my_event)` without passing a handle, so that recording is ergonomic from anywhere in my code.
6. As an application developer, I want to pass an explicit handle via `handle.record_event(my_event)` for testability and multi-runtime scenarios.
7. As a non-Tokio runtime integrator, I want to implement the `Source` trait from `dial9-core` and plug my runtime's events into the event bus.
8. As a non-Tokio runtime integrator, I want to call `dial9_core::set_tl_handle(handle)` from my runtime's thread-start hook so that the implicit `record_event` API works.
[...]

PR #519, opened 2026-06-11
refactor(crate structure v2): Split RecordHandle from TelemetryHandle

### LINKED ISSUE #509: Decouple in place: split TelemetryHandle

jv1i:
closes #509

## Summary

Prep for #356 step 1. Before the recording engine can move to a runtime-agnostic
crate, it has to stop naming Tokio types. This does that in place, no public API break yet.

Two changes:

The handle is split into `RecorderHandle` (agnostic record/control) and `TelemetryHandle` (a `Deref` newtype that adds Tokio spawn). `current().spawn()` and `.enable()` still work unchanged.

Queue sampling + runtime metadata, which lived in `SharedState.contexts` and the flush loop, is now a single `TokioRuntimeSource`. `SharedState` drops the `contexts` field so it no longer names a Tokio type, and the flush loop just drains `Source`s. The runtime registry moves to `TelemetryGuard` (the Tokio layer), which is where `build_and_attach` finds it.

### Reviewer notes

RecorderHandle has to be pub (it's TelemetryHandle's Deref target) but isn't used externally yet. It becomes core's Handle later.

I considered using a runtime-specific trait rather than Source but at this point it was hard to justify.

Instead of summing in TelemetryGuard I considered independent per-runtime events, which we might want to do eventually, but `analysis.rs` builds one global-queue timeline and looks up by timestamp, so independent events would need a discriminator on the event (and some analysis changes). Preserving behavior for now, just bringing up in case it's something we want to support later.

### Next up

Create `dial9-core`, move the foundation (primitives, event types, collector, buffer, Source). Then a second wave with the Tokio-tied items, those might require some discussion.


rcoh:
we can rename it later but can we make sure that `TelemetryHandle` ends up being something more specific like `Dial9TokioRuntimeHandle` or something?

rcoh (approved):

PR #526, opened 2026-06-12
Fix: detect buckets with no prefix (date partitions at root)

### LINKED ISSUE #471: improve handling of buckets w/ no prefix
right now, the viewer will try to use a date as the prefix, it has a hard time when there are no prefixes, you have to manually set the prefix to the empty string

rcoh:
Fixes #471

When a bucket's root children are all date partitions (e.g. `2026-06-11/`, `2026-06-12/`) it means there is no key prefix — the trace data starts directly at the date layer. Previously these dates were offered as prefix suggestions, which is incorrect.

This change:
- Adds `isDateLayer()` detection that checks if all root children match `YYYY-MM-DD`
- When detected, sets the prefix to empty with a placeholder indicating no prefix
- Only triggers when *every* child is a date (mixed listings still show suggestions)

**Testing:**
- Added `test_prefix_detection.js` covering date-only, single-date, mixed, empty, and non-date cases


dial9-viewer/ui/test_prefix_detection.js
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env node
```
jlizen, line 1: This script needs to be added to `scripts/e2e-trace-tests.sh`

dial9-viewer/ui/test_prefix_detection.js
```diff
@@ -0,0 +1,92 @@
...
+
+"use strict";
+const fs = require("fs");
+const path = require("path");
+const vm = require("vm");
+
+const html = fs.readFileSync(path.join(__dirname, "index.html"), "utf8");
+
+function extract(name) {
+    const re = new RegExp("function " + name + "\\([\\s\\S]*?\\n    \\}\\n");
```
jlizen: This seems pretty brittle, should we extract these helpers into a pure module?

dial9-viewer/ui/index.html
```diff
@@ -293,6 +293,17 @@ <h1>dial9 Trace Browser</h1>
                 updateSearchReady();
                 return;
             }
+            // Issue #471: when the root children are all date partitions
+            // (YYYY-MM-DD/), the bucket has no key prefix — the trace data
+            // starts directly at the date layer. Don't offer the dates as
+            // prefix suggestions; the correct prefix is empty.
+            if (isDateLayer(prefixes)) {
```
jlizen:
```suggestion
            if (!serverHasPrefix && isDateLayer(prefixes)) {
```

jlizen (changes requested):

dial9-viewer/ui/index.html
rcoh: applied in db7a68f.

jlizen (approved):

PR #534, opened 2026-06-14
Per-process boot_id namespace isolation for shared trace directories

### LINKED ISSUE #532: Multiple processes sharing the same `dial9` directory causes runaway collapse
If multiple processes are sharing the same directory on the host, each background worker will see the others files.

The background worker will then attempt to upload the files from the other workers on host. This is a problem in and of itself, but it gets worse when there are large numbers of competeting processes:
1. The memory will balloon because each process will attempt to load potentially GB of files into memory each time it processes them
2. The disk usage is much larger than potentially expected
3. The CPU will balloon as it tries to compress and symbolize a huge amount of trace files

The fix is to namespace (probably by a long boot-id), the dial9 trace files produced by the rotating writer.
- [ ] unify s3 boot id and fs boot id

<summary>A detailed / AI assisted summary of the design is below</summary>
<details>

Namespace each process's trace files into a per-process subdirectory named by
`boot_id`:
```
{parent}/{boot_id}/{stem}.{index}.bin        # was {parent}/{stem}.{index}.bin
{parent}/{boot_id}/.lock                      # flock held for the process lifetime
```
- **`boot_id` = per-*process* identifier** (NOT the OS boot id). Format unchanged:
  `{4-alpha}-{pid}` — memorable and recognizable; sort order is intentionally not
  guaranteed (`ls -latr` for chronological). Uniqueness is guaranteed by the lock
  (see below), not by the id's entropy.
- **One owner.** `boot_id` is generated once at the managed-runtime layer
  (`TracedRuntime` construction) and defaults to the auto value; an optional
  `with_boot_id(...)` override lives on the runtime builder. It is injected into
  the disk layer by **rewriting `base_path`** to `{parent}/{boot_id}/{filename}`,
  so `RotatingWriter` and the worker stay `boot_id`-agnostic (they just see a
  deeper path). The S3 pipeline **inherits** the same `boot_id`, so the on-disk
  directory and the S3 key's `boot_id` component always match. `boot_id` is
  **removed from `S3Config`'s public builder** (breaking; acceptable for the
  unreleased 0.4.0).
- **`flock` advisory lock**, not an existence-check. The runtime holds
  `flock(LOCK_EX)` on `{boot_id}/.lock` for its whole lifetime. The kernel
  releases it automatically on process death (including crash/SIGKILL), so a peer
  that can acquire the lock knows the owner is dead. It also serves as the
  collision backstop: on the astronomically-rare live collision, bump and retry.
- **Unconditional** for the managed runtime — no opt-out. The failure mode is
  severe and a single process just gains one harmless subdir level.
- **Guarded, delete-only startup GC.** On startup a process reclaims dead peers'
  namespaces. A sibling dir is eligible only if (1) its name matches the
  `boot_id` shape `^[a-z]{4}-[0-9]+$` and isn't ours, (2) its `.lock` can be
  `flock`ed (owner dead), and (3) deletion unlinks only recognized files
  (`*.bin`, `*.bin.active`, `*.bin.gz`, `.lock`) then `rmdir`s — if anything
  unrecognized remains, `rmdir` fails and we leave the dir and log a warning.
  Never a recursive `rm -rf`. Reclamation logs at `INFO`, fail-closed at `WARN`.
- **`single_file()` is dropped.** It implied "this exact path," which namespacing
  makes impossible. Replaced by a `no_rotation` builder option (still namespaced).
  Path discovery: the guard exposes `boot_id()` and `trace_dir()`
  (= `{parent}/{boot_id}`); for tests/custom pipelines, `WriteBackProcessor`
  optionally reports each written path over an mpsc channel
  (`.write_back_observed(tx)`), letting tests `await` the finalized `.bin.gz`
  path deterministically instead of scanning.
## Considered alternatives
- **UUID `boot_id` instead of the lock.** Solves only entropy, not liveness; the
  lock additionally gives orphan detection. UUID also sorts randomly.
- **`boot_id` in the filename** (`trace.{boot_id}.N.bin`) instead of a subdir.
  Forces the worker to parse-and-filter by `boot_id` (a matching bug reintroduces
  cross-talk) and complicates eviction; a subdir isolates by simply pointing the
  scan at a different directory and is atomically removable.
- **Lock as a sibling** (`{parent}/.{boot_id}.lock`) instead of inside the dir.
  Cheaper liveness scan, but splits a namespace into two objects that can
  desync.
- **Full orphan *recovery* (upload dead peers' segments) now.** Deferred: it hits
  the cross-process symbolization-correctness problem (a recovering process can't
  symbolize a dead peer's trace with its own maps) and is a design surface of its
  own. This change ships delete-only GC; recovery is a follow-up that reuses the
  same `flock` liveness primitive.
## Consequences
- Fixes the cross-process runaway and the (incorrect) cross-process symbolization.
- Removes today's *accidental* cross-restart upload (a restarting process used to
  pick up and upload the previous run's leftover segments). Delete-only GC bounds
  disk but does **not** preserve a dead run's data — that awaits the recovery
  follow-up.
- `dial9 serve --local-dir` still works: `LocalBackend` recurses and namespace
  dirs aren't dot-prefixed. `collect_files` should gain a dotfile/extension skip
  so `.lock` doesn't surface as a fake trace.
</details>

skanjalkar:
When multiple processes share the same trace directory, each process's background worker loads and processes all peers' files, causing memory/CPU/disk runaway. This means every background worker tries to upload every file to S3, not just its own.

This PR fixes it by namespacing trace segments into per-process subdirectories keyed by a `boot_id` (4 random lowercase letters + the PID, like `qmxz-48291`). At config build time we rewrite the base path from `{trace_dir}/trace.bin` to `{trace_dir}/{boot_id}/trace.bin`, so each process only ever sees its own files. To detect when a peer has died, each process holds an `flock` on `{boot_id}/.lock` for its entire lifetime the kernel releases this automatically on crash or SIGKILL. 

On startup, we check sibling directories and clean up any whose lock we can acquire (meaning the owner is dead), but only if the directory contains recognized files and nothing unexpected. The viewer's local file listing also skips dotfiles now so `.lock` doesn't show up as a trace.

The flock has a runtime cost of a single syscall at startup (this is the one I am most concerned with multi processes dial9 usecases) and then the kernel just holds a flag on the fd. The only work that happens is the GC scan at startup (one readdir + one flock probe per sibling directory), this should be in sub millis - i haven't had a chance to benchmark it but open to options.

I considered a few alternatives before landing here. I first looked at using UUIDs for the directory name, but that gives no way to tell if the owner is still alive so there's no path to cleaning up orphaned directories. Then I looked at PID files with /proc checks which is simpler, but PID reuse means you can mistakenly think a dead process is still running. The simplest option would be to just isolate into subdirectories without any cleanup, but then dead dirs accumulate until someone manually intervenes or the container restarts (maybe a cron job can do this but I thought that was overkill). Maybe a time-based TTL cleanup would be nice and cross-platform but I was worried about slow processes.

Closes #532


dial9-tokio-telemetry/src/background_task/boot_id.rs
```diff
@@ -0,0 +1,361 @@
+//! Per-process namespace isolation for shared trace directories.
+//!
+//! Each process writes to `{trace_dir}/{boot_id}/` so background workers
+//! never cross-process. Liveness is tracked via `flock(LOCK_EX)` on
+//! `{boot_id}/.lock`; dead peers are GC'd at startup.
```
rcoh, line 5: I think we probably want an option to configure whether or not we GC at startup. Or at least make sure there is a nice local flow where we let you create multiple trace files from multiple runs

dial9-tokio-telemetry/src/telemetry/writer.rs
```diff
@@ -457,6 +470,19 @@ impl<M: WriterMode> SegmentWriter<M> {
         &self.base_path
     }
 
+    /// Per-process namespace subdirectory name, if active.
```
rcoh: Seems like it would probably be more useful to return the boot directory directly?

rcoh (commented):
Probably needs to be documented. I'm surprised no tests were impacted by this

dial9-tokio-telemetry/src/telemetry/writer.rs
skanjalkar: i think trace_dir already gives us that, this is just for logging (also doc comments in the code to clarify that)

While I was in here I noticed the `local boot_id` and the `S3Config boot_id` are generated independently, so they don't match: a segment in `qmxz-481/` locally uploads under a different prefix in S3. I am not sure if I should address them here in this PR or create another PR since that's a rather big change (something like 15 tests)? 

Unifying means removing .boot_id() from the public S3Config builder (breaking, touches tests and the example), so happy to do it as a follow-up

rcoh: Yeah we can do it as a follow up. We should either unify them or name them differently. We can put a check box on the original ticket

dial9-tokio-telemetry/src/background_task/boot_id.rs
```diff
@@ -0,0 +1,464 @@
...
+
+use std::io;
+use std::path::{Path, PathBuf};
+
+use crate::primitives::fs;
+
+// Startup-only operations that don't need shuttle fault-injection.
+use std::fs as stdfs;
```
rcoh: there's no reason not to make this shuttle-covered. it will definitely make it easier to test various failure modes

dial9-tokio-telemetry/src/background_task/boot_id.rs
```diff
@@ -0,0 +1,464 @@
...
+        if !meta.is_dir() {
+            continue;
+        }
+        let Some(name) = path.file_name().and_then(|n| n.to_str()) else {
+            continue;
+        };
+        if name == own_boot_id || !is_valid_boot_id(name) || is_lock_held(&path) {
+            continue;
+        }
+        let _ = try_remove_namespace(&path);
```
rcoh: we should log errors here

dial9-tokio-telemetry/src/background_task/boot_id.rs
```diff
@@ -0,0 +1,464 @@
...
+    }
+}
+
+fn try_remove_namespace(dir: &Path) -> io::Result<()> {
+    let entries: Vec<_> = stdfs::read_dir(dir)?.filter_map(|e| e.ok()).collect();
+
+    let stem = infer_stem(&entries).unwrap_or_else(|| "trace".to_string());
```
rcoh: I'm a little confused about the fallback to trace here. From other places, I think we can probably just completely GC this stem parsing flow

dial9-tokio-telemetry/src/background_task/boot_id.rs
```diff
@@ -0,0 +1,464 @@
...
+
+    if entries.iter().any(|e| !is_safe_to_delete(e, &stem)) {
+        return Ok(());
+    }
+
+    for entry in &entries {
+        let _ = fs::remove_file(&entry.path());
```
rcoh: this returns `io::Result` but we are discarding all these errors

dial9-tokio-telemetry/src/background_task/boot_id.rs
```diff
@@ -0,0 +1,464 @@
...
+    }
+    let name = entry.file_name();
+    let Some(name) = name.to_str() else {
+        return false;
+    };
+    name == ".lock" || parse_segment_artifact(name, stem).is_some()
+}
+
+/// Infer the trace stem from any recognized file in the directory.
+fn infer_stem(entries: &[stdfs::DirEntry]) -> Option<String> {
```
rcoh: is this actually required? I don't really know what looking at the stem buys us

dial9-tokio-telemetry/src/background_task/boot_id.rs
```diff
@@ -0,0 +1,464 @@
...
+    }
+}
+
+#[cfg(not(unix))]
+fn is_lock_held(_namespace_dir: &Path) -> bool {
+    // No flock — conservatively assume alive so GC never runs.
+    true
+}
+
+fn is_safe_to_delete(entry: &stdfs::DirEntry, stem: &str) -> bool {
```
rcoh: I think we can consider the entire trace directory as deletable; if you put your own stuff in the directory its probably fine to delete it. We should be looking at the trace directory and the presence of a `.lock` as the guarding item rather than trying to be clever

dial9-tokio-telemetry/src/background_task/boot_id.rs
```diff
@@ -0,0 +1,464 @@
...
+        // Acquired — owner is dead. Release immediately.
+        unsafe { libc::flock(file.as_raw_fd(), libc::LOCK_UN) };
+        false
+    } else {
+        true
+    }
+}
+
+#[cfg(not(unix))]
+fn is_lock_held(_namespace_dir: &Path) -> bool {
```
rcoh: I'm a little concerned about this non-unix behavior. we should at least document it.

dial9-tokio-telemetry/src/background_task/boot_id.rs
```diff
@@ -0,0 +1,464 @@
...
+
+    let lock_path = namespace_dir.join(".lock");
+    let file = match stdfs::OpenOptions::new().read(true).open(&lock_path) {
+        Ok(f) => f,
+        Err(_) => return false,
+    };
+    let rc = unsafe { libc::flock(file.as_raw_fd(), libc::LOCK_EX | libc::LOCK_NB) };
+    if rc == 0 {
+        // Acquired — owner is dead. Release immediately.
+        unsafe { libc::flock(file.as_raw_fd(), libc::LOCK_UN) };
```
rcoh: these both need safety comments

dial9-tokio-telemetry/src/background_task/boot_id.rs
```diff
@@ -0,0 +1,464 @@
...
+
+    fs::create_dir_all(namespace_dir)?;
+    let lock_path = namespace_dir.join(".lock");
+    let file = stdfs::OpenOptions::new()
+        .create(true)
+        .write(true)
+        .truncate(false)
+        .open(&lock_path)?;
+
+    let rc = unsafe { libc::flock(file.as_raw_fd(), libc::LOCK_EX | libc::LOCK_NB) };
```
rcoh: needs safety comment

dial9-tokio-telemetry/src/background_task/boot_id.rs
```diff
@@ -0,0 +1,464 @@
...
+        && alpha.bytes().all(|b| b.is_ascii_lowercase())
+        && !pid.is_empty()
+        && pid.bytes().all(|b| b.is_ascii_digit())
+}
+
+/// Acquire an exclusive advisory lock on `{namespace_dir}/.lock`.
+/// Returns the file handle — lock is held until the handle is dropped.
+/// Kernel releases automatically on process death (including SIGKILL).
+#[cfg(unix)]
+pub(crate) fn acquire_namespace_lock(namespace_dir: &Path) -> io::Result<stdfs::File> {
```
rcoh: I recommend making a small inline `flock` module and doing the cfg on the entire module instead of function by function, its a bit easier to read

dial9-tokio-telemetry/src/background_task/boot_id.rs
```diff
@@ -0,0 +1,464 @@
...
+use std::path::{Path, PathBuf};
+
+use crate::primitives::fs;
+
+// Startup-only operations that don't need shuttle fault-injection.
+use std::fs as stdfs;
+
+use crate::background_task::sealed::parse_segment_artifact;
+
+pub(crate) fn generate_boot_id() -> String {
```
rcoh, line 18: this function is verbatim duplicated with the s3 one, can we unify them?

dial9-tokio-telemetry/src/background_task/boot_id.rs
skanjalkar: will do

dial9-tokio-telemetry/src/background_task/boot_id.rs
skanjalkar: will do

dial9-tokio-telemetry/src/background_task/boot_id.rs
skanjalkar: ah missed this. updating.

rcoh (approved):

PR #535, opened 2026-06-15
refactor!: collapse handles into Dial9Handle / Dial9TokioHandle

jv1i:
Follow-up to #519, part of #356.

## What this is for

Toward #356, the recording handle needs to be Tokio-free so it can move to `dial9-core`. 
#519 carved the agnostic `Dial9Handle`, leaving it reachable through a `Deref` shim on `TelemetryHandle` plus two overlapping spawn handles on the Tokio side (`TelemetryHandle`, `RuntimeTelemetryHandle`). 
This PR finishes the handle side of the split. After it, we have two handles:

- `Dial9Handle`:  record and control, next up moves into `dial9-core` as `Handle`.
- `Dial9TokioHandle`: spawn only, stays in `dial9-tokio-telemetry` with this name.

Recording becomes a method `handle.record_event(event)` (see the explicit-handle API from #356's user stories).

## Breaking changes

- `TelemetryHandle` and `RuntimeTelemetryHandle` are gone.
- `guard.handle()` returns `Dial9Handle` (record/control), not a spawn handle.
- Spawn via `Dial9TokioHandle::current()`, `guard.tokio_handle(&runtime)`, or the handle from
  `trace_runtime().build()`.
- `record_event(event, &handle)` is now `handle.record_event(event)`.

## Next up

Create `dial9-core`, `Dial9Handle` moves there, implicit `dial9::record_event(event)` (thread-local, no handle) and public `set_tl_handle` arrive with the `dial9` facade.


dial9-tokio-telemetry/examples/production_use.rs
```diff
@@ -12,10 +12,10 @@
 //!
 //! - Setting [`.enabled(false)`](dial9_tokio_telemetry::DiskConfigBuilder::enabled) on the config builder produces a
 //!   pass-through config: the `#[main]` macro builds a plain, unmodified tokio runtime with zero dial9 overhead.
-//!   [`TelemetryHandle::current()`](dial9_tokio_telemetry::telemetry::TelemetryHandle::current) returns an inert
+//!   [`Dial9TokioHandle::current()`](dial9_tokio_telemetry::telemetry::Dial9TokioHandle::current) returns an inert
 //!   handle, and `handle.spawn` falls through to `tokio::spawn`, so application code does not need branches.
 //! - Alternatively, you can install dial9 but leave recording disabled at runtime via the handle's
-//!   [`disable()`](dial9_tokio_telemetry::telemetry::TelemetryHandle::disable). The runtime hooks are installed
+//!   [`disable()`](dial9_tokio_telemetry::telemetry::Dial9TokioHandle::disable). The runtime hooks are installed
```
rcoh: I think this is `Dial9Handle::disable()`

rcoh (approved):

PR #536, opened 2026-06-15
feat: Manual trigger for pipeline

### LINKED ISSUE #469: add ability to trigger a dump on an application trigger

Fluzko:
closes #469 

## Summary
  Adds an on-demand mode for the dial9 worker. Today the worker processes every sealed segment continuously. Wiring a trigger flips the *same* pipeline into on-demand operation: segments keep accumulating in the ring, and the pipeline only runs when the application asks for a dump, on a panic, an idle-ratio drop, a `/dump` handler, etc. You pay processing/upload cost only when something is worth keeping.

### API

  ```rust
  let (control, rx) = dump::trigger();

  TracedRuntime::builder()
      .with_s3_uploader(cfg)   // any pipeline you'd run continuously
      .with_trigger(rx)        // only changes *when* it runs
      .build_and_start(..)?;

  let receipt = control
      .dump_time_range(lookback, lookforward)
      .with_metadata("reason", "idle-ratio-drop")
      .await?;
```

  - With_trigger(rx) is orthogonal to pipeline selection.
  - Dump_current_data() / dump_time_range(lookback, lookforward) dispatch
  immediately; awaiting only retrieves the `DumpReceipt`.
  - Both sides are best-effort: a window wider than the ring retained captures
  what survived, with no error and no effect on the live stream.

 ### S3 discoverability

  Dumped objects land in the normal key layout, each tagged with dump-id user
  metadata. The S3 stage also writes one manifest per dump at
  {prefix}/dumps/{dump_id}.json listing the keys it produced. Manifest
  presence means success - a failed dump writes none.

  ### Work done

  - dump module: DumpControl, DumpRun, DumpReceipt, DumpId, trigger()
  - Builder: with_trigger(rx) plumbs the receiver to the worker
  - Worker: on-demand drain loop, matching segments by [creation, seal] span
  overlap against active dump windows
  - S3: per-object dump-id tag + per-dump manifest
  - Runnable example (examples/on_trigger_dump.rs) and integration tests

### CODE REVIEW

File: dial9-tokio-telemetry/examples/on_trigger_dump.rs [hunk 1]
```diff
@@ -0,0 +1,122 @@
...
+
+#[dial9_tokio_telemetry::main(config = || {
+    let _ = std::fs::remove_dir_all(TRACE_DIR);
+    let _ = std::fs::create_dir_all(TRACE_DIR);
+    let trace_path = format!("{TRACE_DIR}/trace.bin");
+
+    // Create the trigger pair. `rx` is moved into the builder below;
+    // `control` is published so the body can dump on demand.
+    let (control, rx) = dump::trigger();
+    let _ = DUMP.set(control);
```
Fluzko: This is what we agreed during the plan, but now that I see in action it's probably not as handy as I thought it would be. I was thinking about leaning on `TelemetryHandle` instead of having to manually store it and move it around.

What about this API?:

```rust
Dial9Config::builder()
          .other_config()
          .with_runtime(move |r| r
              .with_task_tracking(true)
              .with_custom_pipeline(|p| p.gzip().write_back())
              .with_trigger()) // << no rx

// ...

async fn main() {
      let control = handle
          .dump_control()
          .expect("on-demand mode enabled");
}
```

dial9-tokio-telemetry/examples/on_trigger_dump.rs
```diff
@@ -0,0 +1,122 @@
...
+    // Something noteworthy happened - dump what the ring holds, tagging the
+    // dump with a reason. The dump dispatches immediately; awaiting only
+    // retrieves the receipt.
+    println!("incident detected, dumping buffered segments ...");
+    let control = DUMP.get().expect("DumpControl published by config closure");
+    let receipt = control
+        .dump_current_data()
+        .with_metadata("reason", "demo")
+        .await
+        .expect("dump receipt");
```
rcoh: since this is an example, an example that is slightly more realistic may be useful (e.g. having a background thread decide to dump).

dial9-tokio-telemetry/tests/dump_trigger.rs
```diff
@@ -0,0 +1,54 @@
+//! Builder wiring for on-trigger pipeline runs.
+
+use dial9_tokio_telemetry::background_task::s3::S3Config;
+use dial9_tokio_telemetry::dump::{self, DumpError};
+use dial9_tokio_telemetry::telemetry::{Disk, DiskWriter, TracedRuntime};
+
+/// `with_trigger` is available in every pipeline state (compile check).
+#[allow(dead_code)]
+fn with_trigger_compiles_in_all_pipeline_states() {
```
rcoh: what happens if two dumps get triggered concurrently?

dial9-tokio-telemetry/tests/dump_trigger.rs
```diff
@@ -0,0 +1,54 @@
...
+    let (_control, rx) = dump::trigger();
+    let _custom = TracedRuntime::builder()
+        .with_custom_pipeline::<_, Disk>(|p| p.gzip().write_back())
+        .with_trigger(rx);
+}
+
+/// A trigger without a configured pipeline never spawns the worker; the
+/// receiver is dropped and every dump resolves `WorkerStopped`.
+#[test]
+fn trigger_without_pipeline_resolves_worker_stopped() {
```
rcoh, line 31: lets add a version of the s3 integration test that triggers a dump in addition to the continuous upload one

rcoh (commented):
Overall this looks good. I like your proposed API to avoid needing to hand an `rx` in. The handling of duplicate invocation of dump invocations also seems reasonably but I wonder if we want to build some debouncing or other clean method.

At the very least, we should add a test of the case where we have two concurrent calls to `dump_current_data()` (sorry if one exists that I missed)

I imagine there are two ways customers will use this:
1. a background thread that checks some stat every 500ms and triggers a dump
2. A application timing based approach where if a request took >p999 or something along those lines it triggers a dump.

In the case of option 2, you'd probably want some coordination or a cooldown period or something? I think its probably ok for now to just document the current behavior but I suspect folks may re-implement the same set of primitives if we don't provide them.

Fluzko:
> Overall this looks good. I like your proposed API to avoid needing to hand an `rx` in. The handling of duplicate invocation of dump invocations also seems reasonably but I wonder if we want to build some debouncing or other clean method.
> 
> At the very least, we should add a test of the case where we have two concurrent calls to `dump_current_data()` (sorry if one exists that I missed)
> 
> I imagine there are two ways customers will use this:
> 
>     1. a background thread that checks some stat every 500ms and triggers a dump
> 
>     2. A application timing based approach where if a request took >p999 or something along those lines it triggers a dump.
> 
> 
> In the case of option 2, you'd probably want some coordination or a cooldown period or something? I think its probably ok for now to just document the current behavior but I suspect folks may re-implement the same set of primitives if we don't provide them.

I agree on the debouncing, I'll implement it as part of this PR.
Regarding how users would actually use the feature, based on your hunch about likely usage, I think this is something we can defer and implement if needed. Both options come down to extending the API with the right knobs, so I'd wait until we get an actual request for them.
That said, neither is a complex addition, so if you feel strongly about it, I'm happy to include them now.

dial9-tokio-telemetry/tests/dump_trigger.rs
Fluzko: Documented it

dial9-tokio-telemetry/examples/on_trigger_dump.rs
```diff
@@ -0,0 +1,132 @@
...
+            .with_task_tracking(true)
+            .with_custom_pipeline(|p| p.gzip().write_back())
+            .with_trigger(|t| t.debounce(Duration::from_secs(30))))
+        .build_or_disabled()
+})]
+async fn main() {
+    let handle = TelemetryHandle::current();
+    // Reach the dump control through the ambient handle, the runtime stashed
+    // it when `with_trigger` was configured.
+    let control = handle.dump_control().expect("on-demand mode enabled");
```
rcoh: in one place this api is called `trigger` and in the other it is called `dump_control`. we should unify those names.

Lets have it be something like:

`dump_on_demand`

and `handle.dump_control`

open to other suggestions

dial9-tokio-telemetry/examples/on_trigger_dump.rs
```diff
@@ -0,0 +1,132 @@
...
+        .max_total_size(10 * 1024 * 1024)
+        .rotation_period(Duration::from_millis(500))
+        .with_tokio(|t| { t.worker_threads(2); })
+        // The pipeline is whatever you would run continuously (here: gzip +
+        // write_back); `with_trigger(...)` only changes *when* it runs. The
+        // debounce gate folds a burst of re-trips into a single dump.
+        .with_runtime(|r| r
+            .with_task_tracking(true)
+            .with_custom_pipeline(|p| p.gzip().write_back())
+            .with_trigger(|t| t.debounce(Duration::from_secs(30))))
```
rcoh: its weird this is inside `with_runtime`. its not related to Tokio. I think it should be a top level setting?

dial9-tokio-telemetry/Cargo.toml
```diff
@@ -65,6 +65,7 @@ metrique = { version = "0.1.23", features = ["local-format"] }
 metrique-timesource = "0.1"
 dial9-macro = { workspace = true }
 shuttle = { version = "0.9.1", optional = true }
+ulid = "1"
```
rcoh, line 74: is there a reason you did `ulid` instead of `uuid`?

dial9-tokio-telemetry/src/dump.rs
```diff
@@ -0,0 +1,624 @@
+//! On-trigger pipeline runs.
+//!
+//! By default the background worker processes sealed trace segments
+//! continuously. Wiring a trigger flips the same pipeline into on-demand
+//! operation: segments keep accumulating in the ring (memory or disk), and
+//! the pipeline only runs when the application explicitly requests a dump.
+//!
+//! ```ignore
```
rcoh: can these be `no_run`? I don't like ignored examples in general because they go stale really easily

dial9-tokio-telemetry/Cargo.toml
Fluzko: Pretty much because of the timestamp, I would allow to sort dump ids, S3 object keys and manifest names. It's also used in the code to get a trigger timestamp from the dump id. pretty much a matter of handyness

dial9-tokio-telemetry/examples/on_trigger_dump.rs
Fluzko: so Tokio related knobs lives at `with_tokio()`, `with_runtime` are Dial9 related stuff not exactly about Tokio runtime.  So not sure if you still want to move it one layer on upper, since `with_custom_pipeline` which is what we are aiming, lives inside `with_runtime`

rcoh: Hmm...what is with_runtime for then? Is there a logical reason why something would be in our out?

I was thrown off by with_task_tracking which we should honestly just get rid of, I don't think there is a good reason to disable it

Fluzko: My understanding is that pre-startup config goes outside, so let's say everything needed by `#[main]` to bootstrap the app.
`with_runtime` hides the recorder config, what we capture and where it ships. The hard constraint is the pipeline `with_custom_pipeline` / `with_s3_uploader` carry the `Disk` `Memory` type marker, so they can't live top-level without leaking that generic across the whole config builder.

It's fair to challenge `with_task_tracking` also... I don't recall the reason about adding it, but seems an inoffensive knob, probably add more docs around it? add a default to avoid setting it to true every time?

rcoh (approved):

rcoh:
Tests still seem flaky

jv1i:
Quickly checking this from the perspective of https://github.com/dial9-rs/dial9/issues/356, and more precisely everything that's being moved and de-coupled from tokio in https://github.com/dial9-rs/dial9/pull/549.
There are some things we could do in this feature to make moving it into dial9-core easier (I'll still take care of the move, conflicts, etc.).

Asking about them, but we don't necessarily have to solve them in this PR:

- The trigger channel is `tokio::sync::mpsc`, it seems only the receive side needs to be async (the worker awaits it in the select), but the send side doesn't. So could we use something like async-channel instead, so it's not tied to tokio?
- Same for the oneshot receipt that `DumpFuture` awaits. Could we swap it for something like `futures-channel::oneshot`? (Had to do something similar with Notify in Fs -> event-listener in da41be1, though that one carried no payload.)

rcoh:
You can use tokio with only the sync feature which works on any runtime and only brings in the concurrency primitives

rcoh:
still flaky; the macos runner is super good at catching race conditions, especially related to atomics

rcoh:
You could try to move the test to `shuttle`?

PR #539, opened 2026-06-15
refactor: drop thread_roles from FlushContext / Source

jv1i:
Part of #356, prep for the `dial9-core` bus extraction.

### Summary

Removes producer-side worker attribution. CPU/sched samples emit `WorkerId::UNKNOWN`
plus the OS `tid`; worker attribution is reconstructed at analysis time from the `tid`
on park/unpark events. This drops the `thread_roles` map (a tokio worker concept) off
`Source`/`FlushContext`/`SharedState`, keeping the bus runtime-agnostic for `dial9-core`.

#### Changes

- Producer: drop `thread_roles` from `SharedState` and the thread-start/stop hooks,
  drop the now-dead `ThreadRole` enum. `CpuProfiler`/`SchedProfiler` emit `WorkerId::UNKNOWN`.
  `FlushContext` is back to `{ collector, drain_epoch }`.
- Rust analysis: (`detect_sampled_polls`) and **viewer** (`trace_parser.js`): rebuild
  `tid -> worker` from park/unpark, resolve each sample by `tid`. Fall back to the wire
  `worker_id` when the tid is unmapped (old traces whose park/unpark predate the `tid` field).

#### Behavior change

Blocking-pool samples now resolve to `UNKNOWN`, not `BLOCKING`: only worker threads emit
park/unpark, and a runtime-agnostic producer can't know tokio's blocking pool. Nothing downstream read the blocking bucket, so no report regresses. The `blocking_pool_tracking` example, whose only point was the now-gone 254 attribution, is removed.

#### Not in scope

The wire `CpuSample.worker_id` field stays (the backwards-compat fallback). #356 slates it
for removal but says this refactor need doesn't need to address it yet, I'm guessing we'll do it eventually as a wire-format change.


dial9-tokio-telemetry/src/telemetry/recorder/builder.rs
```diff
@@ -840,15 +840,21 @@ impl TelemetryCore {
         #[cfg(feature = "cpu-profiling")]
         {
             if let Some(ref config) = cpu_profiling {
-                match crate::telemetry::cpu_profile::CpuProfiler::start(config.clone()) {
+                match crate::telemetry::cpu_profile::CpuProfiler::start(
+                    config.clone(),
+                    shared.thread_roles.clone(),
```
rcoh: we can actually drop thread roles entirely. We now infer thread roles from the worker park / unpark events. We can put `-1` in there or something to preserve some level of backwards compat if we want. The UI already stopped reading thread role info I believe.

jv1i: got it, sorry I should've catched that better from #356.
I dropped it now but used `WorkerId::UNKNOWN` instead of `-1` but still maintaining backwards compat AFAICT. let me know if this is what you had in mind.

rcoh (approved):

dial9-tokio-telemetry/src/telemetry/analysis.rs
```diff
@@ -660,6 +660,10 @@ pub fn detect_sampled_polls(events: &[Dial9Event]) -> Vec<SampledPoll> {
 
     let mut polls: Vec<PollSpan> = Vec::new();
     let mut poll_starts: HashMap<WorkerId, (u64, u64, String)> = HashMap::new();
+    // CPU samples identify their thread by OS tid, not worker id. Park/unpark
+    // events pair tid with worker id, and a worker thread keeps one tid, so they
+    // give a stable tid → worker map to attribute samples by.
+    let mut tid_to_worker: HashMap<u32, WorkerId> = HashMap::new();
```
rcoh, line 666: this is fine for our tests but note that this isn't actually generally correct because of `block_in_place`

PR #540, opened 2026-06-15
refactor: setup dial9-core

jv1i:
prep for #510, part of #356.

### Summary

Scaffolding only. Creates the `dial9-core` crate and moves the leaf utilities (`primitives`, `rate_limit`, `sampling`) into it. 
`dial9-tokio-telemetry` now depends on `dial9-core` and re-exports them, so nothing else moves yet, the
bus (collector, flush loop, writer, Handle, Source) will land next.

These re-exports are `pub` in core so sibling crates share one shuttle shim, but
they're internal plumbing, not public API, so they're `#[doc(hidden)]`.

No breaking changes here, the moved modules are re-exported `pub(crate)` from `dial9-tokio-telemetry`,
its public API is unchanged. `dial9-core` is a new crate.

### DIFF
--- a/dial9-core/src/lib.rs
+++ b/dial9-core/src/lib.rs
@@ -1,0 +1,6 @@
+/// Cfg-gated concurrency primitives (std / shuttle).
+pub mod primitives;
+/// Per-call-site rate limiting for log lines.
+pub mod rate_limit;
+/// Geometric/Poisson sampling primitives (RNG, exponential draws).
+pub mod sampling;
--- a/dial9-core/src/lib.rs
+++ b/dial9-core/src/lib.rs
@@ -1,6 +1,12 @@
 /// Cfg-gated concurrency primitives (std / shuttle).
+///
+/// `pub` so sibling crates share one shuttle shim, but not part of the public
+/// API.
+#[doc(hidden)]
 pub mod primitives;
 /// Per-call-site rate limiting for log lines.
+#[doc(hidden)]
 pub mod rate_limit;
 /// Geometric/Poisson sampling primitives (RNG, exponential draws).
+#[doc(hidden)]
 pub mod sampling;
--- a/dial9-core/src/primitives.rs
+++ b/dial9-core/src/primitives.rs
@@ -8,21 +8,21 @@
 // ── std path (production) ───────────────────────────────────────────────────
 
 #[cfg(not(shuttle))]
-pub(crate) mod sync {
-    pub(crate) use std::sync::atomic;
-    pub(crate) use std::sync::mpsc;
+pub mod sync {
+    pub use std::sync::atomic;
+    pub use std::sync::mpsc;
     #[allow(unused_imports)]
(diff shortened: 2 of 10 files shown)


dial9-core/src/primitives.rs
```diff
@@ -8,21 +8,21 @@
 // ── std path (production) ───────────────────────────────────────────────────
 
 #[cfg(not(shuttle))]
-pub(crate) mod sync {
-    pub(crate) use std::sync::atomic;
-    pub(crate) use std::sync::mpsc;
+pub mod sync {
```
rcoh, line 11: interesting. these need to be `pub`?

dial9-core/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+/// Cfg-gated concurrency primitives (std / shuttle).
+pub mod primitives;
```
rcoh, line 6: if this needs to be `pub`, we should probably make it `#[doc(hidden)]`

dial9-tokio-telemetry/src/telemetry/recorder/handle.rs
```diff
@@ -143,6 +144,15 @@ impl Dial9Handle {
         })
     }
 
+    /// Record a custom event into the trace.
+    ///
+    /// Any type implementing [`dial9_trace_format::TraceEvent`] (typically via
+    /// `#[derive(TraceEvent)]`) works directly. No-op on a disabled handle or
+    /// when recording is paused.
+    pub fn record_event(&self, event: impl crate::telemetry::buffer::Encodable) {
```
rcoh, line 152: should this function be here or just on `dial9-core`?

examples/metrics-service/src/axum_traced.rs
```diff
@@ -175,13 +175,10 @@ where
...
 
                 let addr_string = format!("{remote_addr:?}");
-                record_event(
-                    ConnectionAccepted {
-                        timestamp_ns: clock_monotonic_ns(),
-                        remote_addr: addr_string.clone(),
-                    },
-                    &handle,
-                );
+                handle.record_event(ConnectionAccepted {
```
rcoh, line 178: presumably these could all call the naked `record_event` we define in `dial9-core`?

rcoh (approved):
seems fine as a first step. My comments mostly failed to understand that this pr only creates the packages and moves a couple of thing, rather than moving the event buffer into core.

dial9-core/src/lib.rs
jv1i: yeah I wanted to share the shuttle shims with dial9-tokio-telemetry to avoid duplicating those, I set them to hidden now.

dial9-tokio-telemetry/src/telemetry/recorder/handle.rs
jv1i: yup I was planning to do that as a follow-up

PR #543, opened 2026-06-15
Fix flaky capture_produces_frames: isolate SIGSEGV-handler-replacement test

rcoh:
## Problem

`unwinder::tests::linux::capture_produces_frames` flakes in CI:

```
thread 'unwinder::tests::linux::capture_produces_frames' panicked at perf-self-profile/src/unwinder.rs:112:9:
Unwinder::capture called but our SIGSEGV handler is no longer active; something
replaced it without chaining. See Unwinder::verify_handler.
```
(e.g. [run 27564559880](https://github.com/dial9-rs/dial9/actions/runs/27564559880/job/81484999193), `Build (ubuntu-24.04-arm, 1.94.1)`)

## Root cause

A process-wide signal-handler race between sibling tests.

`capture_debug_asserts_when_handler_replaced` deliberately replaces the **process-wide** SIGSEGV handler with `SIG_IGN` to verify that `Unwinder::capture`'s `debug_assert!(self.verify_handler())` fires. Its comment justified the safety of this with:

> *Nextest runs each test in its own process, so replacing the SIGSEGV handler here cannot affect sibling tests.*

But CI runs plain **`cargo test --all-targets --all-features`**, not nextest. The default libtest harness runs all unit tests **on threads within a single process**. During the brief window where SIGSEGV is `SIG_IGN`, any sibling test thread that calls `capture()` — most easily `capture_produces_frames` — observes `verify_handler() == false` and panics on the `debug_assert!`. The restore guard closes the window, which is why it's intermittent and timing-dependent (and showed up on the slower arm runner).

## Fix

Move the handler-replacement test into its own integration-test binary, `perf-self-profile/tests/handler_replacement.rs`. Each file under `tests/` compiles to a separate test binary that runs in its **own process**, so the global SIGSEGV mutation can no longer race with the unit tests — the isolation the original comment assumed is now actually provided.

Left a breadcrumb comment in the unit-test module pointing at the new file and explaining why it can't live there.

## Testing

- New `tests/handler_replacement.rs` passes (debug build, frame pointers on).
- Ran the `unwinder::` unit tests + the new integration binary in a **30× loop** under the CI `RUSTFLAGS` (`--cfg tokio_unstable -C force-frame-pointers=yes`): 0 failures (previously reproducible under thread interleaving).
- `cargo fmt --check` and `cargo clippy --test handler_replacement -- -D warnings` clean.


perf-self-profile/tests/handler_replacement.rs
```diff
@@ -0,0 +1,72 @@
+//! Verifies that [`Unwinder::capture`] panics via `debug_assert!` when its
```
jlizen, line 1: I would add a note in these module docs to not have multiple tests exercising `capture()` in here

perf-self-profile/tests/handler_replacement.rs
```diff
@@ -0,0 +1,72 @@
...
+    new_action.sa_sigaction = libc::SIG_IGN;
+    unsafe { libc::sigemptyset(&mut new_action.sa_mask) };
+    let mut old_action: libc::sigaction = unsafe { std::mem::zeroed() };
+    let rc = unsafe { libc::sigaction(libc::SIGSEGV, &new_action, &mut old_action) };
+    assert_eq!(rc, 0, "failed to replace SIGSEGV handler");
+
+    struct RestoreGuard(libc::sigaction);
+    impl Drop for RestoreGuard {
+        fn drop(&mut self) {
+            // Restore the original handler so concurrent tests under
```
jlizen, line 47: this is stale

jlizen (approved):

PR #546, opened 2026-06-17
Make trace= viewer param repeatable, ungzipping each component

rcoh:
## What

Make the `trace=` query parameter **repeatable** in the trace viewer (`viewer.html`) and standalone flamegraph (`flamegraph.html`), so you can load several trace components in one URL:

```
viewer.html?trace=<url1>&trace=<url2>&trace=<url3.gz>
```

## Why this needed work

The decoder already concatenates raw trace streams (a mid-stream `TRC\0` header resets the frame parser, so segments parse as one trace), and the `/api/trace` endpoint already gunzips its S3 objects server-side. The gap was on the fetch side: when components are fetched as **independent** `trace=` URLs, each one may be gzipped on its own, and nothing ungzipped them before concatenation.

## How

- **`trace_parser.js`**: new exported `fetchTraces(urls, opts)` — fetches each URL (in parallel, honoring an `AbortSignal`), runs each component through the existing `maybeGunzip`, then concatenates the raw bytes into one `ArrayBuffer`. Exported from both the Node and browser export blocks.
- **`viewer.html`**: reads `urlParams.getAll('trace')`; `loadTraceFromUrl` takes a single URL or a list and delegates to `fetchTraces`. Pop-out forwards all `trace=` params to the flamegraph.
- **`flamegraph.html`**: reads `params.getAll("trace")` and fetches via `fetchTraces`.
- **`index.html`**: the `?trace=` passthrough now forwards the whole query string, preserving every repeated `trace=` value.

## Tests

- **`test_fetch_traces.js`** (new): single raw, single gzipped→ungzipped, mixed gzip/raw concatenation that parses as one trace with 2× the events, order preservation, and fetch-error propagation. Mocks `fetch`, so no network.
- Registered the new test in **`scripts/e2e-trace-tests.sh`** (the `trace-integrity` CI job runs it) — CI does not auto-discover JS tests.
- Existing JS suite still green; `cargo build -p dial9-viewer` confirms `rust-embed` picks up the new files.

## Docs

Added `dial9-viewer/ui/README.md` and a note in `AGENTS.md` documenting that new `test_*.js` files must be registered in `e2e-trace-tests.sh`, since CI won't find them otherwise.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -3,6 +3,7 @@
 - You MUST use Red/Green TDD.
 - BEFORE COMMITING: `cargo nextest run --stress-duration 20s`. The package has no flaky tests. If you find a flaky test, you created it.
 - **JS/HTML-only changes** (no `.rs` files touched, no trace format changes): you do NOT need to run the full Rust test suite or the stress test. Run the relevant JS tests under `dial9-viewer/ui/test_*.js` with `node <test>` and a quick `cargo build -p dial9-viewer` to confirm `rust-embed` picks up any new files. Skip `cargo nextest` / stress run.
+- **Adding a new `dial9-viewer/ui/test_*.js` file:** CI does NOT auto-discover JS tests. You MUST register the new file in `scripts/e2e-trace-tests.sh` (the `trace-integrity` CI job runs that script), or it will never run in CI. See `dial9-viewer/ui/README.md`.
 
 ## API Design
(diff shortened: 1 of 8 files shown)


marcbowes (approved):

jv1i (approved):
nice addition of a readme for the viewer!

PR #549, opened 2026-06-17
refactor: move event bus to dial9-core

### LINKED ISSUE #510: Extract dial9-core: collector/source/flush_loop/writer/handle-core/guard

jv1i:
closes #510 and #511, part of #356.

### Summary

Moves the generic event bus and its consumer side out of `dial9-tokio-telemetry` into`dial9-core`.

In core:
- Bus: `Dial9Handle`/`SharedState`, the buffer (`Encodable`/`ThreadLocalEncoder`), clock + bus events, the collector, `Source`/`FlushContext`, the writer/`Fs`/sealed-segment lifecycle, the flush loop, and `CoreSession` (handle + flush-thread lifecycle).
- Pipeline: the `SegmentProcessor`/`SegmentData`/`ProcessError` contract.
- Worker (behind a `pipeline` feature, so the default build pulls no runtime): the segment-processing engine, `GzipCompressor`/`WriteBackProcessor`, on-trigger dumps (`DumpTrigger`/`DumpRun`/`DumpReceipt`), and S3 upload + dump manifests behind `pipeline-s3`.

New core deps: `dial9-trace-format`, `tracing`, `metrique`/`-timesource`/`-writer`, `bytes`,
`bon`, `ulid`, `tokio` (`sync` only by default). `pipeline` adds tokio rt + `flate2`;
`pipeline-s3` adds the aws SDK + `serde`/`serde_json`. Telemetry's public `worker-s3` forwards
to `pipeline-s3`, so the user-facing feature name is unchanged.

### Needed some decoupling (not a straight move)

`Fs` async drain: `Fs` had tokio coupling on its async drain. I lifted that onto the worker so the drain lives runtime-side.

Task-dump config: stays off core's `SharedState` (a tokio concept). The runtime hooks install it per-thread, including the block_on thread for current-thread runtimes. `TaskDumped` reads it lazily on first poll. 
I considered a generic per-session slot on core's `SharedState`, but it felt like early abstraction for now, kept it tokio-side so core doesn't have to deal with that.

`Source`/`FlushContext` are now genuinely `pub`: the fields are private and recording goes through `FlushContext::record_event`/`with_encoder`, so `CentralCollector` stays internal.

### API / behavior changes

- The pipeline contract, the worker, and the dump types are now `pub` same as `Source`/`FlushContext` above.
- The dump wiring seam is `pub` because the split puts the builder (telemetry) and the dump types (core) in different crates: `dump::channel`/`DumpRx`/`DumpTriggerConfig::debounce`/`SharedState::set_dump_trigger`.
- `SegmentProcessor` gains `finalize_dump` (default no-op); `ProcessError` gains `into_parts`.
- A processor panic now keeps the per-stage timings + the real segment duration. The per-segment metrics guard moved off `SegmentData` onto a worker-local, so the old "throw them away and emit a fresh empty guard" path is gone.

### Next PR (the `dial9` facade)

- The facade assembles the pipeline and owns `PipelineBuilder` + presets + worker spawn.
- memory-profiling moves to `perf-self-profile`
- `tracing-layer` + `memory-profiling` behind facade features, implicit `dial9::record_event`,`#[dial9::main]` re-export.
- `symbolize` -> `dial9-perf-self-profile` (breaks the core/perf cycle).
- Tighten remaining doc-hidden core internals to keep the public API small (and some semver adjustments).


dial9-core/Cargo.toml
```diff
@@ -10,8 +10,27 @@ description = "Telemetry event bus for dial9"
 unexpected_cfgs = { level = "warn", check-cfg = ["cfg(shuttle)"] }
 
 [dependencies]
+bon = "3"
+bytes = "1"
 crossbeam-queue = "0.3"
+dial9-trace-format = { workspace = true }
+event-listener = "5"
```
rcoh: If this is tokio with just the sync feature let's use that instead. People probably have it in tree anyway.

The only tokio freedom that matters is that a library or application can emit events.

A follow on that I think is worthwhile is adding a off-by-default `impl` feature.
With the feature disabled, the event recording APIs do nothing.

This lets a library depend on dial9-core and emit events basically for free even on compile time because we only wire the impl in when a consumer actually enabled the impl feature.

dial9-core/Cargo.toml
```diff
@@ -10,8 +10,27 @@ description = "Telemetry event bus for dial9"
...
+libc = { workspace = true }
+metrique = { version = "0.1.23", features = ["local-format"] }
```
rcoh: Local format shouldn't be a prod dependency?

dial9-core/Cargo.toml
```diff
@@ -10,8 +10,27 @@ description = "Telemetry event bus for dial9"
...
+metrique-timesource = "0.1"
+metrique-writer = "0.1"
 shuttle = { version = "0.9.1", optional = true }
```
rcoh, line 28: Generally you shouldn't need metrique-writer directly

dial9-core/src/buffer.rs
```diff
@@ -268,42 +244,38 @@ impl ThreadLocalBuffer {
...
         self.event_count = 0;
-        crate::telemetry::collector::Batch {
-            encoded_bytes,
-            event_count,
-        }
+        crate::collector::Batch::new(encoded_bytes, event_count)
     }
 
-    pub(crate) fn has_pending_events(&self) -> bool {
+    pub fn has_pending_events(&self) -> bool {
```
rcoh: Curious why this needs to be pub

dial9-core/src/buffer.rs
```diff
@@ -268,42 +244,38 @@ impl ThreadLocalBuffer {
...
         self.event_count > 0
     }
 }
 
+/// Encode a single event into a self-contained batch (header + event).
+/// Used by tests that need to write individual events through the batch API.
+#[doc(hidden)]
+pub fn encode_single(event: &dyn Encodable) -> Vec<u8> {
```
rcoh: This is really just for tests right? Should probably put this in a test-util module

dial9-core/src/buffer.rs
```diff
@@ -323,7 +295,8 @@ impl Drop for ThreadLocalBuffer {
 
 /// A handle to a thread-local buffer, held by `SharedState` so the flush
 /// thread can intrusively drain idle/silent buffers.
-pub(crate) struct TlBufferHandle {
+#[doc(hidden)]
```
rcoh: I'm a little worried about all these doc hidden.

We should have a clear set of APIs that are intentionally public

dial9-core/src/clock.rs
```diff
@@ -0,0 +1,59 @@
...
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .expect("system clock should not be before the unix epoch")
+        .as_nanos() as u64
+}
+
+/// Snapshot `(monotonic_ns, realtime_ns)` as close together as possible.
+/// Reads M₁ -> R -> M₂ and pairs `R` with the midpoint of M₁ and M₂ so
+/// the correlation error is half the `clock_gettime` interval.
+pub fn clock_pair() -> (u64, u64) {
```
rcoh: If this is a public function (does it really need to be?) it should return a struct

dial9-core/src/flush_loop.rs
```diff
@@ -77,7 +64,7 @@ pub(super) fn flush_once<M: WriterMode>(
     }
 
     while let Some(batch) = shared.collector.next() {
-        if batch.event_count > 0 {
+        if batch.event_count() > 0 {
             if let Err(e) = writer.write_encoded_batch(&batch) {
```
rcoh, line 68: Seems like the method we really need is is_empty

dial9-core/src/format.rs
```diff
@@ -0,0 +1,25 @@
...
+
+/// Clock-correlation anchor. `timestamp_ns` (monotonic) and `realtime_ns`
+/// (nanoseconds since Unix epoch) are captured at the same instant via
+/// [`clock_pair`], so offline consumers can recover wall clock from the
+/// monotonic event stream.
+///
+/// [`clock_pair`]: crate::clock::clock_pair
+#[derive(TraceEvent)]
+#[traceevent(wire_slot)]
+pub struct ClockSyncEvent {
```
rcoh: Does this actually need to be public? A lot of this stuff seems like it should be an internal detail of dial9-core

dial9-core/src/fs/mod.rs
```diff
@@ -190,45 +189,42 @@ impl TakenSegment {
 }
 
 /// Per-cycle snapshot returned by `Fs::take_files`.
-pub(crate) struct TakenFiles {
-    pub(crate) segments: Vec<TakenSegment>,
+pub struct TakenFiles {
```
rcoh: Needs non_exhaustive

dial9-core/src/handle.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// Per-thread [`Dial9Handle`], populated via [`set_tl_handle`] and cleared
+    /// via [`clear_tl_handle`] (from a runtime's thread-start/stop hooks).
+    /// Backs [`Dial9Handle::current`] and [`current_handle`].
+    static CURRENT_HANDLE: RefCell<Option<Dial9Handle>> = const { RefCell::new(None) };
+}
+
+/// Commands sent to the flush thread from the handle / guard.
+#[doc(hidden)]
+pub enum ControlCommand {
+    /// Flush, finalize (seal segment), then exit the thread.
```
rcoh, line 15: Again why pub?

dial9-core/src/metrics.rs
```diff
@@ -0,0 +1,88 @@
...
+
+/// Per-cycle counters returned by [`SharedState::drain_all_tl_buffers`], also
+/// used as a `#[metrics(subfield)]` so callers can flatten these fields into
+/// their top-level metrics without duplication.
+///
+/// [`SharedState::drain_all_tl_buffers`]: crate::shared_state::SharedState::drain_all_tl_buffers
+#[doc(hidden)]
+#[metrics(subfield)]
+#[derive(Debug, Default)]
+pub struct TlDrainStats {
```
rcoh: Why pub?

rcoh (commented):
Going in the right direction. We will need to carefully evaluate each struct and figure out if it actually needs to be pub or not.

jv1i:
@rcoh thanks for the review!

Yeah most of these I have on the todo (for this PR and follow ups), some things are pub because their consumer hasn't moved yet into core, and some things haven't been split/decoupled. In general I didn't intend this PR to be the final state of all the public APIs, but rather just moving the event bus to core. Most of the pub but hidden ones will go away entirely, and I need to do a pass over all the actual pub structs.

That said I'll tighten a few of these today, and I'll pack some more things into this PR rather than follow-ups. So that dial9-core doesn't look off and it's maybe better for review.

dial9-core/Cargo.toml
```diff
@@ -10,8 +10,74 @@ description = "Telemetry event bus for dial9"
...
+pipeline = [
+    "tokio/rt",
+    "tokio/time",
+    "tokio/macros",
+    "dep:tokio-util",
+    "dep:futures-util",
+    "dep:flate2",
+]
+## Built-in S3 upload stage. Implies the pipeline runner, pulls the aws SDK.
+pipeline-s3 = [
```
rcoh: I wonder if we should put the s3 stuff in the facade or in a by-default util crate

dial9-core/src/clock.rs
```diff
@@ -0,0 +1,59 @@
...
+        tv_sec: 0,
+        tv_nsec: 0,
+    };
+    unsafe {
+        libc::clock_gettime(libc::CLOCK_REALTIME, &mut ts);
+    }
+    ts.tv_sec as u64 * 1_000_000_000 + ts.tv_nsec as u64
+}
+
+#[cfg(not(target_os = "linux"))]
```
rcoh: We need to bring Franco's fixes in here

rcoh (commented):
Let's enable #![warn(unreachable_pub)]

That makes pub precise and it makes it much easier when looking at the code to see what it actually pub since pub in a private module is not allowed

dial9-core/Cargo.toml
jv1i: good idea! moved it into a `dial9-utils` for now, happy to rename into something like `dial9-backends/exports`, `dial9-contribs`, etc. if we want it more focused.

jv1i:
@rcoh I think this is should be in a good spot now, there are some minor things not included in the CHANGELOG.md yet because they'll change again when we make `dial9` into a facade and some of the public APIs are tweaked a bit. 

Next up on my list is:
- move mem profiling into perf-self-profile
- setup dial9 as a facade, bump to 0.5.0-alpha, do a migration guide, and here get rid of the final pub-but-hidden elements I have, most between core and tokio-telemetry, but most will move into dial9 (facade) and be changed there (actually most are for tests rather than production code).

I'll still do another thorough pass with the next work in case there's remaining items to tidy up I've missed. 
Let me know if there's anything we should definitely not merge yet.

rcoh (approved):

PR #552, opened 2026-06-18
feat(viewer): Make custom events clickable to inspect

Fluzko:
## Make custom events clickable to inspect
Custom event ticks in the events panel are now interactive. Click one to inspect it; hover shows a guide line.
Since custom events don't carry task/worker context in the trace. Previously you could see a tick and its fields on hover, but had no way to connect it back to the task that emitted it or locate it precisely on the lanes timeline.

- A click selects the resolved task (same effect as clicking its poll bar) and draws a persistent labeled vertical marker across all lanes. Clicking the same tick again toggles everything off.
- A cluster is clickable only if every event in it maps to the same single task.
image

### DIFF
--- a/dial9-viewer/ui/viewer.html
+++ b/dial9-viewer/ui/viewer.html
@@ -1122,6 +1122,17 @@
             const taskDetailPanel = document.getElementById("task-detail");
             const clearRangeBtn = document.getElementById("btn-clear-range");
             const tooltip = document.getElementById("tooltip");
+            // Position the (already-populated, display:block) tooltip near the
+            // cursor, clamped to the viewport on both axes. Flips above the cursor
+            // when it would spill off the bottom. Call after setting innerHTML.
+            function placeTooltip(e, dy = 16) {
+                const tw = tooltip.offsetWidth, th = tooltip.offsetHeight;
+                let tx = Math.min(e.clientX + 12, window.innerWidth - tw - 8);
+                let ty = e.clientY + dy;
+                if (ty + th > window.innerHeight - 8) ty = e.clientY - th - 8;
+                tooltip.style.left = Math.max(8, tx) + "px";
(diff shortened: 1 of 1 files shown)


rcoh:
this helps a little bit but I think there is still some more to handle before events really become usable:
1. if you click it doesn't pin the tooltip in anyway
2. the tooltip flows off the bottom of the screen almost always at least on my computer
3. what would the minimal work to be able to _graph_ data from one of the fields be?

Working backwards from the events we have right now, e.g. what would a good user flow look like to be able to graph the number of accepting connections?

rcoh:
can you make this a little nicer looking? 
Screenshot 2026-06-18 at 3 49 25 PM

the tracing span viewer _also_ can show metadata, it would be really nice to unify these.

I don't mind the right panel but the tracing span viewer has some nice features like a copy button and the k/v's are styled differently so its easier to read

Screenshot 2026-06-18 at 3 51 49 PM

Fluzko:
So I went further and did add a new `related` tab:

For a single selected event, a sectioned and navigable view of everything connected to it:
  - Enclosing spans: outermost first, indented by nesting depth.
  - Same span: events inside the innermost enclosing span.
  - Same task: events resolving to the same task.
  - Same type: other occurrences of the same event name, labeled by their resolved task.
  - Field correlation: a "<->" button on each field opens a section listing other events sharing that field value,
  shown only when more than one event matches.

  Each section is collapsible (state persists across selections) and supports per-direction "load more" around the
  selected event. Clicking any row navigates the timeline to that span or event, re-centering at the current
  zoom.

There're some minor other stuff like code structure enhances and minor visuals, like adding some `z-index` to avoid event dashed lane overlap non related stuff & event side panel initial width.

rcoh:
nice!

rcoh (approved):
nice! big improvement

PR #554, opened 2026-06-18
Add dial9-flamegraph + dial9-flamegraph-diff skills

rcoh:
## What

Two **headless** (non-viewer) skills for working with on-CPU profiles from trace segments, plus a stale-toolkit fix surfaced along the way.

The viewer already renders flamegraphs interactively in the browser (`ui/flamegraph.js`, embedded by `dial9-html-report`). These skills cover the cases the viewer can't: producing a standalone flamegraph **file** from the CLI, aggregating an **arbitrary union** of segments (e.g. N hosts × M spread-out minutes), and **diffing two profiles** — which the html-report skill explicitly punts on ("Use two separate flamegraphs instead").

### `dial9-flamegraph`
- `fold.js` — collapse segment(s) into **folded stacks** (`root;…;leaf <count>`), the lingua franca of flamegraph.pl / inferno / speedscope. Flags: `--leaf`, `--max-depth N`, `--json` (cache form).
- `merge_folded.js` — assemble cached per-segment `--json` folds into any subset/union without re-parsing (parsing is the slow step).
- Renders to SVG via `inferno-flamegraph`, or open the folded file in speedscope.

### `dial9-flamegraph-diff`
- `diff_folded.js` — compare two profiles and report normalized **"top movers"** (text or JSON), or emit inferno differential-folded for a red/blue diff SVG.
- Designed for **two time periods** (before/after a deploy, peak vs trough, incident vs baseline), **host vs host**, or **node vs fleet**.
- The ranked report is **dependency-free** (Node only); only the SVG path needs inferno.

## Why folded stacks
A folded intermediate is the pivot for both rendering *and* programmatic analysis, so one dial9-specific step (`fold.js`) unlocks the whole standard-tooling ecosystem and lets agents reason over deltas as plain data.

## Also in this PR: a stale `decode.js` fix
While testing against the demo trace I found `skills/dial9-toolkit/scripts/decode.js` was **stale** — missing `TAG_SCHEMA_ANNOTATIONS (0x06)` — so the bundled `analyze.js` (and anything using the unpacked toolkit parser) fails on current traces with `Unknown frame tag: 0x6`. `trace_parser.js` and `trace_analysis.js` were already identical between `ui/` and the toolkit; `decode.js` was the only drift. Synced it from the authoritative `ui/decode.js`. (Happy to split this into its own PR if preferred.)

## Tests
- New `ui/test_flamegraph_fold.js` exercises fold (full/leaf/max-depth, sample-count conservation), merge, and diff (round-trip + direction) against the demo trace.
- Wired into `scripts/e2e-trace-tests.sh` (per the "CI does NOT auto-discover tests" note).
- `cargo build -p dial9-viewer` green; `build.rs` embeds both skills; `dial9 agents skills` emits them and `fold.js` runs end-to-end from the unpack; `cargo fmt --all --check` clean.

## Open questions for reviewers
- inferno is an external dep for the SVG path. Acceptable as an optional renderer, or would you prefer the diff also be wired into the in-browser `buildFlamegraphTree` renderer (a viewer feature, separate effort)? A native S3-browser "diff two nodes / two time windows" view is the natural follow-up.
- Split the `decode.js` sync into its own PR?

### DIFF
--- a/dial9-viewer/skills/dial9-flamegraph-diff/SKILL.md
+++ b/dial9-viewer/skills/dial9-flamegraph-diff/SKILL.md
@@ -1,0 +1,92 @@
+---
+name: dial9-flamegraph-diff
+description: Compare two dial9 on-CPU profiles and report what got hotter or colder — as a ranked text/JSON "top movers" list and/or a red/blue differential flamegraph. Use to compare two time periods (before/after a deploy, peak vs trough, incident vs baseline), two hosts/nodes, or one node vs the fleet. Builds on dial9-flamegraph's folded stacks.
+---
+
+# dial9 differential flamegraph
+
+Answers "**what changed** between these two captures." Works on any two sets of
+segments folded the same way — the two sides can be:
+
+- **Two time periods** — before vs after a deploy, peak vs trough, during-incident
+  vs baseline-hour, today vs last-week. (The single most useful comparison.)
+- **Two hosts / nodes** — is one box behaving differently from another?
+- **One node vs the fleet** — is this host an outlier?
+
+Output is either a **ranked mover list** (text or JSON — what an agent reasons
+over) or a **red/blue differential flamegraph SVG** (what a human eyeballs).
+Red = hotter in B (candidate), blue = hotter in A (baseline).
+
+## Prerequisites
+
+Same as `dial9-flamegraph` (dial9 toolkit + segments). `diff_folded.js` itself is
+**dependency-free** — it only needs Node — so the ranked report works with no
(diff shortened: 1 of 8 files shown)


Fluzko:
New UI drifts
- Diffing profiles: The viewer now diffs two flamegraphs in-browser (before/after, host vs host)
- Aggregating segments: server-side /api/flamegraph already folds + aggregates across a host/time scope.
- Proposed follow-up already exists: PR asks about wiring diff into the browser as future work. Already done.

PR #564, opened 2026-06-23
feat: Add CPU usage visualization to the viewer

### LINKED ISSUE #555: display cpu usage in the viewer

noxware:
## Changes

- Adds a panel into the viewer that analyze `ProcessResourceUsageEvent` pairs to produce a graph of overall CPU usage.
- Segment metadata now carries the available parallelism reported by Rust.
- Panels are now collapsible.
  - They are all collapsed by initially, but their last state is persisted for future sessions. 
  
## Screenshot

With the new trace:

Screenshot 2026-06-22 at 10 52 40 PM

> `11` comes from my Docker settings.

From an old trace:

Screenshot 2026-06-22 at 11 21 25 PM

> No available parallelism reported, so the max reported is used as the cap.

## Note

The current changes still treat `ProcessResourceUsageEvent` as a custom event without dedicated parsing in the frontend, and these events are still visible in the custom events panel, as they still contain other useful data.

## Other

Closes #555


rcoh:
UX wise, I think the viewer is a little bit too honest about the time range that are being averaged over. I really just want a graph, I don't think we need three separate vertical crosshairs.

saying "40 intervals" isn't really that useful either.

In short, I think the view that is being shown is too deeply coupled to the implementation details of where the data is coming from.

One final thing, we need to be able to hide the panel

noxware:
- "40 intervals" is probably unnecessary yeah. I will remove that 👍
- I may have a bias since it was my own idea but... I think the extra vertical crosshairs are useful 😅
  - It may look as implementation specific but, the algorithm is pretty standard as software like `top` does the same of taking 2 samples as a reference.
  - We have those 2 samples in the UI, in the panel above, so I wanted to have an easy way to "trace the relationship visually".
  - If not with a crosshair, I would highlight the "block" some other way, as it's a natural way of visualizing the "precision" of the calculation (the more separated these lines are, the less precise the calculation is).
- We talked about hiding panels today with the other WyeWorks folks, what a coincidence 😆 Yeah. If we add another "memory graph panel" to this, we will have no space 🤣 I will try to make these panels toggleable, good call 👍

noxware:
@rcoh Sorry the delay for such small changes. I had a pretty busy day. 

I think you were right 👍  The extra crosshairs are unnecessary. The CPU graph is already composed of visible "blocks", so it's already behaving as I wanted in my last point. Sorry I didn't notice, I didn't have much time to think quietly earlier today 😖

I removed that and removed the "40 intervals" thing as well.

And, I made all panels collapsible 🚀 They are all collapsed by default, but if you open one, it will be remembered in `localStorage` for the future. [Check this™:](https://deploy-preview-564--dial9-tokio-telemetry.netlify.app/)

Screenshot 2026-06-24 at 12 38 51 AM

rcoh:
love it

rcoh (approved):

PR #566, opened 2026-06-23
unify S3 boot_id with the on-disk namespace boot_id

### LINKED ISSUE #548: Unify namespace `boot_id` with the S3Config `boot_id`
Follow-up to #532 / #534. After #534 got merged in there are two boot_ids generated independently:

- the on-disk namespace dir uses one (from setup_namespace)
- the S3 upload key uses a different one (from S3Config's default_boot_id)

So a segment that lands in qmxz-481/ locally gets uploaded under a different bvtn-481/ prefix in S3, and you can't correlate a local trace file with its upload. We should unify them: generate one boot_id and have the S3uploader inherit it, so the local dir and the S3 key share one identity.

This means removing boot_id from the public S3Config builder and having the runtime inject the namespace boot_id into the S3 config at build time. It falls back to the random default when S3Config is used outside the managed Dial9Config path.

skanjalkar:
After #534 there were two boot_ids generated independently: the on-disk namespace dir used one (from setup_namespace) and the S3 upload key used a different one (from S3Config's default). So a segment that landed in qmxz-481/ locally got uploaded under a different bvtn-481/ prefix in S3, and you couldn't correlate a local trace file with its upload.

Now the runtime injects the namespace boot_id into the S3 config and the embedded segment metadata at build time, so the local dir and the S3 key share one identity. It falls back to the random {4-alpha}-{pid} default when S3Config is used outside the managed Dial9Config path.

Added an integration test that decodes a sealed segment and asserts its boot_id metadata matches the namespace directory name. I confirmed it actually guards the behavior: with the injection removed, the test fails (segment boot_id rojy-... vs dir lvkt-...).

NOTE: This is a breaking change because it removes the `.boot_id()` from the public S3Config builder. I am not sure if this is the best way forward. I am open to ideas - should we keep it backwards compatible by keeping the builder method and have the namespace `boot_id` override it at build time instead or just roll with this?

Closes #548

### DIFF
--- a/dial9-tokio-telemetry/src/background_task/s3.rs
+++ b/dial9-tokio-telemetry/src/background_task/s3.rs
@@ -47,11 +47,10 @@ where
 
 /// Configuration for S3 uploads.
 ///
-/// Only `bucket` and `service_name` are required. The remaining fields have
-/// sensible defaults:
+/// Only `bucket` and `service_name` are required. The remaining builder fields
+/// have sensible defaults:
 ///
 /// - `instance_path`: system hostname
-/// - `boot_id`: `{4 alpha derived from timestamp}-{pid}` (unique per process start)
 /// - `prefix`: none (keys start at the time bucket)
 /// - `region`: auto-detected via `HeadBucket`
 /// - `key_fn`: built-in time-first layout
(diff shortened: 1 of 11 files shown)


rcoh:
We are preparing the 0.4 release series now so breaking changes are fine

dial9-tokio-telemetry/src/background_task/s3.rs
```diff
@@ -107,6 +110,12 @@ impl S3Config {
         &self.bucket
     }
 
+    /// Override the boot_id so S3 keys match the on-disk namespace directory.
+    /// Called by the runtime builder when the writer is namespaced.
+    pub(crate) fn set_boot_id(&mut self, boot_id: impl Into<String>) {
```
rcoh, line 117: I haven't looked at the code carefully but I wonder if there's a cleaner pattern where the dial9 handle exposes boot id and the s3 uploader can read it instead of needing to have it be set.

rcoh (commented):
This seems fine; give a quick look to whether we could read bootid instead of setting it. May be cleaner

rcoh:
I played around with the pull model. Doesn't work super well

rcoh (approved):

skanjalkar:
oh sorry I forgot to reply here - i looked a bit a while ago. its a pretty involved change with the pull model yeah

PR #567, opened 2026-06-23
viewer: stream trace objects from S3 (parallel download, no server-side buffering)

rcoh:
## What

Stream trace objects from S3 instead of buffering them server-side, and download multiple files in parallel from the browser.

Two commits:

1. **download trace files in parallel, stop decompressing server-side** — Adds `GET /api/object?bucket=&key=` which serves one object's bytes verbatim (still gzipped). The S3 browser now emits one `trace=/api/object?…` param per selected file; the viewer/flamegraph download them in parallel and gunzip client-side. Far less network transfer (files stay gzipped end-to-end). The old `/api/trace` (fetch-all, gunzip, merge, ship uncompressed) is kept but documented deprecated; nothing in the UI links to it.
2. **stream /api/object from S3 instead of buffering** — `get_object` previously ran `ByteStream::collect()` into a `Vec<u8>` before writing a byte to the client (~2s TTFB on real traces). Adds `StorageBackend::get_object_stream` returning an `ObjectStream` (boxed `Stream<Item = Result<Bytes, io::Error>>` + optional `content_length`). S3 overrides it: `GetObject` is still sent synchronously so `NoSuchKey`/auth/not-found classification surfaces (correct 404/401/403) before any body streams, then the body is streamed via `futures::stream::unfold` over `ByteStream::next()`. `ObjectStream` is marked `#[non_exhaustive]` per the API rules. Default impl wraps `get_object` so `LocalBackend`/test backends are unchanged.

## Why

Old path forced the backend to fetch every file, gunzip each, concatenate, and ship one large uncompressed blob — more bytes over the wire and serialized work, with no byte reaching the browser until the whole object downloaded from S3.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1692,6 +1692,7 @@ dependencies = [
  "aws-smithy-types",
  "axum",
  "axum-extra",
+ "bytes",
  "clap",
  "flate2",
  "futures",
--- a/dial9-viewer/Cargo.toml
+++ b/dial9-viewer/Cargo.toml
@@ -26,6 +26,7 @@ required-features = ["dev-server"]
 [dependencies]
 axum = "0.8"
 axum-extra = { version = "0.10", features = ["query"] }
+bytes = "1"
 clap = { version = "4", features = ["derive"] }
 flate2 = "1"
 futures = "0.3"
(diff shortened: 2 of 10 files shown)


jv1i (approved):
Nice!

PR #569, opened 2026-06-23
viewer: pixel-LOD canvas rendering + recycled span-id highlight fix

rcoh:
## What

Keep the viewer interactive on multi-million-span traces with pixel-level LOD canvas rendering, and fix a recycled-span-id highlight bug. (This was previously one commit bluntly titled "more trace improvements" — split into three reviewable commits here.)

Three commits:

1. **add pixel-LOD render helpers to trace_analysis (pure + tested)** — Pure, unit-tested helpers, not yet wired in: `makeBarCoalescer` (merge adjacent same-color pixel bars into one fillRect run), `pixelDownsampleSpans` (one representative per pixel column, longest wins), `pixelCoverage` (per-pixel coverage→opacity for sparse timelines), `computePollWakes` (extracted O(P·logP) wake→poll matching), and color memoization in `pollHeatmapColorQuantized`.
2. **pixel-LOD canvas rendering for large traces** — Wire those helpers into `renderLane` (active/park/poll drawing: downsample then coalesce; selected polls get `Infinity` weight so they can't be dropped) and `renderTaskDetail` (coverage-histogram opacity when polls outnumber pixels). Removes the ad-hoc `nsPerPx > 5000` LOD branch (downsampling is a pass-through when sparse, so zoomed-in rendering is unchanged). Plus UX/infra: `scheduleRenderAll` rAF-coalescing for hover/pan, synchronous render in `showViewer` (no blank-canvas flash), spinner on its own compositor layer, and a `?prof=1` render profiler.
3. **highlight all instances of a recycled span id** — Span ids are recycled, so one id maps to several intervals. The highlight used `spansById` (last-wins), silently missing earlier instances. Adds `spansByIdAll` (id → all spans) and iterates the tiny `selectedSpanIds` set, which also replaces the old O(workers × allSpans) per-render scan.

## Why

Fully zoomed out, a worker lane could draw one `fillRect` per span over ~1500px — ~99.9% overdraw and multi-second frames. Clicking a poll on a large trace took seconds.

## Note — one behavioral fix vs. the original branch

While splitting this out I found `computePollWakes`, as originally extracted, was **not** identical to the O(P²) loop it claimed to replace: at an exact shared poll boundary (`poll[k].end == poll[k+1].start == wake`) the old loop picked the **left** poll (no bump), the binary search picked the **right** poll. Impact is tiny (one ambiguous instant, visual-only) but the "identical" claim was wrong. **Fixed** to walk left to the lowest-index match — a true behavioral no-op — and added `testPollWakesSharedBoundary` plus on-boundary wakes in the randomized equivalence test (confirmed: the test fails on the old code, passes on the fix). This is the only change in this PR series that differs from the original branch.

## Verification

- `node dial9-viewer/ui/test_trace_analysis.js` → **all pass** (incl. new coalescer/downsample/coverage/poll-wakes tests and the boundary regression test).
- `cargo build -p dial9-viewer` clean.

JS/HTML-only; no `.rs` and no trace-format change. Independent of #PR1 and #PR2.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dial9-viewer/ui/test_trace_analysis.js
+++ b/dial9-viewer/ui/test_trace_analysis.js
@@ -20,6 +20,10 @@ const {
   getTraceTimeRange,
   hasCpuProfileSamples,
   analyzeAllocations,
+  makeBarCoalescer,
+  computePollWakes,
+  pixelDownsampleSpans,
+  pixelCoverage,
 } = require("./trace_analysis.js");
 
 async function main() {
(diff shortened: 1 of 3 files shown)


jv1i (approved):
LGTM.

testing NIT: would it be worth testing that `fillRect` count scales with pixels instead of polls? AFAICT now `renderLane` is not tested directly, so if something later doesn't go through the LOD wiring we might not catch a regression

PR #572, opened 2026-06-24
Add Link Sharing for trace backend

rcoh:


dial9-viewer/src/server/share.rs
```diff
@@ -0,0 +1,144 @@
...
+        trace_url: format!("/api/shared/{token}"),
+        viewer_url: format!("/viewer.html?shared={token}"),
+        token,
+    };
+    Ok((StatusCode::OK, axum::Json(resp)).into_response())
+}
+
+/// `GET /api/shared/{token}` — serve a previously shared trace.
+pub async fn get_shared(
+    State(state): State<AppState>,
```
Fluzko: At `create_shared` we're checking if `sharing_enabled` is turned on to continue, but here we're not. Not sure if you want to leave this endpoint open by design, but it seems strange to me, I would turn both endpoints off/on together

dial9-viewer/src/cli.rs
```diff
@@ -53,6 +53,13 @@ enum Commands {
         /// Off by default; there is no auth, so only enable on a trusted network.
         #[arg(long)]
         enable_upload: bool,
+
+        /// Enable shareable links: lets the viewer `POST /api/shared` to copy
+        /// a trace to the server's bucket under `shared/<token>.bin.gz` and
+        /// produce a link anyone can open. Requires `--bucket`. Off by default;
+        /// there is no auth on the write path — enable on a trusted network.
+        #[arg(long)]
+        enable_sharing: bool,
```
Fluzko, line 62: If `--bucket` is required, maybe is worth adding Clap's  `requires = "bucket"` to sync them

dial9-viewer/src/server/share.rs
```diff
@@ -0,0 +1,144 @@
...
+            (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                format!("compression failed: {e}"),
+            )
+        })?
+    };
+
+    state
+        .backend
+        .put_object(bucket, &key, stored, Some("application/gzip"))
```
Fluzko, line 106: This an unlimited `put_object`, if something goes wrong is a free-bucket-filler, should we add a quota? or support some config around it?

rcoh: yeah adding a hard limit is a good call. These are only intended to be served on trusted networks behind authentication but a hard limit is never a bad idea.

dial9-tokio-telemetry/README.md
```diff
@@ -642,6 +642,31 @@ dial9 serve --local-dir /tmp/my_traces
 dial9 serve --bucket my-trace-bucket
 ```
 
+### Shareable links
```
jasgin, line 645: Should there be a mechanism to delete as well?

PR #578, opened 2026-06-24
feat(viewer): export flamegraph as interactive SVG or folded stacks

prabrat:
Adds an "Export" control to the flamegraph toolbar that turns the rendered tree into two portable artifacts:

- Interactive SVG: a standalone, self-contained flame graph (hover, click-to- zoom, Ctrl-F regex search, URL state) ported from brendangregg/FlameGraph.
- Folded stacks (.txt): the universal `frame1;frame2 <count>` format consumed by inferno, flamegraph.pl, and speedscope.

The export module (flamegraph_export.js) is pure (no DOM/globals) so it runs under node for tests, and is embedded into the binary via rust-embed so the control is available without --dev.

Cleanups applied during review:
- Heap exports label weights as bytes/allocs (via a formatValue option), matching the on-screen graph, instead of hardcoding "samples".
- CPU-region and heap views thread descriptive export titles + filenames.
- filenameStem trims leading/trailing dots so exports can't become dotfiles.
- Folded export skips empty panel sections (no dangling "# header").
- flamegraphColor deduped into trace_analysis.js (single source shared by the canvas and the SVG export, so colors can't drift).
- Export menu: Esc closes it, closes on filter change / help open, aria attributes, initial disabled state; outside-click listener cleaned up in destroy(); export module resolved once and warns instead of failing silently.

Tests: test_flamegraph_export.js (registered in scripts/e2e-trace-tests.sh) covers folded serialization, SVG generation, panel merge, formatValue units, and filenameStem edge cases.

### DIFF
--- a/THIRD_PARTY_LICENSES
+++ b/THIRD_PARTY_LICENSES
@@ -1,0 +1,402 @@
+THIRD-PARTY LICENSES
+====================
+
+dial9 is licensed under Apache-2.0 (see LICENSE). It additionally incorporates
+the following third-party software, which is licensed separately as noted
+below.
+
+--------------------------------------------------------------------------------
+Brendan Gregg's FlameGraph (flamegraph.pl)
+--------------------------------------------------------------------------------
+
+The embedded interactive SVG script produced by `dial9-viewer/ui/flamegraph_export.js`
+(the `embeddedScript()` function and the surrounding SVG structure) is derived
+from flamegraph.pl by Brendan Gregg.
+
+    Copyright 2016 Netflix, Inc.
+    Copyright 2011 Joyent, Inc.  All rights reserved.
+    Copyright 2011 Brendan Gregg.  All rights reserved.
+
+    https://github.com/brendangregg/FlameGraph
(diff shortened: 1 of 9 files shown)


rcoh:
clicked around, this is awesome! thanks for the contribution!

rcoh:
the code looks great! there are some licensing issues. we need something like this in flamegraph_export.js:

```
  // The embedded interactive script (embeddedScript function) is derived from
  // flamegraph.pl by Brendan Gregg, licensed under the CDDL 1.0:
  //   Copyright 2016 Netflix, Inc.
  //   Copyright 2011 Joyent, Inc. All rights reserved.
  //   Copyright 2011 Brendan Gregg. All rights reserved.
  //   https://github.com/brendangregg/FlameGraph
  //   License: CDDL-1.0 (see THIRD_PARTY_LICENSES)
```

And add a THIRD_PARTY_LICENSES file (or add to an existing one) with the CDDL-1.0 text.

rcoh (approved):

PR #581, opened 2026-06-24
perf(telemetry): skip segment metadata work on idle flush cycles

### LINKED ISSUE #444: Update `segment_metadata` methods to delay allocations until actually required
right now, segment metadata can trigger an allocation on every flush loop even though they _usually_ don't provide any updates.

Refactor the way segment metadata and the flush loop interact to optimize this code path.

This will be a breaking change for `TraceWriter` and `Source`

Fluzko:
closes #444 

## Summary
Reworks the `Source::segment_metadata` trait method from a pull-everything-every-cycle model into a change-aware one.

```rust
fn segment_metadata(&mut self, out: &mut Vec<(String, String)>) -> bool // note the new 'out' param
```
A source appends its entries to a shared, reused buffer only when they changed since the last call, and returns whether it appended anything. The flush loop merges into the writer only when at least one source reported a change. Steady-state idle cycles now allocate nothing.

 Per-source change detection:
- MemoryProfileSource: metadata is fixed at construction, so it emits once (an emitted flag) and reports no change thereafter. The writer keeps it in its merged cache and re-emits it on every rotation.
- TokioRuntimesSource: metadata grows as runtimes/workers attach. It self-detects via a fingerprint (runtime count + total registered worker count). Both only ever grow and a worker's global id is fixed once assigned, so an unchanged fingerprint means unchanged metadata.

Notes:
- The flush loop is a hot, ~5ms-cadence path. Cloning every source's metadata each cycle was pure waste in steady state.
- Self-detection over an external change signal for the Tokio source: a caller attaching a runtime cannot forget to announce it, because the source derives change from the runtime/worker counts directly. This is what perf/segment-metadata-self-detect refers to.
- Sources are polled with `|=` so every source is polled and consumes its own change regardless of order.
- The out buffer is reused across cycles (drained, not moved out), so it keeps its capacity and change cycles stay allocation-light too.
- On shutdown, the writer writes final metadata before sealing, so single-segment traces still carry the runtime > worker mappings.


rcoh:
even on basically idle hosts, I can't actually see this on flamegraphs. Not positive this optimization is worth the added complexity to the source trait.

I'm not sure what the boolean `out` param buys us as well.

This minimal diff:
```
  -        let source_entries: Vec<(String, String)> = shared
  -            .sources
  -            .lock()
  -            .unwrap()
  -            .iter()
  -            .flat_map(|s| s.segment_metadata())
  -            .collect();
  -        if !source_entries.is_empty() {
  -            let mut merged = static_metadata.clone();
  -            merged.extend(source_entries);
  -            writer.update_segment_metadata(merged);
  -        }
  +        let source_entries: Vec<(String, String)> = shared
  +            .sources
  +            .lock()
  +            .unwrap()
  +            .iter()
  +            .flat_map(|s| s.segment_metadata())
  +            .collect();
  +        if !source_entries.is_empty() {
  +            writer.update_segment_metadata(source_entries);
  +        }
```

would save us the clone of static metadata since the writer merge will handle updates. I think we may at some point try to optimize / improve the source API but I'm not sure this is the right shape

Fluzko:
@rcoh answering by steps:

> can't see it on flamegraphs

That's expected, since I think it's the wrong instrument, the right metric is allocations/cycle.

I measured it with a counting allocator (pretty much as `pipeline_heap_bench` `TrackingAllocator`): steady state of 2 runtimes x 4 workers, memory profiling on, 1000 idle cycles, warm-up excluded. I tracked merge's return on every cycle (false for all variants) so output is identical and the only thing that differs is allocation:

| variant                          | allocs/cycle | 200Hz |
|----------------------------------|--------------|--------|
| main (static clone + merge)      | 25           | 5000/s |
| your diff (drop static clone)    | 24           | 4800/s |
| change-aware (fingerprint, idle) | 0            | 0/s    |

Your diff is a correct cleanup, but it removes the static-metadata clone, which was only 1 of 24. The rest is the per-cycle source rebuild, tokio's metadata_entry is sort + to_string/worker + join + format! per runtime, also merge's own collect + key clones. The merge no-op saves the wire write but not the allocation, because it runs after everything is already allocated. So your diff leaves ~96% of the per-cycle allocs in place. Only the early-return (before any rebuild) takes idle cycles to 0, which is what I understand from #444.

I'm fine if you want to review the actual implementation to simplify it, I could give it a shot. But this is _actually_ what's removing the allocs

rcoh (approved):
Yeah 5k allocs / s is real and worth fixing

PR #582, opened 2026-06-25
fix: Avoid truncation when rendering the s3 browser view

rcoh:
When browsing a bucket with >1k files per hour, traces would be truncated and at the end of each hour, it would look like data was missing.

This reworks that, deletes old unused APIs and generally cleans stuff up.

Sorry about the JS file churn; we should start using a JS file formatter.


dial9-viewer/src/server/browse.rs
```diff
@@ -0,0 +1,328 @@
...
+        .filter(|s| !s.is_empty());
+    let base = match (&state.default_prefix, key_prefix) {
+        (Some(pfx), Some(kp)) => format!("{}/{}", pfx.trim_end_matches('/'), kp),
+        (Some(pfx), None) => pfx.clone(),
+        (None, Some(kp)) => kp.to_string(),
+        (None, None) => String::new(),
+    };
+
+    let window = params.to - params.from;
+    let gran = if window < MINUTE_GRANULARITY_THRESHOLD_SECS {
```
jasgin, line 100: nit: mismatch w/ docs where it says "Window at or below which we drop to minute-granularity prefixes"

jasgin (approved):

PR #591, opened 2026-06-30
refactor: move memory profiling to dial9-perf-self-profile

jv1i:
part of #356.

### Summary

Moves memory profiling out of `dial9-tokio-telemetry` into `dial9-perf-self-profile`, structured like the existing CPU/sched profiling: a **standalone base** that needs no `dial9-core`, plus a **dial9 `Source` integration** behind the `dial9-source` feature. 

Currently the dial9 consumer is in dial9-tokio-telemetry but this will soon be moved into the new dial9-as-facade crate.

Initially I was moving the feature mostly "as-is" (see [e56c50c](https://github.com/dial9-rs/dial9/pull/591/commits/e56c50c7885d5f80672968efd2948916c50c2f50)), but it looked awkward making memory-profiling be dial9-core coupled while cpu profiling wasn't. So the second commits makes it work standalone without dial9-core, besides the dial9-source integration.

### Layering

The base (`memory-profiling` feature) is the `SamplingAllocator` hook, the buffer rings, and `MemoryProfiler::install() -> MemorySampler`. `MemorySampler::drain(|MemorySample|)` + `take_dropped()` is the standalone read API (the analogue of `PerfSampler::for_each_sample`). 
 
dial9 integration (`dial9-source`): `MemoryProfileSource` wraps a `MemorySampler` and impls `Source`, emitting events. A convenience `MemoryProfiler::install_into(&handle)` installs and registers the source. 

Telemetry's `memory-profiling` forwards to both perf features, the builder wires the source same as it does `CpuProfiler`.

### API / behavior changes

- New public base API: `MemorySampler`, `MemorySample`/`AllocSample`/ `FreeSample`, `DroppedSamples`.
- `MemoryProfiler::install` no longer takes a handle (returns a `MemorySampler`). The dial9 entry point is `install_into(&handle)`.
- Removed `MemoryProfilerGuard`. Memory now drains for the session/guard lifetime (matching CPU profiling) rather than being a permanent process-wide install.
- `telemetry::{AllocEvent, FreeEvent}` are gated behind `memory-profiling` (they were always compiled before). The decode-side `analysis_events::{AllocEvent, FreeEvent}` are unchanged.
- `push_test_alloc` is behind perf's new `test-util` feature (was`#[doc(hidden)] pub`)

### Deps & features

`scc` + `crossbeam-queue` move to `perf-self-profile` (behind `memory-profiling`), `bon` is added there for the config builder.

New perf features: `memory-profiling` (standalone), `unstable-events` (gates `non_exhaustive` on the wire events), `test-util`.


perf-self-profile/src/memory_profiling/clock.rs
```diff
@@ -0,0 +1,23 @@
...
+    }
+    ts.tv_sec as u64 * 1_000_000_000 + ts.tv_nsec as u64
+}
+
+#[cfg(not(target_os = "linux"))]
+pub(crate) fn clock_monotonic_ns() -> u64 {
+    use std::sync::OnceLock;
+    use std::time::Instant;
+    static EPOCH: OnceLock<Instant> = OnceLock::new();
+    (EPOCH.get_or_init(Instant::now).elapsed().as_nanos() as u64).saturating_add(1)
```
rcoh: why is this duplicated here?

perf-self-profile/Cargo.toml
```diff
@@ -13,13 +13,25 @@ all-features = true
 
 [features]
 __internal-bench = []
-## Enables the `Source` trait implementations for `CpuProfiler` and `SchedProfiler`.
+## Enables the dial9 `Source` integration for the profilers (`CpuProfiler`,
+## `SchedProfiler`, and the memory profiler when `memory-profiling` is on).
 dial9-source = ["dep:dial9-core"]
```
rcoh, line 18: in practice, this is going to be the only _useful_ way to use this crate right?

perf-self-profile/src/memory_profiling/tid.rs
```diff
@@ -0,0 +1,20 @@
...
+    unsafe { libc::syscall(libc::SYS_gettid) as u32 }
+}
+
+#[cfg(not(target_os = "linux"))]
+pub(crate) fn current_tid() -> u32 {
+    // No gettid off Linux; hand out a stable per-thread id from a counter.
+    use std::sync::atomic::{AtomicU32, Ordering};
+    static NEXT: AtomicU32 = AtomicU32::new(1);
+    thread_local! { static TID: u32 = NEXT.fetch_add(1, Ordering::Relaxed); }
+    TID.with(|t| *t)
```
rcoh: I really don't think we should duplicate these; we tweak them on a pretty regular basis.

dial9-tokio-telemetry/src/telemetry/mod.rs
```diff
@@ -24,14 +24,16 @@ pub(crate) use dial9_core::writer;
 pub use crate::traced::TracedFuture;
 pub use custom_events::{CustomEventsConfig, CustomEventsContext};
 pub use dial9_core::buffer::{Encodable, ThreadLocalEncoder};
+#[cfg(feature = "memory-profiling")]
+pub use dial9_perf_self_profile::{AllocEvent, FreeEvent};
 #[cfg(feature = "cpu-profiling")]
 pub use dial9_perf_self_profile::{
     CpuProfiler, CpuProfilingConfig, CpuSampleSource, SchedEventConfig, SchedProfiler,
 };
 pub use events::clock_monotonic_ns;
```
rcoh, line 33: seems like `clock_monotonic_ns` should move into core?

rcoh (commented):
overall it seems reasonable but I'm not sure attempting to avoid a dependence on `core` is useful

dial9-tokio-telemetry/src/lib.rs
```diff
@@ -12,9 +12,14 @@
 pub mod background_task;
 /// On-trigger pipeline runs: dump triggers, receipts, and ids.
 pub use dial9_core::dump;
+/// Sampled memory allocation profiling.
 #[cfg(feature = "memory-profiling")]
-pub mod memory_profiling;
-pub(crate) use dial9_core::{primitives, rate_limit, sampling};
+pub mod memory_profiling {
+    pub use dial9_perf_self_profile::memory_profiling::*;
```
rcoh, line 18: I don't think we want a glob re-export here; but I assume this isn't long lived?

perf-self-profile/src/memory_profiling/clock.rs
```diff
@@ -0,0 +1,23 @@
...
+
+/// Read monotonic time in nanoseconds.
+#[cfg(target_os = "linux")]
+pub(crate) fn clock_monotonic_ns() -> u64 {
+    let mut ts = libc::timespec {
+        tv_sec: 0,
+        tv_nsec: 0,
+    };
+    // SAFETY: `ts` is a valid timespec; CLOCK_MONOTONIC always succeeds.
+    unsafe {
```
rcoh: this doesn't include @noxware 's optimizations of macOS; I think that is a compelling reason not to duplicate this uneccessarily

dial9-tokio-telemetry/src/lib.rs
jv1i: yup these are just to avoid churn for now, it will all go away when moving this to facade.

perf-self-profile/Cargo.toml
jv1i: Thinking about it I guess a few weeks for some reason I misinterpreted #356 a bit, where someone could use cpu-profiling (and now mem-profiling) wihtout the rest of dial9. Rather than what we're aiming for which is _just without tokio_ instead. This was aiming to make it more decoupled so you can use it on its own, but I'll switch back to depending on core to avoid the other issues you mentioned.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `#[dial9_tokio_telemetry::main]` now performs an implicit graceful shutdown after the async body returns: it drops the runtime and drains the background worker so the final segment is symbolized, compressed, and uploaded. Configure the deadline with `Dial9Config::builder()...graceful_shutdown(Duration)` (default 1s) or skip it with `.disable_graceful_shutdown()`. The low-level `TracedRuntime` API is unchanged — call `TelemetryGuard::graceful_shutdown` yourself ([#479](https://github.com/dial9-rs/dial9/issues/479))
 - `SegmentProcessor::finalize_dump`: custom processors can flush per-dump state when an on-demand dump completes ([#549](https://github.com/dial9-rs/dial9/pull/549))
 - `ProcessError::into_parts`: destructure a processing error into its kind and the carried segment data ([#549](https://github.com/dial9-rs/dial9/pull/549))
+- Memory profiling can run standalone, without `dial9-core`: depend on `dial9-perf-self-profile` with the `memory-profiling` feature, set `SamplingAllocator` as the global allocator, call `MemoryProfiler::install()`, and drain `MemorySample`s via `MemorySampler::drain`. The dial9 trace integration is that crate's `dial9-source` feature.
```
rcoh, line 16: Is this still true?

rcoh (approved):

CHANGELOG.md
jv1i: no my bad that's leftover, I'll update in next PR

PR #597, opened 2026-06-30
Allow the viewer to assume roles when accessing buckets

rcoh:
A follow on to BYOC, this allows the trace viewer to use STS Assume role to assume roles it has access to when accessing trace data.


dial9-viewer/src/server/mod.rs
```diff
@@ -289,6 +308,58 @@ impl AppState {
...
+    ///   name the role/account) is logged server-side, never reflected.
+    pub(crate) async fn assume(
+        &self,
+        role_arn: &credentials::RoleArn,
+        region: Option<&str>,
+    ) -> Result<credentials::TempCredentials, (StatusCode, String)> {
+        let Some(assumer) = &self.role_assumer else {
+            return Err((
+                StatusCode::BAD_REQUEST,
+                "this server does not support assume-role credentials".to_string(),
```
jasgin, line 328: Suggestion for the UI portion if not already in there/planned, probably a good idea to ensure we hide the role arn input altogether when support_assume_role is false.

dial9-viewer/src/server/credentials.rs
```diff
@@ -185,6 +392,98 @@ pub fn parse_cred_headers(
...
 }
 
+/// Mints credentials by assuming a role. Abstracted as a trait so the STS call
+/// is an injectable seam: production uses [`StsRoleAssumer`]; tests inject a
+/// fake that returns canned credentials without a network call (mirroring how
+/// [`crate::storage::EphemeralS3Config`] injects the S3 connector).
+pub trait RoleAssumer: Send + Sync {
+    /// Assume `role_arn` and return temporary credentials. `region`, when set,
+    /// pins both the STS endpoint used to assume the role and the region carried
+    /// on the resulting credentials. Errors are opaque: the resolve path maps any
```
jasgin, line 402: Is there a set of errors that could stem from something that would need to be addressed from the client side? Maybe if they need to update their role's trust policy so the viewer can assume it? If so, should these be handle with more robust messages so the client knows what they need to do to get past it?

rcoh: Yeah I need this as a baseline to start testing stuff

jasgin (approved):

PR #602, opened 2026-07-01
feat(viewer): stream flamegraph/tokio-stats aggregation over SSE

rcoh:
## What

Replaces the **client polling** model for the aggregate flamegraph and tokio-stats APIs with **Server-Sent Events streaming**, and surfaces fold failures in the UI.

Previously each poll folded a small batch and the browser re-polled, re-aggregating the whole folded set every time (O(folded²) reads). Now **one request holds the connection open**: the server resolves the scope once (list → order → cap → folded set), emits the already-folded snapshot immediately, then folds the not-yet-folded files up to the sampling cap and pushes a fresh full snapshot **per file**, closing at the cap. Aggregation is incremental (merge one part-file at a time), and folding **stops when the client disconnects** (the fold `JoinSet` drops, cancelling in-flight folds).

The client reads the stream via `fetch`, not native `EventSource` — `EventSource` can't send the `x-dial9-aws-*` bring-your-own-credentials headers.

## Fold-error surfacing

`fold_stream` now yields a `FoldOutcome` (`Folded` | `Failed`) instead of silently dropping failures. `Coverage` carries `fold_errors` + a `fold_error_sample` message, and the UIs render a warning banner/status — so a systematic failure (e.g. an unwritable output bucket → `PutObject` AccessDenied) no longer looks like a silent empty tree. This came out of real testing, where an unwritable output bucket produced an empty flamegraph with no visible error.


rcoh:
Reviewed and pushed three commits:

- **Merge of `main`** — resolved the conflict with #607 (`agg_context_for` became `async`).
- **`fix(viewer): commit a fold by writing the samples part last`** — the `samples/` part listing is the durable "folded" record, but `write_parts` issued all three part PUTs concurrently. If the samples PUT landed while dict/polls failed **or were cancelled**, the file was permanently recorded as folded with dict/polls missing (folded files are never re-folded). This PR makes that window routine: dropping the fold `JoinSet` on client disconnect/Stop aborts folds mid-write. Now dict+polls are written first and samples last, so only complete folds commit. Includes a regression test (fails pre-fix). Also inlined the `fold_stream` → `async_stream_via_joinset` delegation.
- **`fix(viewer): clean up SSE stream teardown in the UIs`** — (1) flamegraph: a mid-stream error after events had rendered left the stats bar on an animated "refining…" spinner forever; now stamps "interrupted". (2) tokio-stats: `loadPeriod`'s promise never settled on abort (`openSse` calls neither `onClose` nor `onError` when aborted), so every superseded `loadAll`'s `Promise.all` hung forever. (3) corrected the `openSse` doc (its promise never rejects).

Verified: full `cargo nextest run` (813 tests) + `--stress-duration 20s` (2 clean iterations), JS tests (`test_sse.js` 21, `test_flamegraph_api.js` 43), `cargo fmt --check`, clippy (`__nonlinux_all_features`) — no new warnings.

dial9-viewer/src/server/flamegraph.rs
```diff
@@ -258,96 +282,213 @@ fn sample_filter(params: &FlamegraphParams) -> SampleFilter {
...
+                                &ctx.agg.output_prefix,
+                                &f.full_key,
+                            )
+                            .await
+                                && let Err(e) = state.accum.merge(samples, dict)
+                            {
+                                rate_limited_warn("flamegraph: merge failed", &e);
+                            }
+                            state.folded.insert(aggregate::part_leaf_of(&f.full_key));
+                        }
```
jv1i, line 425: minor: would it be worth counting merge/decode failures on already-folded files too? maybe semantically doesn't match with fold_errors, but I wonder if this fetch failing should it be counted somehow.

jv1i (approved):
LGTM. minor comment about error counting

PR #603, opened 2026-07-01
feat(viewer): Add differential flamegraph + A/B diff capture (unified flamegraph/tokio-stats)

rcoh:
## What

Adds a **two-sided differential flamegraph** and wires **A/B diff capture into the landing page**, unified so the same comparison can be launched in **either** the flamegraph or Tokio-stats view.

### 1. Flamegraph diff (reapplied onto main)
A two-panel differential flamegraph at `?diff=1&a=<b64>&b=<b64>`, where each side is an independent aggregate scope encoded by a base64url **scope-link codec** (carries the full refetch scope; never credentials).

- `flamegraph_diff.js` — tree merge (union by frame name, carrying both sides' total+self counts), relative-hotness color mapping, and the scope-link codec.
- `flamegraph_diff_view.js` — two-panel canvas view with independent per-side polling.
- `flamegraph.html`/`.css` — the `?diff=1` branch, the `Diff…` / `Copy link` dialogs, and styles.
- `storage.rs` — `S3Backend::list_objects_all` now classifies errors via `classify_s3_error` (auth → `Unauthorized`), consistent with the sibling listing methods, so callers can distinguish "needs credentials" from "empty window". Adds a unit test.

### 2. Landing-page A/B capture + unification (the main new work)
- `index.html` — a shared `selectionScope()` helper (single source of truth for turning a heatmap box into an aggregate scope, now used by the flamegraph button, the tokio-stats button, and diff capture); an **"Add to diff"** button + a **diff tray** (capture side A then B, show each side's meta, swap / clear / remove-a-side); and a **"Compare in: Flamegraph / Tokio Stats"** launcher that opens `<page>.html?diff=1&a&b`. Diff is aggregate-only and browse-mode only; scopes are held in memory (nothing persisted).
- `tokio_stats.html` — now consumes the **same** `?diff=1&a&b` codec with **two independent scopes**. Previously its diff only varied *time* under one shared scope; each period can now carry its own `scope` (bucket/prefix/service/host), resolved per-side via `effectiveScope()`. A `diffMode` flag gates all new behavior so the normal single-scope flow is unchanged (P1/P2 labels become A/B in diff mode).

## How it flows
Select a region → **Add to diff** (side A) → select another region → **Add to diff** (side B) → **Compare in: [🔥 Flamegraph] [⚡ Tokio Stats]**. Both targets receive the same two-independent-scope encoding; `api=1` is added per-side only for the flamegraph target.

## Testing
- `test_flamegraph_diff.js` — **92 tests pass** (added coverage for independent per-side time windows round-tripping through the codec).
- All CI-registered JS tests pass; `test_parse_key.js` passes.
- `cargo build -p dial9-viewer` clean; the ported `storage.rs` auth-classify test passes; `cargo fmt --check` + `cargo clippy --all-targets` clean.
- Per `AGENTS.md`, this is a JS/HTML change with one small Rust edit — no trace-format change, so no full nextest/stress run.


rcoh:
Brought this branch up to date with the SSE aggregation work (#602) and ported the diff view to the new protocol. Until #602 merges, this PR's diff includes #602's commits (merged in, not stacked — base stays `main` per the repo's no-stacked-PRs rule); once #602 lands, the diff collapses back to just the flamegraph-diff work.

Three commits:

1. **Merge of `viewer-sse-aggregation`** — conflict resolution: flamegraph.html keeps the diff-mode block + capture UI over the SSE api-mode rewrite; tokio_stats.html keeps the SSE `loadPeriod`/`loadAll` rewrite with diff-mode's per-period `effectiveScope()` feeding the stream URL; e2e script registers both new tests.
2. **SSE port of `flamegraph_diff_view.js`** — each side now opens one `Dial9Sse.openSse` stream instead of polling `refine=true` on a timer with `resp.json()` (which would hang-then-throw against the streaming endpoint) and plateau detection via the now-deleted `isCoverageFrozen`/`shouldAutoStopRefining` helpers. `repollSide`/`destroy` abort via `AbortController`, which also cancels server-side folding. Per-side fold failures now surface in the panel badge. `openSse` gained `err.status` on HTTP errors so the side-B credentials prompt can keep branching on 401/403.
3. `apiUrlFor` drops its `refine` parameter; tests updated to pin the new shape.

Verified: all 17 registered JS suites pass (including test_flamegraph_diff: 91, test_sse: 21), full `cargo nextest run` (814 tests) + `--stress-duration 20s` (2 clean iterations), `cargo fmt --check`, clippy — no new warnings.

dial9-viewer/ui/flamegraph_diff.js
```diff
@@ -0,0 +1,298 @@
...
+// the ratio finite when a frame is present on only one side — that frame lands
+// fully saturated rather than at +/-Infinity. Scale: +/-4 octaves of the B/A
+// fraction ratio clamps to full saturation.
+function diffColor(a, b, totalA, totalB) {
+  const tA = totalA || 1;
+  const tB = totalB || 1;
+  const epsA = 0.5 / tA;
+  const epsB = 0.5 / tB;
+  const fa = a / tA + epsA;
+  const fb = b / tB + epsB;
```
jv1i: I think there's a bug in diffColor's epsilon floor, it's sized relative to each side's own total (0.5/totalA, 0.5/totalB), not comparable across sides. So when the two captures have very different totals (quiet baseline vs incident, small capture vs fleet-wide), a frame unique to the bigger side can lose to the smaller side's own floor and get colored backwards (i.e. reads as heavier-in-A even though it doesn't exist in A at all).

I repro'd it with: A=1 sample, B=8846, almost the whole B panel came out blue, where if I'm understanding it right, it should come out red, right?

Image

dial9-viewer/ui/tokio_stats.html
```diff
@@ -367,44 +456,79 @@ <h1>⚡ Tokio Stats</h1>
...
+  // the stream closes (or is aborted); rejects only on a genuine error.
+  function loadPeriod(period, token, signal) {
+    const u = new URL("/api/tokio-stats", window.location.origin);
+    // effectiveScope resolves the right scope per side: in diff mode each
+    // period carries its own independent bucket/prefix/service/host scope; in
+    // normal mode it falls back to the shared top-level params. It also folds
+    // in the (editable) time window, so we just copy its keys onto the URL.
+    for (const [k, v] of effectiveScope(period)) u.searchParams.append(k, v);
+
+    const credHeaders = window.Dial9Creds ? Dial9Creds.headers() : {};
```
jv1i, line 523: I think tokio-stats diff mode is missing per-side credentials, `loadPeriod` sends the same session creds for both A and B, no headersFor(side)/credsB equivalent like flamegraph has. AFACIT if B is a different account this just fails silently, and unlike the flamegraph diff this one doesn't have the retry dialog

Image

jv1i (commented):
Overall feature looks and works great! commenting two issues I found

dial9-viewer/ui/tokio_stats.html
rcoh: Yeah... That feature was complex and hacky enough in flamegraph that I think I'll probably not support it that way in the long term but possibly with the SSE unification I can support both in the same way

jv1i (approved):

PR #613, opened 2026-07-06
small flamegraph / diff improvements

rcoh:


Fluzko (approved):

dial9-viewer/ui/flamegraph_diff_view.js
```diff
@@ -475,23 +497,41 @@
...
+    // shared client-driven `maxFiles` (small initial fold, raised by "Load
+    // more") rather than the server default.
     function startSide(side, scope, status) {
       const sse = getSse();
       const ctl = new AbortController();
+      status.refining = true;
+      updateMoreButton();
 
-      sse.openSse(apiUrlFor(scope), {
+      sse.openSse(apiUrlFor(scope, undefined, maxFiles), {
```
Fluzko: When this happens, it's usually better to do it in this format: 

```js
apiUrlFor({scope, url: undefined, maxFiles}) // << I would prefer 'null' instead of undefined

// or even better
apiUrlFor({scope, maxFiles})
```

PR #614, opened 2026-07-06
refactor: make dial9 the facade

### LINKED ISSUE #541: change argument of `on_disk_buffer` to be a directory, not a file
Right now `on_disk_buffer` refers to a file, but really, its a directory. this leads to lots of unneeded complexity

jv1i:
closes #541, #512, #513, part of #356

### Summary

Turns `dial9` from a CLI-only crate into the facade: the one dependency users add. 
Migration is `dial9-tokio-telemetry = "0.3"`-> `dial9 = "0.5"` plus import paths, mostly the same APIs.

The facade owns the user-facing surface: re-exports from core/telemetry/perf,
the feature gates, `dial9::record_event`, and `#[dial9::main]` (the macro now expands to `::dial9`; `#[dial9_tokio_telemetry::main]` is gone). 

`Dial9Config` is replaced: config is now `dial9::recorder(writer).with_tokio(..)` (programmatic) and `dial9::recorder_from_env()` (env). Examples and macro tests come along. The viewer CLI is unchanged behind the default `cli` feature.

#### New

The new piece is `dial9_core::recorder`, the recorder builder:
`recorder(writer).source(..).pipe(..).build()` → `Recorder`. This is the non-Tokio path of #356 user stories (1/7/9): plug any `Source`, no runtime required.

For that to work, a few things had to find their real home first:
- `CustomEventsSource` → core
- boot_id + namespace isolation → core
- `ProcessResourceUsageSource` / `SocketAcceptQueuesSource` → perf-self-profile (features `process-resource`, `linux-socket`)
- `SymbolizeProcessor` → perf-self-profile, so a non-Tokio recorder can symbolize via `.pipe(..)`

`Recorder` is now the RAII recording guard; on the tokio path, `TracedRuntime` wraps it.

### Needed some decoupling (not a straight move)

Assembly: instead of keeping two copies of the bus/flush/worker wiring, the tokio path delegates to `dial9_core::recorder`. The perf thread registration (ctimer) rides the builder's thread-start hook, so behavior is unchanged.

For config-to-runtime: `recorder(w).with_tokio(..)` builds a `TracedRuntime` without reimplementing assembly, and the facade wires the source->pipeline decision (cpu -> symbolize -> gzip -> writeback/S3).

Source registration is a small core trait, `RecorderSourceExt`, implemented by both the core recorder and the tokio `TracedRuntimeBuilder` (the tokio one forwards to its inner recorder). So `.source(..)` and the `.with_*()` perf sugar chain on either side of `.with_tokio(..)`, not only before it.

`RecorderSourceExt` also carries `on_recording_start(&Dial9Handle)`, a hook run once when recording starts. `with_memory_profiling` rides it, so it installs on the plain recording path as well as under tokio, with no tokio-side special case.

### API / behavior changes

Breaking: `Dial9Config` (+ builders, `Dial9ConfigBuilderError`/`ValidationError`/`TelemetryRuntimeError`) and the low-level `TracedRuntime::builder()` / `TracedRuntimeBuilder` are removed. Config is `dial9::recorder(writer).with_tokio(..)` or `dial9::recorder_from_env()`. Per-source knobs (`with_cpu_profiling`, etc.) move to `.with_*()` convenience methods on the recorder. 
`build_or_disabled()` -> `dial9::recorder_or_disabled(writer, ..)`. 
Multi-runtime stays: `TracedRuntime::trace_runtime()`. 
The `main` macro is owned by `dial9` now.

Breaking: `trace_path` collapsed into the writer. `with_trace_path`, `RecorderBuilder::worker_trace_path`, and the `NoTracePath`/`HasTracePath` type-state are gone. The writer's `base_path` is the single source of truth.

Breaking: the pipeline-mode type-state (S3-vs-custom markers) is gone. S3 vs custom is now chosen at build by `TracedRuntimeBuilder::with_s3_uploader` / `with_custom_pipeline`; `write_back` still needs a disk writer (enforced by `PipelineBuilder<Mode>`).

One behavior flip: process resource usage sampling was on by default on unix, it's now behind the `process-resource` feature.

The profiling features (`cpu-profiling`, `memory-profiling`, `process-resource`, `linux-socket`) no longer imply `tokio`, they're standalone `Source`s, auto-wired when `tokio` is also on. 
Facade defaults are `["cli"]`, so `cargo install dial9` works as before; `tokio` is off by default, and a minimal library build compiles tokio with only the `sync` feature (no runtime). 

## Migrating from `Dial9Config`

`Dial9Config` was Tokio-only. Its whole job was to become a `TracedRuntime`. It's replaced by a single builder: `dial9::recorder(writer)` records events with no runtime at all (`.build()` gives a `Recorder`), and `.with_tokio(..)` is the opt-in step that adds Tokio instrumentation and produces a `TracedRuntime`.

`#[dial9::main]` stays inherently Tokio (it builds a runtime to run your async `main`), so config for the macro is the Tokio branch, a `TracedRuntimeBuilder`. `Traced` marks the Tokio-instrumented types, as in `TracedRuntime` and `TracedFuture`.

As a user you now build a writer, wrap it in `dial9::recorder(writer)`, and chain what you want onto it.

It reads top to bottom, and sources go on either side of `.with_tokio`:

1. **Make a writer.** `DiskBuffer::builder()...build()` (or `MemoryBuffer`). This is where a bad path or budget fails now, so writer errors are just `io::Error`.
2. **`dial9::recorder(writer)`.** Start the recorder.
3. **Plug sources.** `.with_cpu_profiling(cfg)`, `.with_sched_events(..)`, `.source(..)`, and so on. Sources run on dial9's flush thread; the natural place is here, but they also chain after `.with_tokio`. Core-only knobs like `.metrics_sink(s)` go before it.
4. **`.with_tokio(|t| ...)`.** Attach a Tokio runtime; the closure tweaks the `tokio::runtime::Builder`. You now have a `TracedRuntimeBuilder`.
5. **Chain Tokio and pipeline options.** `.with_task_tracking(true)`, `.with_runtime_name("api")`, `.with_s3_uploader(cfg)`, `.with_custom_pipeline(..)`, `.graceful_shutdown(d)`, and so on.

Hand the result to `#[dial9::main]`. No `.build()` needed; the macro does that.

### The common case, before and after

```rust
// Before: Dial9Config
fn config() -> Dial9Config {
    Dial9Config::builder()
        .on_disk_buffer("/tmp/trace.bin")
        .max_total_size(64 << 20)
        .with_tokio(|t| { t.worker_threads(4); })
        .with_runtime(|r| r.with_task_tracking(true))
        .build_or_disabled()
}

// After: recorder builder
fn config() -> TracedRuntimeBuilder {
    let writer = DiskBuffer::builder()
        .base_path("/tmp/dial9-traces")
        .max_total_size(64 << 20)
        .build()
        .expect("build trace writer");
    dial9::recorder(writer)
        .with_tokio(|t| { t.worker_threads(4); })
        .with_task_tracking(true)
}
```

`.with_tokio` chains just like it did on `Dial9Config`; the writer is the one extra `let`, since `recorder(writer)` takes a built writer. To add CPU profiling or another source, chain it on either side of `.with_tokio`: `dial9::recorder(writer).with_cpu_profiling(cfg).with_tokio(f)`.

`build_or_disabled`'s "never panic on a bad trace dir" behavior is now `dial9::recorder_or_disabled(writer_result, |t| {...})`, which downgrades to a plain runtime if the writer can't be created. Reach for it in a service that must boot regardless; the `.expect(..)` above is fine for most code and examples.

For zero-code production config, there's nothing to build:

```rust
#[dial9::main(config = dial9::recorder_from_env)]
async fn main() { /* ... */ }
```

All `DIAL9_*` environment variables work exactly as before.

If you're not using Tokio, skip `.with_tokio`. `dial9::recorder(writer).source(..).build()` gives a `Recorder` that records events directly, no runtime and no `TracedRuntimeBuilder`.

### Quick reference

First, bring the extension traits into scope: `use dial9::{RecorderBuilderTokioExt, RecorderPerfExt, RecorderSourceExt};`.

| You had | You now write |
| --- | --- |
| `Dial9Config::builder().on_disk_buffer(p)...build()?` | `dial9::recorder(DiskBuffer::builder().base_path(p)...build()?)` |
| `...in_memory_buffer()...build()?` | `dial9::recorder(MemoryBuffer::builder()...build()?)` |
| `...build_or_disabled()` | `dial9::recorder_or_disabled(writer, tokio_fn)` |
| `...enabled(false)` | `dial9::TracedRuntimeBuilder::disabled()` |
| `Dial9Config::from_env()` | `dial9::recorder_from_env()` |
| `.with_cpu_profiling(cfg)` (and sched / process-resource / socket) | `.with_cpu_profiling(cfg)` on the recorder, either side of `.with_tokio` (order no longer matters) |
| `.with_worker_metrics_sink(s)` | `.metrics_sink(s)`, before `.with_tokio` (core-only, not on the traced builder) |
| `.with_runtime(\|r\| r.with_task_tracking(true))` and the other `with_runtime` knobs | chain them after `.with_tokio(f)`: `.with_task_tracking(true)`, `.with_s3_uploader(cfg)`, `.with_custom_pipeline(..)`, `.with_dump_trigger(..)` |
| `TracedRuntime::builder()...build_and_start(tb, writer)?` (low-level) | `dial9::recorder(writer).with_tokio(\|t\| /* flavor */).build()?`, then `.runtime()` / `.recorder()` / `.block_on(..)` |
| `TelemetryCore::builder()` + `guard.trace_runtime()` (multi-runtime) | `traced.trace_runtime(name).build(builder)` (on a built `TracedRuntime`) |
| `Dial9ConfigBuilderError` / `ValidationError` / `TelemetryRuntimeError` | removed; writer creation and `.build()` both surface plain `io::Error` |

To propagate a source-start failure instead of the log-and-skip default, use `.source(CpuProfiler::start(cfg)?)`.

# Follow ups

- **Tokio as a source** : collapse `with_tokio` and the multi-runtime path onto `recorder.attach_tokio(..)` one Recorder, runtimes attached as feeds. This kills the need for `RecorderSourceExt`

- **Move `tracing_layer` and `analysis`** out of tokio-telemetry (own crate or dial9); drop `tracing_layer`'s `tokio`
  implication.

- **Lazy writer**: `.enabled(false)` still builds the `DiskBuffer`, which touches disk. Something like a `recorder_on_disk(path)` that defers the writer to `.build()` would let disabled configs skip disk (see the `conditionally_enable` example). In general it would be handy for users (if worth the wiring complexity) to be able to do `.on_disk_buffer` and `.on_memory_buffer()` on top of the main recorder builder. Rather than having to construct a writer first.


dial9-macro/src/lib.rs
```diff
@@ -163,18 +155,20 @@ fn expand_main(args: MainArgs, input: ItemFn) -> Result<TokenStream2, syn::Error
 /// Using a named function:
 ///
 /// ```rust,ignore
```
rcoh: this example is `ignore`. Can it be `no_run`?

dial9-macro/src/lib.rs
```diff
@@ -163,18 +155,20 @@ fn expand_main(args: MainArgs, input: ItemFn) -> Result<TokenStream2, syn::Error
...
-/// use dial9_tokio_telemetry::{main, Dial9Config, telemetry::Dial9TokioHandle};
+/// use dial9::telemetry::Dial9TokioHandle;
+/// use dial9::{main, DiskWriter, RecorderBuilderTokioExt, TracedRecorder};
 ///
-/// fn my_config() -> Dial9Config {
-///     Dial9Config::builder()
-///         .on_disk_buffer("/tmp/trace.bin")
+/// fn my_config() -> TracedRecorder {
```
rcoh: not sure about the `TracedRecorder` name. What is "traced" about the recorder?

dial9-macro/src/lib.rs
```diff
@@ -187,30 +181,25 @@ fn expand_main(args: MainArgs, input: ItemFn) -> Result<TokenStream2, syn::Error
 /// Using an inline closure:
 ///
 /// ```rust,ignore
-/// #[dial9_tokio_telemetry::main(config = || {
-///     Dial9Config::builder()
-///         .on_disk_buffer("/tmp/trace.bin")
+/// #[dial9::main(config = || {
+///     let writer = dial9::DiskWriter::builder()
+///         .base_path("/tmp/trace.bin")
```
rcoh: the base_path should be a folder and not a file although we can do that in a separate PR

dial9-macro/src/lib.rs
```diff
@@ -187,30 +181,25 @@ fn expand_main(args: MainArgs, input: ItemFn) -> Result<TokenStream2, syn::Error
 /// Using an inline closure:
 ///
 /// ```rust,ignore
-/// #[dial9_tokio_telemetry::main(config = || {
-///     Dial9Config::builder()
-///         .on_disk_buffer("/tmp/trace.bin")
+/// #[dial9::main(config = || {
+///     let writer = dial9::DiskWriter::builder()
```
rcoh: I like `DiskBuffer` and `MemoryBuffer` for the names of these two things. We can call it a "buffer" instead of a writer

dial9/examples/blocking_sleep.rs
```diff
@@ -1,4 +1,5 @@
-use dial9_tokio_telemetry::telemetry::{DiskWriter, TracedRuntime};
+use dial9::prelude::*;
```
rcoh: I don't love having a `prelude`, there are only three types in there anyway. `*` imports are a big semver hazard as well.

dial9/src/lib.rs
```diff
@@ -1,3 +1,146 @@
...
+
+// One-call `.with_*` source sugar on the recorder builder. Available whenever a
+// perf source is compiled in.
+#[cfg(any(
+    feature = "cpu-profiling",
+    feature = "process-resource",
+    feature = "linux-socket",
+    feature = "memory-profiling"
+))]
+pub use dial9_perf_self_profile::RecorderPerfExt;
```
rcoh, line 132: wouldn't this belong in the prelude?

dial9/src/lib.rs
```diff
@@ -1,3 +1,146 @@
...
+
+// Offline symbolization processor for the segment pipeline. Needs both the CPU
+// profiler (to produce stack frames) and the pipeline (to run the processor).
+#[cfg(all(feature = "cpu-profiling", feature = "pipeline"))]
+pub use dial9_perf_self_profile::SymbolizeProcessor;
```
rcoh: why is this at the crate root?

dial9/src/lib.rs
```diff
@@ -1,3 +1,146 @@
...
+
+// Memory profiling.
+#[cfg(feature = "memory-profiling")]
```
rcoh, line 143: seems like this could be in a memory module?

dial9/src/lib.rs
```diff
@@ -1,3 +1,146 @@
...
+pub use dial9_perf_self_profile::{AllocEvent, FreeEvent, memory_profiling};
```
rcoh: Image

That's a lot of top level structs! Can you organize them?

dial9-tokio-telemetry/README.md
```diff
@@ -168,30 +167,30 @@ dial9 is fundamentally a central buffer that can collect data from different sou
...
-```rust
+```rust,ignore
 # #[cfg(feature = "worker-s3")]
 # mod inner {
-use dial9_tokio_telemetry::Dial9Config;
+use dial9::{DiskWriter, RecorderBuilderTokioExt, TracedRecorder};
 use dial9_tokio_telemetry::background_task::s3::S3Config;
 
-fn my_config() -> Dial9Config {
+fn my_config() -> TracedRecorder {
```
rcoh: TracedRecorder is actually a builder, not a recorder?

rcoh (commented):
We have `TelemetryCore` and `SessionCore` what is the difference? It seems like SessionCore and Telemetry core should be merged?

rcoh:
Meta: not sure about the Session name

dial9-tokio-telemetry/tests/memory_profiling_hook_realloc.rs
```diff
@@ -10,7 +10,7 @@ use dial9_tokio_telemetry::memory_profiling::{
     Dial9Allocator, MemoryProfiler, MemoryProfilingConfig,
 };
 use dial9_tokio_telemetry::telemetry::analysis_events::Dial9Event;
-use dial9_tokio_telemetry::telemetry::{InMemoryWriter, TracedRuntime};
+use dial9_tokio_telemetry::telemetry::{MemoryBuffer, RecorderBuilderTokioExt, recorder};
 use std::alloc::{GlobalAlloc, Layout, System};
 use std::sync::atomic::{AtomicUsize, Ordering};
```
rcoh: we should probably move these tests

dial9-tokio-telemetry/tests/background_symbolization.rs
```diff
@@ -31,24 +33,23 @@ fn burn_cpu_work() {
 #[test]
 fn background_symbolization_produces_symbol_table_entries() {
     let trace_dir = tempfile::tempdir().unwrap();
-    let trace_path = trace_dir.path().join("trace.bin");
 
     // Small segments to force rotation so the worker has segments to process.
     // Large total size so segments aren't evicted before the worker processes them.
-    let writer = DiskWriter::new(&trace_path, 4 * 1024, 10 * 1024 * 1024).unwrap();
+    let writer = DiskBuffer::new(trace_dir.path(), 4 * 1024, 10 * 1024 * 1024).unwrap();
```
rcoh: DiskBuffer needs a `#[bon]` builder

dial9/src/env_config.rs
```diff
@@ -0,0 +1,1299 @@
+//! Environment-driven configuration for the `#[dial9::main]` macro.
```
rcoh, line 1: lets call this `env_config`

dial9-core/src/shared_state.rs
```diff
@@ -1,6 +1,6 @@
-use crate::buffer;
```
rcoh, line 1: as far as I can tell from docs.rs this is a public module with no public types. (because shared state is doc hidden).

should we hide the whole module? does anything actually use `SharedState::new`?

dial9-tokio-telemetry/tests/background_symbolization.rs
```diff
@@ -61,11 +62,8 @@ fn background_symbolization_produces_symbol_table_entries() {
         tokio::time::sleep(std::time::Duration::from_secs(3)).await;
     });
 
-    drop(runtime);
     // Graceful shutdown: seals final segment, worker drains all remaining.
-    guard
-        .graceful_shutdown(std::time::Duration::from_secs(10))
-        .expect("graceful shutdown");
+    traced.graceful_shutdown();
```
rcoh: did we remove the timeout?

jv1i:
Tomorrow I'm finishing cleanup of the public types and some unnecessary re exports, and I'll write a list of the final follow up items (wrote the two main ones on the description above: Tokio as source and writer convenience). Will likely get through those tomorrow as well.

rcoh (approved):

dial9-tokio-telemetry/tests/background_symbolization.rs
jv1i: it actually already had bon, but I had this leftover `::new` that was unnecessary (was only using it in tests), removed now

dial9-core/src/shared_state.rs
jv1i: yeah currently the only prod use is from core itself, outside of that it's only used for tests on sibling crates, so I made it a test-util now and hid the whole module.

my goal is to get the whole thing to pub(crate). tokio's poll/park hooks still use the type directly, for recording and to run the sources' `on_worker_thread_start`. we already have a source hook for this, so it's mostly switching to recording through `Dial9Handle::record_event` and exposing a public dispatch on the handle for the worker-thread hooks. then I think can drop all `SharedState` to just pub(crate). I can do this in the tokio as a source follow up

dial9-tokio-telemetry/tests/background_symbolization.rs
jv1i: had by mistake, fixed

PR #619, opened 2026-07-07
Sample SchedStat::read_current in worker park/unpark

### LINKED ISSUE #557: sample `SchedStat::read_current` in worker park/unpark
Reading from sched stat to check for kernel scheduling latency can add a non-trivial amount of CPU usage. But there is really no reason to do it on every park, you can still see what its a problem even if we don't catch every scheduler pause.

We should have it be configurable, probably 1 in 10 by default? Note that we must record a park->unpark pair for the data to be accurate.

rcoh:
## Problem

`SchedStat::read_current` reads `/proc/self/task/<tid>/schedstat` on **every** worker park and unpark to measure kernel scheduling latency. That read costs measurable CPU, and per #557 there's no need to catch every scheduler pause: the problem is still visible with periodic sampling.

## Change

Read schedstat on only **1-in-N** parks. `N` defaults to `10` and is overridable via the `DIAL9_SCHED_WAIT_SAMPLE_RATE` environment variable (clamped to `>= 1`; `1` restores read-on-every-park). A per-thread `PARK_COUNTER` drives the decision.

The correctness constraint from the issue (*"we must record a park->unpark pair for the data to be accurate"*) is preserved: the park handler records whether it sampled in a thread-local `SCHED_SAMPLED_THIS_PARK` flag, and the matching unpark on the same worker thread reads schedstat (and computes the wait-time delta) **iff** the park did.

### `sched_wait_ns` is now `Option<u64>`

Because only a subset of unparks now carry a measurement, reporting `0` for the rest would be ambiguous with a genuine zero-wait unpark and would dilute downstream averages. `WorkerUnparkEvent::sched_wait_ns` becomes `Option<u64>`:

- `None` — this park->unpark pair was not sampled (or the read failed).
- `Some(delta)` — sampled; `Some(0)` is a real zero-wait measurement.

The wire format already supports optional fields, and the decode-side structs are `#[non_exhaustive]`, so old traces (which encode a plain `u64`) still deserialize.

Consumers updated accordingly:
- **analysis**: only `Some(..)` folds into `total_sched_wait_ns` / `max_sched_wait_ns`; added `sched_wait_sample_count` as the correct denominator for the average (instead of `unpark_count`); `detect_sched_delays` skips `None`.
- **viewer JS parser**: decodes `null` (not `0`) for unsampled unparks; the sched-delay overlay already treats a null/absent wait as "no delay".

## Testing

- `cargo test -p dial9-tokio-telemetry --features analysis` — all 153 lib tests pass, plus the `js_parser` round-trip test (Rust encode -> JS decode) and the Node viewer test suites (`test_trace_integrity`, `test_block_in_place`, `test_trace_analysis`).
- Added tests: `park_counter_samples_one_in_n` (1-in-N logic incl. rate 1 and the `rate == 0` guard), `unsampled_unparks_are_excluded_from_sched_wait_stats` (None excluded from the average, `Some(0)` counted), and `detect_sched_delays_skips_unsampled_unparks`.
- `cargo fmt --check` clean; `cargo clippy` clean (only a pre-existing unrelated dead-code warning).

## Notes

- The env-var default (10) matches the "1 in 10" suggested in #557. If you'd prefer this wired through a builder/`Config` field instead of an environment variable, happy to adjust.

Fixes #557

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
+++ b/dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
@@ -90,7 +90,7 @@ fn sched_wait_sample_rate() -> u64 {
 /// process-global rate or a live schedstat read.
 fn advance_park_counter(counter: u64, rate: u64) -> (u64, bool) {
     let next = counter.wrapping_add(1);
-    (next, next % rate.max(1) == 0)
+    (next, next.is_multiple_of(rate.max(1)))
 }
 
 /// Returns a strictly monotonic timestamp for this thread.
--- a/dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
+++ b/dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
@@ -427,17 +427,17 @@ pub(super) fn make_worker_unpark(ctx: &RuntimeContext, shared: &SharedState) -> 
     let cpu_time_nanos = crate::telemetry::events::thread_cpu_time_nanos();
     // Only read schedstat on unpark if the matching park sampled it, so the
     // delta below always pairs with a park-time reading. Reset the flag either
-    // way so a subsequent unsampled park can't reuse a stale pair.
+    // way so a subsequent unsampled park can't reuse a stale pair. Report `None`
+    // when unsampled so consumers can distinguish "not measured" from a genuine
+    // zero-wait unpark rather than diluting their averages with false zeros.
     let sampled = SCHED_SAMPLED_THIS_PARK.with(|c| c.replace(false));
     let sched_wait_delta_nanos = if sampled {
-        if let Ok(ss) = SchedStat::read_current() {
+        SchedStat::read_current().ok().map(|ss| {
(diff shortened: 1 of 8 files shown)


dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -54,6 +54,43 @@ thread_local! {
...
+
+/// How often to read `SchedStat::read_current` in the worker park/unpark path.
+///
+/// Reading `/proc/self/task/<tid>/schedstat` on every park is measurable CPU
+/// overhead, and there is no need to catch every scheduler pause: periodic
+/// sampling still surfaces scheduling latency. We therefore only read schedstat
+/// on 1-in-N parks. `N` defaults to 10 and is overridable via the
+/// `DIAL9_SCHED_WAIT_SAMPLE_RATE` environment variable (values are clamped to at
+/// least 1; `1` restores read-on-every-park).
+fn sched_wait_sample_rate() -> u64 {
```
jasgin, line 75: Is it worth adding this to the trace metadata to know what the representation is?

jasgin (approved):

PR #635, opened 2026-07-09
viewer: classify S3 NoSuchBucket/NoSuchKey as 404, not 500

rcoh:
Browsing a nonexistent bucket returned HTTP 500 with a server `fault`. The S3 `NoSuchBucket`/`NoSuchKey` error codes fell through the `classify_s3_error` catch-all to `StorageError::Other`, which logged an "unclassified S3 error" and mapped to 500. A user typo in the bucket name is a client mistake, not a server fault — but it was the only source of faults in a 24h prod sample, so it polluted the viewer's fault metric and gave the user an opaque "500 Internal Server Error" instead of an actionable message.

Map those codes to `StorageError::NotFound` (→ 404), mirroring the existing `PermanentRedirect` → `WrongRegion` treatment and the call-site `NoSuchKey` handling in `get_object`. Add a regression test.

### DIFF
--- a/dial9-viewer/src/storage.rs
+++ b/dial9-viewer/src/storage.rs
@@ -239,6 +239,17 @@ where
         // form) or the generic `Redirect`. Surface a clear message rather than
         // the opaque "unclassified S3 error" this used to fall through to.
         Some("PermanentRedirect" | "Redirect") => StorageError::WrongRegion,
+        // Missing bucket/key: the user pointed at something that does not exist
+        // (typo'd bucket name, deleted object). This is a client mistake, so map
+        // it to 404 rather than letting it fall through to the `Other` arm —
+        // which logged an "unclassified S3 error" and returned a 500 `fault`,
+        // polluting the fault metric with user input. The list/prefix paths hit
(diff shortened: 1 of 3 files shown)


dial9-viewer/src/storage.rs
```diff
@@ -239,6 +239,17 @@ where
...
+        // (typo'd bucket name, deleted object). This is a client mistake, so map
+        // it to 404 rather than letting it fall through to the `Other` arm —
+        // which logged an "unclassified S3 error" and returned a 500 `fault`,
+        // polluting the fault metric with user input. The list/prefix paths hit
+        // this via the bucket-level `NoSuchBucket`; `GetObject`'s `NoSuchKey` is
+        // additionally handled at the call site (with the bucket/key in the
+        // message), so plain code matching here is the fallback.
+        Some("NoSuchBucket" | "NoSuchKey" | "NotFound") => {
+            StorageError::NotFound("the specified bucket or object does not exist".to_string())
+        }
```
jasgin, line 259: what about InvalidBucketName?

jasgin (approved):

PR #639, opened 2026-07-09
feat(viewer): Tokio Stats "Longest polls" rollup + focus deep-link

prabrat:
Add a top-N longest-polls rollup to the Tokio Stats page: the single futures that held a worker thread longest in one poll (these starve other tasks on that worker). The server ranks polls by duration and ships a bounded top-100 (task_id is high-cardinality, so this is a server-side reduction, not a per-poll axis); the UI renders a threshold-filtered "Longest polls" card with duration heat coloring.

Each row deep-links into the trace viewer focused on the exact poll via new non-destructive focus_* URL params. These are distinct from start/end (a hard parse filter that drops surrounding events and loads an empty page): focus_* pans/zooms the already-parsed trace to the poll window, scrolls to its worker lane, and highlights the task.

The aggregate stores poll timestamps in wall-clock ns while the viewer timeline is raw monotonic ns, so focusOnWindow converts wall->monotonic by subtracting the trace's clock offset (the same offset the decoder applied), keeping the deep link on-timeline.

No trace-format change: worker_id/task_id are already in the polls Parquet and already decoded; the endpoint just wasn't reading them. New response fields are additive.

Tests: extend the demo-trace tokio-stats test (top-N populated, bounded, ranked, carries deep-link coords); add JS coverage for the latency-heat helper and focus_* deep-link params; add test_focus_timebase.js pinning the wall->monotonic round-trip (registered in e2e-trace-tests.sh).

### DIFF
--- a/dial9-viewer/ui/viewer.html
+++ b/dial9-viewer/ui/viewer.html
@@ -1298,6 +1298,11 @@
             let currentPoiIndex = -1;
             // Task selection state
             let selectedTaskId = null;
+            // Pending one-shot focus request from `focus_*` URL params (a poll
+            // deep-linked from Tokio-stats). Applied after the first successful
+            // load, then cleared so a later Set/Clear-Range re-parse doesn't
+            // yank the view back to it. Null when no focus was requested.
+            let pendingFocus = null;
             // A clicked custom event: {events, timestamp, taskId, name} —
             // draws a labeled vertical marker across all lanes and, if the
             // event ran inside a poll, selects that task.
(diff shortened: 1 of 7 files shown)


dial9-viewer/ui/viewer.html
```diff
@@ -1993,6 +2005,21 @@ <h3>⌨ Keyboard</h3>
                 window.history.replaceState(null, '', newUrl);
             }
 
+            // One-shot focus window from `focus_*` params (Tokio-stats "Longest
```
rcoh, line 2008: Love it! We should teach the skills about this; that will allow them to construct deep link URLs for users to show them what's going on

prabrat: Absolutely, that would be awesome! Can add it to me todo incase I get some time to get to it, since I saw it already had some analyzer skills

rcoh (approved):
LGTM!

PR #660, opened 2026-07-10
feat: explicit CPU profiling backend selection via CpuProfilingConfig constructors

jasgin:
Introduce CpuProfilingConfig::with_perf_backend() and CpuProfilingConfig::with_ctimer_backend() constructors for explicit backend selection:

- with_perf_backend() — perf only, no ctimer fallback. Fails if
  perf_event_open is blocked.
- with_ctimer_backend() — ctimer only, avoids perf's per-CPU inherited
  context overhead on thread-heavy workloads.

https://github.com/dial9-rs/dial9/issues/579

### DIFF
--- a/dial9-tokio-telemetry/README.md
+++ b/dial9-tokio-telemetry/README.md
@@ -287,6 +287,21 @@ Dial9Config::builder()
     // ...
 ```
 
+By default, dial9 tries the perf backend and falls back to ctimer if
+`perf_event_open` is blocked. You can select the backend explicitly:
+
+```rust,ignore
+// Use ctimer directly — zero thread lifecycle overhead, ideal for workloads
+// with high thread churn (e.g. saturated block_in_place usage).
+CpuProfilingConfig::with_ctimer_backend()
+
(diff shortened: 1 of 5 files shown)


dial9-tokio-telemetry/README.md
```diff
@@ -287,6 +287,23 @@ Dial9Config::builder()
...
+
+```rust,ignore
+// Use ctimer directly — zero thread lifecycle overhead, ideal for workloads
+// with high thread churn (e.g. saturated block_in_place usage).
+CpuProfilingConfig::default().backend(CpuBackend::Ctimer)
+
+// Require perf — fail instead of silently degrading. Needed for kernel
+// stacks or hardware event sources.
+CpuProfilingConfig::default().backend(CpuBackend::Perf {
+    event_source: EventSource::SwCpuClock,
```
rcoh: this will inherently be a semver risk, I'd probably use a builder. You could do `.with_perf_backend()`, `with_ctimer_backend()` etc. potentially?

rcoh (approved):

PR #661, opened 2026-07-10
fix(viewer): resolve span task + reconstruct active/idle from task polls

rcoh:
Clicking a tracing span often failed to jump to its task when spans crossed workers

Root causes:
- A span segment recorded the ENTER timestamp paired with the EXIT event's worker. A long-lived span's task migrates workers between enter and exit, so the span->task lookup (poll overlapping seg.start on seg.workerId) searched the wrong worker and found nothing.
- Long-poll spans are emitted as a single SpanEnter/SpanExit pair spanning the whole request, so the one on-wire segment reported the entire await as on-CPU: activeNs == wall-clock, no idle gap.

Fix:
- Pair each segment with the ENTER worker (matches its enter-time start).
- buildSpanData now optionally takes workerSpans. It resolves each span's owning task from the poll covering its enter (binary search over the worker's sorted, non-overlapping polls), stores taskId on the span, and reconstructs active segments by intersecting each raw segment with that task's poll timeline. Idle gaps between polls now render, and activeNs reflects on-CPU time. Without workerSpans the old raw-segment behavior is unchanged.
- Span-click handler uses the stored taskId instead of re-deriving it.

All lookups are binary-search based (traces routinely exceed 10M events; a worker holds hundreds of thousands of polls, so resolveSpanTask must not scan linearly per span).

Validated on a real 172k-event beta trace: span->task resolution 133/364 ->
364/364; spans showing an idle band 0/364 -> 363/364; summed active time corrected from a bogus 1.8Ms (overlapping coarse segments) to 66ms on-CPU.

Tests: reconstructSpanSegments (carve/clamp/fallback/identity/multi-segment), resolveSpanTask (binary search, gap/unknown-worker/taskId-0 -> null), buildSpanData integration (reconstruction, recycled-task-id not borrowed, no-workerSpans backwards compat).

### DIFF
--- a/dial9-viewer/ui/test_trace_analysis.js
+++ b/dial9-viewer/ui/test_trace_analysis.js
@@ -14,6 +14,9 @@ const {
   flattenFlamegraph,
   buildFgData,
   buildSpanData,
+  indexPollsByTask,
+  resolveSpanTask,
+  reconstructSpanSegments,
   collectDescendants,
   selectSpanRenderSet,
   computeSpanLayout,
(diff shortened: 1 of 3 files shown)


jv1i (approved):
LGTM

PR #662, opened 2026-07-10
viewer: Enable embedding the viewer in larger application

rcoh:
## What
- Adds new public API to construct `AppState`
- Adds an example demonstrating how to embed the viewer in a larger application

### DIFF
--- a/dial9-viewer/src/server/mod.rs
+++ b/dial9-viewer/src/server/mod.rs
@@ -126,6 +126,38 @@ impl AppState {
         }
     }
 
+    /// Build an `AppState` backed by an S3 bucket, with automatic region
+    /// detection, bring-your-own-credentials support, and the assume-role
+    /// credential path enabled.
+    ///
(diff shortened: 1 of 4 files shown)


dial9-viewer/src/server/mod.rs
```diff
@@ -126,6 +126,38 @@ impl AppState {
...
+    ///
+    /// This is the high-level entry point for embedders who want to serve
+    /// traces from S3 without replicating the CLI's setup logic:
+    ///
+    /// ```ignore
+    /// let state = AppState::from_bucket("my-traces", None).await;
+    /// let app = dial9_viewer::server::router(state);
+    /// // … customize app, then bind …
+    /// ```
+    pub async fn from_bucket(bucket: impl Into<String>, prefix: Option<String>) -> Self {
```
jasgin, line 141: Should this return a result to handle failure modes like if the bucket doesn't exist or network errors, or is it not relevant?

jasgin (approved):

PR #663, opened 2026-07-10
Poll-time aware flamegraphs

rcoh:
This adds the ability to load a flamegraph selecting _specifically_ based on the length of the `poll` the sample was a part of. The upshot is that allows you to see a flamegraph but _exclusively_ for poll > 10ms, (etc.)

This is really the POC / stepping stone for a larger effort to add span-awarewess to the profiler / profiler aggregation.


dial9-viewer/ui/flamegraph.html
```diff
@@ -731,6 +834,282 @@
...
+                            meta.poll_duration_histogram);
+                        if (bars.length === 0) { minimap.style.display = "none"; return; }
+                        if (minimapDragging) return; // don't clobber an in-progress brush
+                        minimap.style.display = "block";
+
+                        // Seed the split to the sample-weighted median the FIRST
+                        // time we see a histogram for this scope; a user click
+                        // overrides it (and survives re-renders).
+                        const median = FlamegraphHistogram.sampleWeightedMedianNs(bars);
+                        if (splitNs == null) splitNs = median;
```
Fluzko, line 919: [The comment above](https://github.com/dial9-rs/dial9/pull/663/changes#diff-4bb93a174fcccc5bb88c7ef45c8e5089a229c1f942794cfac24342723c532ca5R852-R853) says split is _"seeded to the sample-weighted median on each new response"_, but this only seeds when `splitNs == null` (once I guess). After changing service/time/facets and re-streaming, the fast/slow split line stays pinned to the previous scope's median, which can sit off the new histogram. I would either reset `splitNs = null` on scope change, or fix the comment.

Fluzko (approved):

PR #664, opened 2026-07-10
feat(viewer): v0 of flamegraph inspect

### LINKED ISSUE #652: Add "inspect" support to flamegraph frames
"inspection" loads a given frame and shows across the _entire flamegraph_, all the paths into and out of the frame.

The frames below demonstrate the weight of those traces contributing upwards into the flamegraph and the frames above show the aggregated weight of frames above that frame.

For example, selecting `tracing::event::dispatch` then right clicking "inspect" would show you all the call paths into tracing::event::dispatch below, then the actual dependency paths of tracing::event::dispatch above

### LINKED ISSUE #653: improve flamegraph search
Add the ability to search for frames and see a list of query results that includes the size of each matched frame.

From each frame, you should be able to click and enter #652 mode of that frame.

rcoh:
This adds an "inspect" mode to the flamegraph + improves the search functionality.

Inspect shows a butterfly that allows you to focus on a frame e.g. `tracing::event::Dispatch` and see all paths into it and out of it.

Screenshot 2026-07-13 at 11 07 29 AM
Screenshot 2026-07-13 at 11 07 15 AM

Fixes #653 
Fixes #652

### DIFF
--- a/dial9-viewer/ui/flamegraph.js
+++ b/dial9-viewer/ui/flamegraph.js
@@ -763,6 +763,9 @@
       unpinTooltip();
       renderAll();
       renderBreadcrumb();
+      // Entering/re-pivoting inspect is a view-state change, so notify the host
+      // (flamegraph.html) to persist the new focus into the URL for deep links.
+      onZoomChange();
     }
 
     // Clear all inspect state without triggering a re-render. Used both by the
@@ -785,6 +788,8 @@
       setInspectVisible(false);
       renderAll();
       renderBreadcrumb();
+      // Leaving inspect clears the persisted focus from the URL.
+      onZoomChange();
     }
 
     function renderInspect() {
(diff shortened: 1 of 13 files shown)


dial9-viewer/ui/flamegraph_view_state.js
Fluzko, line 1: This is not being imported/used, am I wrong?

Fluzko: It also encodes the inspect focus differently from the code that ships. The module writes `inspect` (name) + `inspect_full` (symbol). The live path (`writeViewState` in flamegraph.html, `getInspectFocus`) writes a single `inspect` = `fullName || name` and never emits `inspect_full`.

dial9-viewer/ui/flamegraph.js
```diff
@@ -431,12 +718,145 @@
...
+      inspectCalleesData = null;
+      inspectCallersData = null;
+      inspectHistory = [];
+      hitRegions.callees = [];
+      hitRegions.callers = [];
+    }
+
+    function exitInspect() {
+      if (!inspectActive) return;
+      resetInspectState();
```
Fluzko, line 803: `applyFilters()` calls `resetInspectState()` (which only clears state, no DOM toggle) then `renderAll()`. Neither hides `inspectView`. I think this leaves a stale state, right?

dial9-viewer/ui/flamegraph_view_state.js
rcoh: fixed

Fluzko (approved):

PR #666, opened 2026-07-14
viewer: split serve() into build_app() so embedders own binding

rcoh:
## Summary

`serve()` both assembled the axum app AND bound the listener, so an embedder could not add routes or wrap middleware (e.g. auth) around the whole app without duplicating the ~200-line assembly.

Split it into:
- **`pub build_app(ViewerConfig) -> axum::Router`** — assembles routes/state/aggregation and RETURNS the router (which is a `tower::Service`)
- Caller-owned binding in `cli::run` (standalone binary behavior is unchanged)

### Changes
- `ServeConfig` → `pub ViewerConfig` (+`Default`); drop `port` (a binding concern) and `local` (a logging/metrics concern)
- Logging + per-request metrics become caller-owned process concerns: `pub init_tracing(local)`, `pub shutdown_signal()`, and a re-exported `pub attach_request_metrics(local)` (callers can also supply their own metrique sink)

### Why

This unblocks embedders adding first-party routes (`.merge`) and auth middleware (`.layer`) around the whole app without forking `serve()`. Embedders can now:

```rust
let app = build_app(cfg).await?;
let app = app.merge(my_routes).layer(my_auth);
let listener = TcpListener::bind(addr).await?;
axum::serve(listener, app).await?;
```

### DIFF
--- a/dial9-viewer/src/lib.rs
+++ b/dial9-viewer/src/lib.rs
@@ -6,6 +6,11 @@ pub mod storage;
 
 pub use report_serve::report_serve_router;
 
+// Expose the standard per-request metrics sink so a caller embedding
+// `build_app` can attach it — or supply their own metrique sink instead.
+// Metrics are a process-global concern the caller owns, like logging.
+pub use server::metrics::attach_request_metrics;
+
 use std::path::PathBuf;
 
 async fn detect_bucket_region(bucket: &str) -> Option<String> {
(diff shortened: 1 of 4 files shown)


dial9-viewer/src/lib.rs
```diff
@@ -93,11 +106,35 @@ pub(crate) async fn serve(
             .with_env_filter(env_filter())
             .init();
```
jasgin, line 107: Should we be returning a result w/ try_init?

dial9-viewer/src/lib.rs
```diff
@@ -270,24 +307,13 @@ pub(crate) async fn serve(
...
-
-    Ok(())
+    Ok(app)
 }
 
-async fn shutdown_signal() {
+/// Wait for a shutdown signal (Ctrl-C), then resolve. Pass this to
+/// `axum::serve(...).with_graceful_shutdown(...)` when binding a router from
+/// [`build_app`].
+pub async fn shutdown_signal() {
```
jasgin, line 316: Should this also return a result?

jasgin (approved):

dial9-viewer/src/lib.rs
rcoh: in general during startup, you're just going to panic anyway; I'd kinda rather just panic and find out

PR #668, opened 2026-07-14
Drop aws-sdk-s3-transfer-manager, upload segments via aws-sdk-s3 PutObject

### LINKED ISSUE #659: worker-s3 pulls in rustls-webpki 0.101.7 (RUSTSEC-2026-0098/-0099/-0104), blocking downstream cargo-deny
Enabling the `worker-s3` feature pulls in `rustls-webpki` 0.101.7, which carries three advisories. This makes the feature unusable for any downstream that gates CI on `cargo deny` or GitHub's Dependency Review.

## Chain

```
dial9-tokio-telemetry (feature "worker-s3")
└── aws-sdk-s3-transfer-manager 0.1.3     <- latest release on crates.io
    └── aws-smithy-http-client 1.2.0      <- latest release on crates.io
        ├── hyper-rustls 0.24.2
        │   └── rustls-webpki 0.101.7     <- vulnerable
        └── tokio-rustls 0.24.1
            └── rustls 0.21.12
```

Advisories against `rustls-webpki` 0.101.7:

- [RUSTSEC-2026-0098](https://rustsec.org/advisories/RUSTSEC-2026-0098) — name constraints for URI names were incorrectly accepted ([GHSA-965h-392x-2mh5](https://github.com/rustls/webpki/security/advisories/GHSA-965h-392x-2mh5))
- [RUSTSEC-2026-0099](https://rustsec.org/advisories/RUSTSEC-2026-0099) — name constraints accepted for certificates asserting a wildcard name ([GHSA-xgp8-3hg3-c2mh](https://github.com/rustls/webpki/security/advisories/GHSA-xgp8-3hg3-c2mh))
- [RUSTSEC-2026-0104](https://rustsec.org/advisories/RUSTSEC-2026-0104) — reachable panic in certificate revocation list parsing

Enabling `worker-s3` also drags in the `hyper` 0.14 / `h2` 0.3 / `sct` stack alongside the `hyper` 1.x that the rest of a modern dependency graph already resolves to.

## Why downstream cannot work around it

Cargo's feature unification can add features but cannot drop a transitive dependency, so a downstream crate has no way to steer `aws-smithy-http-client` onto a current rustls. And there is nothing to upgrade *to*: `aws-sdk-s3-transfer-manager` 0.1.3 and `aws-smithy-http-client` 1.2.0 are both the newest published versions.

Concretely: the rest of our workspace already resolves to the safe `rustls-webpki` 0.103.13 and `hyper-rustls` 0.27.9. Turning on `worker-s3` adds a *second*, vulnerable TLS stack purely for trace upload.

## Impact

We are integrating dial9 into [RustFS](https://github.com/rustfs/rustfs). We wired up `with_s3_uploader`, and `cargo deny` plus CI's Dependency Review both rejected the branch. Since we could not fix it downstream, we removed S3 upload entirely; operators now copy sealed segments out of the output directory by hand.

I would guess this has gone unnoticed because `cargo audit` / `cargo deny` do not enable optional features by default — the repo's own RUSTSEC issues (#628, #618, #609, #608, #565) all come from the default feature set. `cargo deny --all-features check advisories` reproduces it.

## Possible directions

1. Let downstream select the TLS backend for the S3 pipeline (e.g. re-export `aws-smithy-http-client`'s rustls/native-tls features), so a current rustls can be chosen.
2. Drop `aws-sdk-s3-transfer-manager` and use `aws-sdk-s3` directly for segment upload. Segments are modest, sequentially produced files; the transfer manager's multipart machinery may not be earning its dependency cost.
3. If neither is near-term, consider documenting `worker-s3` as carrying known advisories, so downstreams do not discover it in CI.

Option 2 looks cleanest from the outside, but I do not know what the transfer manager buys you here.

## Reproduction

```bash
cargo deny --all-features check advisories
```

## Environment

`dial9-tokio-telemetry` 0.3.14, `cargo-deny` with the default advisory database, Linux x86_64 and macOS aarch64.

houseme:
## Summary

Fixes #659. Enabling `worker-s3` pulled in `aws-sdk-s3-transfer-manager`, which dragged a legacy `rustls 0.21` / `hyper 0.14` TLS stack and `rustls-webpki 0.101.7` into the graph, carrying **RUSTSEC-2026-0098 / -0099 / -0104**. Downstreams gating CI on `cargo-deny` or GitHub Dependency Review (e.g. RustFS) could not steer the transitive dependency onto a current rustls, so turning on S3 upload was a non-starter.

Per the discussion in #659 (transfer manager was an experiment, not earning its keep), this takes **option 2**: drop the transfer manager and upload with `aws-sdk-s3` directly.

## What changed

The transfer manager was only ever used internally in `dial9-utils`; every public API already takes `aws_sdk_s3::Client`, so **this is not a breaking change**. Segments are modest, sequentially produced files, so a single `PutObject` per segment (and per manifest) replaces the transfer manager's multipart machinery.

- `S3Uploader` now holds an `aws_sdk_s3::Client`; `upload_and_delete` and `upload_manifest` issue `put_object(...).send()`, streaming the payload `Bytes` into a `ByteStream` with no extra copy.
- New `put_error_kind` classifies `SdkError<PutObjectError>` retryability — transport failures (timeout / dispatch / unparseable response) and service `5xx`/`429` are retryable; `4xx` are terminal — preserving the circuit-breaker/retry semantics the transfer-manager classifier provided.
- Removed `aws-sdk-s3-transfer-manager` from both crates and the `worker-s3` feature; dropped the now-unused `bytes` dep from `dial9-utils`.
- Deleted `.cargo/audit.toml`: the three advisories it ignored are no longer reachable. `s3s-aws 0.14.1` already uses the modern rustls chain, so the transfer manager was the last source of `rustls-webpki 0.101.7`.

## Result

The full workspace graph (`cargo tree --workspace --all-features`) no longer resolves `rustls-webpki 0.101.7`, `rustls 0.21`, or `aws-sdk-s3-transfer-manager`. `worker-s3` builds now share the safe `rustls-webpki 0.103` / `hyper-rustls 0.27` stack the rest of a modern graph already uses, so `cargo deny --all-features check advisories` passes.

## Testing

- `cargo nextest run -p dial9-utils` — 32 passed; `--stress-duration 20s` — 6 iterations, no flakes.
- `cargo nextest run -p dial9-tokio-telemetry --features worker-s3 --test s3_integration` — 9 passed (end-to-end trace→S3 roundtrip, region auto-detection, dump manifests, injected `5xx` failures, hangs, permanent-failure metrics).
- `cargo fmt --check` and `cargo clippy` clean on the touched crates.


dial9-utils/src/s3.rs
```diff
@@ -557,14 +571,8 @@ impl S3PipelineUploader {
             .build();
```
jasgin, line 596: Irrelevant to this change, but we should probably add an operation timeout here

jasgin (approved):

dial9-utils/src/s3.rs
houseme: Good catch — confirmed there was no operation timeout on the uploader's client, so a hung `PutObject`/`HeadBucket` could stall the upload worker indefinitely.

Addressed in 3ac75b8 (kept as a separate commit since it's independent of the transfer-manager removal):

- Added an `operation_attempt_timeout` builder field to `S3Config` (default 30s), wired into the client we build ourselves via `TimeoutConfig`.
- Chose a **per-attempt** timeout rather than a whole-operation one, so each HTTP attempt is bounded without truncating the SDK retry policy or the pipeline circuit breaker that re-drive retryable aborts.
- Only the client built on the `.s3(..)` path gets it; a caller-supplied client passed via `.s3_with_client(..)` keeps its own timeout config untouched.

Two calls I'd welcome your take on: (1) per-attempt vs. a whole-operation ceiling, and (2) scoping it to the self-built client only. Happy to adjust either if you'd prefer.

PR #670, opened 2026-07-15
feat(viewer): Tokio Stats "Worker activity" rollup + focus deep-link

prabrat:
Added`workers` rollup (fallback path: poll counts + share%) into the Tokio Stats aggregate page. Shows per-worker poll distribution to surface work imbalance across the runtime's thread pool.

Server: `worker_activity: Vec<WorkerStats>` on TokioStatsResponse, accumulated via `by_worker: HashMap<u32, WorkerAccum>` — counts ALL polls (before the 100µs floor) for accurate share%, then tracks notable polls + worst exemplar for above-floor rows. Old part-files without worker_id column silently skipped (same guard as long-polls).

UI: `renderWorkerActivity` card with worker/total/share/bar/notable/ worst-poll columns; rows deep-link to the worst poll's trace segment. `workerShareHeat(sharePct, numWorkers)` helper colors imbalance (>2× ideal red, >1.5× amber, else green). Each worker can be clicked upon as well, taking the user to the instance of the longest poll on that worker.

Screenshot 2026-07-16 at 11 50 21 AM

### DIFF
--- a/dial9-viewer/ui/tokio_stats.html
+++ b/dial9-viewer/ui/tokio_stats.html
@@ -422,9 +422,12 @@
   }
 
   // "Worker activity": per-host busyness summary with expandable per-worker
-  // detail. Busyness = sum(poll durations) / worker's observed time span.
-  // One row per host (runtime), click to expand individual workers. Each worker
-  // row shows a verification breakdown: busy_ns / span_ns = busy_pct.
+  // detail. Per-worker busyness = sum(poll durations) / worker's observed time
+  // span; a host's busyness is the mean of its workers' busyness (bounded, and
+  // equal to the average of the per-worker rows shown on expand — see
+  // hostBusyPct). One row per host (runtime), click to expand individual
+  // workers. Each worker row shows a verification breakdown: busy_ns / span_ns
+  // = busy_pct.
   let workerSortKey = "busy_pct"; // sort key for host rows
   let workerSortDesc = true;
   let expandedHosts = new Set(); // hosts whose worker details are visible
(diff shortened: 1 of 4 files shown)


dial9-viewer/ui/tokio_stats.html
```diff
@@ -416,10 +416,137 @@ <h1>⚡ Tokio Stats</h1>
...
+      hostMap[h].push(w);
+    }
+
+    // Compute per-host aggregates.
+    const hostRows = Object.entries(hostMap).map(([host, ws]) => {
+      const totalPolls = ws.reduce((s, w) => s + w.total_polls, 0);
+      const busyNs = ws.reduce((s, w) => s + w.busy_ns, 0);
+      const notablePolls = ws.reduce((s, w) => s + w.notable_polls, 0);
+      const maxSpanNs = Math.max(...ws.map(w => w.span_ns || 0));
+      const busyPct = maxSpanNs > 0 ? (busyNs / maxSpanNs) * 100 : 0;
```
rcoh: maxSpanNs is `Max` but `busyNs` is summed. This means that busyNs the ratio can go over 100% (or more misleading, a runtime w/ 64 workers shows as "100% saturated" if each worker is 1/64th saturated.

dial9-viewer/ui/tokio_stats.html
```diff
@@ -416,10 +416,137 @@ <h1>⚡ Tokio Stats</h1>
...
+      const busyColor = busynessHeat(hr.busyPct);
+      const barWidth = Math.min(hr.busyPct, 100);
+      const barHtml = `<div style="background:${busyColor}33;border-radius:2px;height:14px;width:${barWidth}%"></div>`;
+      const shareFmt = hr.pollSharePct.toFixed(1) + "%";
+      const worstDur = hr.worstPollNs > 0
+        ? `<span style="color:${latencyHeat(hr.worstPollNs)};font-weight:600">${formatDuration(hr.worstPollNs)}</span>`
+        : "—";
+      const hostLabel = escapeHtml(hr.host);
+
+      html += `<tr style="cursor:pointer;font-weight:600" onclick="toggleWorkerHost('${hostLabel}')">
```
rcoh: this is not html escaped properly and a single `'` in hostLabel breaks it

dial9-viewer/ui/tokio_stats.html
```diff
@@ -416,10 +416,137 @@ <h1>⚡ Tokio Stats</h1>
...
+
+    // Compute per-host aggregates.
+    const hostRows = Object.entries(hostMap).map(([host, ws]) => {
+      const totalPolls = ws.reduce((s, w) => s + w.total_polls, 0);
+      const busyNs = ws.reduce((s, w) => s + w.busy_ns, 0);
+      const notablePolls = ws.reduce((s, w) => s + w.notable_polls, 0);
+      const maxSpanNs = Math.max(...ws.map(w => w.span_ns || 0));
+      const busyPct = maxSpanNs > 0 ? (busyNs / maxSpanNs) * 100 : 0;
+      const worstPollNs = Math.max(...ws.map(w => w.worst_poll_ns || 0));
+      const worstWorker = ws.reduce((best, w) => w.worst_poll_ns > (best.worst_poll_ns || 0) ? w : best, ws[0]);
```
rcoh: this line of code is totally dead

rcoh (changes requested):

rcoh:
this works decently well. It would be nice to follow up by using the segment metadata about workers to resolve the total worker count. Right now, its using the number of workers that are _observed_ but that is an undercount of how many workers are actually _available_

rcoh:
What does the `Share` column mean? What does the `Notable` column mean?

prabrat:
> What does the `Share` column mean? What does the `Notable` column mean?

Not sure why its not letting me reply normally, but Share and Notable are both for polls; poll share and notable polls (>1ms)

prabrat:
> this works decently well. It would be nice to follow up by using the segment metadata about workers to resolve the total worker count. Right now, its using the number of workers that are _observed_ but that is an undercount of how many workers are actually _available_

Hmm, didn't think about this. I can look into this and add a follow up pr for this, unless you'd want to include it in this one

rcoh:
The busy percentage isn't really accurate because it reports a node that happens to only be running fewer works as more busy...its actually basically backwards and very misleading
Screenshot 2026-07-17 at 12 06 45 PM

rcoh:
ideally we would show something like Workers (Active / Total) `23 / 64`

I still don't know what "Poll share" means

prabrat:
> The busy percentage isn't really accurate because it reports a node that happens to only be running fewer works as more busy...its actually basically backwards and very misleading Screenshot 2026-07-17 at 12 06 45 PM

Oh thats very weird. I didn't notice this, but I was also testing on a smaller/more basic application. Looking into it now

rcoh:
nice! works pretty well. I think we can continue to enhance it as well

rcoh (approved):

PR #671, opened 2026-07-16
Add local mode for aggregation

rcoh:
When running aggregation locally, its convenient to avoid needing to persist the aggregates to S3. This adds a default mode so that aggregates go to a tempdir instead.

I cleaned up some other arguments while I was there

### DIFF
--- a/dial9-viewer/tests/aggregate_test.rs
+++ b/dial9-viewer/tests/aggregate_test.rs
@@ -523,20 +523,34 @@ impl dial9_viewer::storage::StorageBackend for FailingPuts {
 }
 
 /// Start a server WITHOUT a server-side `AggContext`, exercising the
-/// bring-your-own-credentials `/api/flamegraph?bucket=…` path instead. The
-/// output bucket is configured separately (as `--agg-output-bucket` does), so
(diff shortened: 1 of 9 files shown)


dial9-viewer/tests/aggregate_test.rs
```diff
@@ -1393,29 +1407,43 @@ async fn byoc_writes_output_to_configured_bucket_not_source() {
...
 
-/// When NO output bucket is configured, BYOC aggregation falls back to writing
-/// into the source bucket (the historical behavior, valid when the source is
-/// writable). Guards the fallback branch of the output-bucket routing.
+/// With no output bucket, BYOC aggregation must use the server-local cache and
+/// never attempt to write through source credentials. The injected source
+/// backend rejects every PUT, matching a caller-provided read-only bucket.
 #[tokio::test]
-async fn byoc_without_output_bucket_writes_to_source() {
+async fn byoc_without_output_bucket_uses_temporary_cache_for_read_only_source() {
```
jasgin, line 1414: > The injected source backend rejects every PUT

Meaning this case was broken before?

jasgin (approved):

Issue #496, opened 2026-06-04
sample active-tasks metric to anchor the number of active tasks properly


rcoh:
[metric](https://docs.rs/tokio/latest/tokio/runtime/struct.RuntimeMetrics.html#method.num_alive_tasks)

today, dial9 derives a task count from TaskSpawn / TaskTerminate events but this doesn't work well for traces that are not complete timelines o the application. We should sample the task count at the same time we sample the global queue.

PR #675, opened 2026-07-16
Added active task metrics via QueueSampleEvent (#496)

Scriptize:
This commit aims to address issue #496

### DIFF
--- a/dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
+++ b/dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
@@ -227,9 +227,15 @@ impl Source for TokioRuntimesSource {
             .iter()
             .map(|m| m.global_queue_depth())
             .sum();
+        let total_active_tasks: usize = self
+            .runtime_metrics
+            .iter()
+            .map(|m| m.num_alive_tasks())
+            .sum();
         ctx.record_event(&QueueSampleEvent {
             timestamp_ns: clock_monotonic_ns(),
             global_queue: total_global_queue as u8,
+            active_tasks: total_active_tasks as u64,
         });
     }
 
--- a/dial9-tokio-telemetry/tests/active_tasks_metric.rs
+++ b/dial9-tokio-telemetry/tests/active_tasks_metric.rs
@@ -1,0 +1,142 @@
+//! Tests that `QueueSampleEvent.active_tasks` reflects the number of alive tasks
+//! via `RuntimeMetrics::num_alive_tasks()`.
+
+mod common;
+
+use common::{CAPTURE_BUFFER_SIZE, capture_processor, decode_all};
+use dial9_tokio_telemetry::telemetry::analysis_events::Dial9Event;
+use dial9_tokio_telemetry::telemetry::{
+    MemoryBuffer, RecorderPipelineExt, RecorderTokioExt, TokioAttachOptions, recorder,
+};
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::sync::Barrier;
+
+/// Spawn a known number of long-lived tasks, verify that `QueueSampleEvent`
+/// reports at least that many active via `RuntimeMetrics::num_alive_tasks()`.
+#[test]
+fn active_tasks_reflects_spawned_tasks_via_runtime_metrics() {
+    let (capture, batches) = capture_processor();
+    let num_tasks: usize = 20;
+
+    let rec = recorder(MemoryBuffer::new(CAPTURE_BUFFER_SIZE).unwrap())
+        .with_custom_pipeline(|p| p.pipe(capture))
+        .build();
+    let (rec, rt) = rec
+        .attach_tokio_runtime_with(
+            TokioAttachOptions::builder()
+                .task_tracking_enabled(true)
+                .build(),
+            |t| {
+                t.worker_threads(2);
+            },
+        )
+        .expect("build tokio runtime");
+
+    rt.block_on(async {
+        let barrier = Arc::new(Barrier::new(num_tasks + 1));
+        let mut handles = Vec::new();
+        for _ in 0..num_tasks {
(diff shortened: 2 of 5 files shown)


dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -156,16 +161,19 @@ impl Source for TokioRuntimesSource {
...
         self.last_sample = time_source().instant();
-        let total_global_queue: usize = {
+        let (total_global_queue, total_active_tasks) = {
             let contexts = self.contexts.lock().unwrap();
             if contexts.is_empty() {
                 return;
             }
-            contexts.iter().map(|c| c.global_queue_depth()).sum()
+            let gq: usize = contexts.iter().map(|c| c.global_queue_depth()).sum();
+            let at: u32 = contexts.iter().map(|c| c.active_tasks.load(Relaxed)).sum();
```
rcoh: Rather than maintaining a counter, what we _actually_ need to be doing here is reading the Tokio metric for the number of active tasks so we have a correct absolute count

rcoh (changes requested):

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -246,6 +249,14 @@ impl RuntimeContext {
             .unwrap_or(0)
     }
 
+    ///Active tasks in the runtime.
+    pub(crate) fn active_tasks(&self) -> usize {
+        self.metrics_and_base
+            .get()
+            .map(|(m, _)| m.num_alive_tasks())
+            .unwrap_or(0)
```
rcoh: When is this zero? Generally, avoid defaulting to zero when it is not semantically accurate

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -156,16 +156,19 @@ impl Source for TokioRuntimesSource {
...
             }
-            contexts.iter().map(|c| c.global_queue_depth()).sum()
+            let gq: usize = contexts.iter().map(|c| c.global_queue_depth()).sum();
+            let at: usize = contexts.iter().map(|c| c.active_tasks()).sum();
+            (gq, at)
         };
         ctx.record_event(&QueueSampleEvent {
             timestamp_ns: clock_monotonic_ns(),
             global_queue: total_global_queue as u8,
+            active_tasks: total_active_tasks as u32,
```
rcoh: Why are you casting to u32? u64 uses variable size so it's actually smaller and avoids truncation

Scriptize: Really? I thought all the integer types were fixed width, unless theres something going on I'm not familiar with. Should I just change the type of active_tasks in the QueueSampleEvent struct to u64?

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
Scriptize: I'm thinking if metrics_and_base haven't been initialized? I see theres a OnceLock around it and its get method returns an option, so its possible for it to have not been set up yet.

rcoh (commented):
Also please add tests

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
rcoh: yes, sorry. In the dial9 format itself, u64 is encoded with LEB128

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
rcoh: active tasks is optional so we should return `None` here instead of defaulting to 0

dial9-tokio-telemetry/tests/active_tasks_metric.rs
```diff
@@ -0,0 +1,131 @@
...
+        })
+        .collect();
+
+    assert!(
+        !queue_samples.is_empty(),
+        "expected at least one QueueSampleEvent"
+    );
+
+    // After tasks finish, active_tasks should be low (block_on + timer ≈ 1-3).
+    let last_active = queue_samples.last().unwrap().active_tasks.unwrap_or(0);
```
rcoh: this is exactly where you should not `.unwrap_or(0)`. This test would succeed if you were never successfully able to actually record an active tasks count.

dial9-tokio-telemetry/tests/active_tasks_metric.rs
```diff
@@ -0,0 +1,131 @@
...
+            handles.push(tokio::spawn(async {
+                tokio::task::yield_now().await;
+            }));
+        }
+        for h in handles {
+            h.await.unwrap();
+        }
+
+        // Wait for the flush source to sample after all tasks are done.
+        tokio::time::sleep(Duration::from_millis(100)).await;
```
rcoh: you don't need to wait here because you use graceful_shutdown below

dial9-tokio-telemetry/tests/active_tasks_metric.rs
```diff
@@ -0,0 +1,131 @@
...
+    );
+
+    // After tasks finish, active_tasks should be low (block_on + timer ≈ 1-3).
+    let last_active = queue_samples.last().unwrap().active_tasks.unwrap_or(0);
+
+    assert!(
+        last_active < num_tasks as u32,
+        "expected active_tasks to decrease after tasks complete, \
+         but last sample was {last_active} (spawned {num_tasks})"
+    );
```
rcoh, line 141: weak assertion

rcoh (changes requested):

rcoh (approved):

PR #679, opened 2026-07-17
fix(taskdump): propagate Ready from the capture re-poll

### LINKED ISSUE #678: Race in `TaskDumped`: if capture re-poll returns `Ready`, it's discarded
## Summary

- To capture a task dump, `TaskDumped` polls a future a second time after it returns `Pending`, and discards the result (`task_dumped.rs:293`, `let _ = inner.poll(cx)`). The re-poll is only expected to produce a spurious wake; a `Ready` result is not handled.
- On a multi-threaded runtime the re-poll can legitimately return `Ready` (the resource the future is parked on may be readied by another thread between the primary poll and the capture re-poll). This is a race.
- A `Ready` cannot always be discarded: per the `Future` contract the future must not be polled again, so the next poll is a **poll-after-`Ready`**, which panics.

## Detail

`TaskDumped::poll` re-polls the inner future on a `Pending` primary poll and drops the result:

```rust
// TaskDumped::poll
let result = this.inner.as_mut().poll(cx);              // :177  primary poll -> Pending
match &result {
    Poll::Ready(_) => { /* cleanup */ }
    Poll::Pending => {
        if *this.just_captured { *this.just_captured = false; }
        else {
            this.frames.capture(this.inner.as_mut(), cx); // :198  re-poll to sample a backtrace
            *this.just_captured = true;
        }
    }
}
result                                                  // Pending returned even future
```

```rust
// FrameBuf::capture
tokio::runtime::dump::trace_with(
    || { let _ = inner.poll(cx); },                     // :293  second poll; a `Ready` here is DISCARDED
    |meta| { /* collect frames */ },
);
```                                                                                                                                                                            
The pending wake from the primary poll reschedules the task, so `:177` polls the now-completed future on the next tick. `just_captured` (#375) guards the *capture*, not the primary poll at `:177`.

## Impact

A poll after `Ready` panics for most futures (`async fn resumed after completion`, `Map must not be polled after it returned Poll::Ready`) or silently hangs a fused one. Under `panic = "abort"` it aborts the whole process.

## Detection

I enabled Dial9 spawn tracing in my DataFusion code base and observed panics:

```
thread 'tokio-rt-worker' panicked at futures-util-0.3.32/src/future/future/map.rs:62:17:
Map must not be polled after it returned `Poll::Ready`
  17: poll<Pin<Box<dyn Future..>>, trace_future::{closure_env#1}<(), ..::run_worker::{async_fn_env#0}>, ()>
        at futures-util-0.3.32/src/future/future/map.rs:62:17
        at dial9-tokio-telemetry-0.3.14/src/task_dumped.rs:137   <- primary poll of an already-completed
                                                                    future (0.3.14; `:177` on `main`)
  20: poll<dial9_tokio_telemetry::task_dumped::TaskDumped<..>>   at ../src/traced.rs:218
  22: poll<dial9_tokio_telemetry::traced::TracedFuture<..>>
  23: poll<tokio::runtime::task::trace::Root<..TracedFuture<..>>>
```

## Reproduction

A contract-abiding future that completes on its 2nd poll (the capture re-poll) triggers it deterministically when spawned via the dial9 spawner with task dumps enabled:

```rust
struct CompletesOnSecondPoll { polls: u32 }
impl Future for CompletesOnSecondPoll {
    type Output = ();
    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<()> {
        self.polls += 1;
        match self.polls {
            1 => { cx.waker().wake_by_ref(); Poll::Pending } // primary poll: park + arm wake
            2 => Poll::Ready(()),                            // capture re-poll: completes; Ready discarded
            n => panic!("polled after Ready (poll #{n})"),   // next tick: poll-after-Ready
        }
    }
}
// handle.spawn(CompletesOnSecondPoll { polls: 0 }).await  =>  Err(JoinError::Panic(..))
```

Fails today:

```
future polled again after it returned Ready (poll #3)
JoinError::Panic("future polled again after it returned Ready (poll #3)")
```

## Fix

Have `capture` return its `Poll`; on `Ready`, run the `Ready`-arm cleanup and return it, so the future is never polled again.

mitchsw:
## Which issue does this PR close?

- Closes #678

## Rationale for this change

To sample a task-dump backtrace, TaskDumped polls the inner future a second time and discarded the result. On a multi-threaded runtime that re-poll can legitimately return Ready (a resource the future is parked on is readied by another thread between the primary poll and the capture re-poll). The completion was dropped and Pending returned to the caller.

## What changes are included in this PR?

Makes FrameBuf::capture return the re-poll's Poll, and have TaskDumped::poll adopt a Ready (running the same cleanup as the Ready arm and returning it) so the future is never polled after completion.

Adds a regression test that spawns a future which completes on its second poll (the capture re-poll); it panicked before this fix and passes now.

### DIFF
--- a/dial9-tokio-telemetry/src/task_dumped.rs
+++ b/dial9-tokio-telemetry/src/task_dumped.rs
@@ -195,7 +195,13 @@ impl<F: Future> Future for TaskDumped<F> {
                 if *this.just_captured {
                     *this.just_captured = false;
                 } else {
-                    this.frames.capture(this.inner.as_mut(), cx);
+                    let repoll_result = this.frames.capture(this.inner.as_mut(), cx);
+                    // In rare circumstances, repoll will now be ready.
+                    if repoll_result.is_ready() {
+                        this.frames.clear();
+                        *this.pending_capture_ts = None;
+                        return repoll_result;
+                    }
                     *this.just_captured = true;
                     let poll_end = crate::telemetry::recorder::poll_start_ts_monotonic();
                     *this.pending_capture_ts = NonZeroU64::new(poll_end);
@@ -276,8 +282,11 @@ impl FrameBuf {
     }
 
     /// Capture backtraces at yield points by re-polling `inner` under the
-    /// real waker inside `trace_with`.
-    fn capture<F: Future>(&mut self, inner: Pin<&mut F>, cx: &mut Context<'_>) {
+    /// real waker inside `trace_with`, returning that re-poll's result.
+    ///
+    /// The re-poll can complete `inner`; that `Ready` is returned so the caller
+    /// can adopt it.
+    fn capture<F: Future>(&mut self, inner: Pin<&mut F>, cx: &mut Context<'_>) -> Poll<F::Output> {
         if self.ips.capacity() == 0 {
             self.ips.reserve(FRAME_BUF_INITIAL_CAPACITY);
         }
@@ -288,9 +297,10 @@ impl FrameBuf {
 
         // `trace_with`'s outer closure is `FnOnce`; `Option::take` moves the
         // pinned reference in without requiring a `Copy` bound or unsafe.
+        let mut result = Poll::Pending;
         tokio::runtime::dump::trace_with(
             || {
-                let _ = inner.poll(cx);
+                result = inner.poll(cx);
             },
             |meta| {
                 let ip_start = ips.len();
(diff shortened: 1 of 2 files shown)


rcoh:
This may also be related to the fact that we had to work around spurious double-wakes in Tokio which are now fixed in 1.53. I'm curious if that has anything to do with this.

rcoh (approved):
LGTM! Thanks

PR #680, opened 2026-07-17
feat(viewer): add sanitized trace generator

rcoh:
# Why
Its often very helpful for purposes of working on the viewer to be able to make a fake trace file that is of similar shape to a real one (that may not be able to be shared)

This is real trace in fake trace out; the results are really pretty good as far as I can tell from an overall shape perspective.

### DIFF
--- a/dial9-viewer/src/trace_shape.rs
+++ b/dial9-viewer/src/trace_shape.rs
@@ -6573,7 +6573,7 @@ mod tests {
         );
     }
 
-    #[cfg(unix)]
+    #[cfg(target_os = "linux")]
     #[test]
     fn generate_dev_full_returns_error() {
         // (4) Writing to /dev/full must return an error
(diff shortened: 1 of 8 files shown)


dial9-viewer/src/trace_shape.rs
```diff
@@ -0,0 +1,7665 @@
...
+    string_remap: HashMap<String, String>,
+    next_string_id: u64,
+    // Cardinality tracking
+    worker_ids: HashSet<u64>,
+    task_id_set: HashSet<u64>,
+    // H: Poll quantiles: track starts per (worker_id) using a stack for nested/repeated
+    poll_starts: HashMap<u64, Vec<u64>>,
+    poll_durations: Vec<u64>,
+    // H: Span quantiles: track SpanEnter per (span_id) -> stack of timestamps
+    span_enters: HashMap<u64, Vec<u64>>,
```
Fluzko: These four accumulators are written in the extract remap loop but never read, pretty sure is dead code

Fluzko (approved):

PR #684, opened 2026-07-19
feat(telemetry): record metrique entries into the dial9 trace

Benjscho:
## Summary

Implements #189: dial9 acts as a peer metrique sink. Entries flowing through an existing EMF/JSON metrique pipeline are also recorded into the dial9 trace, correlated by worker id, task id, and a monotonic time span, so a single trace file carries both tokio runtime telemetry and application-level metrics.

- `Dial9Context` (flatten into any `#[metrics]` struct), `Emit`/`Interned` field tags, and a `Dial9Stream` `EntryIoStream` sink, composable via `metrique::writer::stream::tee` alongside an existing sink.
- Implements the architecture in `docs/design/metrique-integration.md`, adapted to the real upstream metrique descriptor/flags API (which landed after the doc was written — `flags(...)`/`default_flags(...)` rather than the doc's illustrative `field_tag(...)` syntax).
- Field dispatch is by resolved field *name*, not position, so `Dial9Context` can be flattened anywhere in the struct (metrique's descriptor order and write-callback order only coincide when a flattened field happens to be declared last — dispatching by name sidesteps that entirely).
- Two behaviors documented as known limitations because they're empirically-discovered constraints of the upstream `ForceFlag`/descriptor machinery, not something fixable on dial9's side: `Vec<T>`/`[T]` fields always land as comma-joined strings (metrique's `ForceFlag` wrapper doesn't forward structured list writes), and `Flex` fields aren't supported (no current metrique `Value` impl produces that shape).

## Test plan

- [x] `cargo nextest run -p dial9-tokio-telemetry --features metrique-integration` (278 tests)
- [x] `cargo nextest run -p dial9-tokio-telemetry --features metrique-integration --stress-duration 20s` (2 iterations, no flakes)
- [x] `cargo fmt --check` / `cargo clippy --all-targets --all-features` clean (no new warnings)
- [x] `cargo test --doc -p dial9-tokio-telemetry --features metrique-integration`
- [x] `examples/metrique_integration.rs` run manually end-to-end (tees to `metrique::local::LocalFormat` stdout + `Dial9Stream`, verified both sinks receive every entry and the resulting trace decodes correctly)


dial9-tokio-telemetry/src/telemetry/metrique_integration/context.rs
```diff
@@ -0,0 +1,95 @@
...
+///
+/// Deliberately no `Default`/`Clone` impl: a zero-valued `Default` would
+/// produce fields that look like validly-captured context (the `Context`
+/// flag is still present) without ever tripping the "no context captured"
+/// diagnostic. Construct with [`capture`](Self::capture) at each use site;
+/// forgetting the field produces a compile error.
+#[metrics(rename_all = "PascalCase")]
+pub struct Dial9Context {
+    #[metrics(flags(Context), no_close)]
+    worker_id: WorkerId,
```
rcoh, line 52: worker_id can actually change across requests so including it can actually be somewhat misleading

dial9-tokio-telemetry/src/telemetry/metrique_integration/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+//!
+//! Flatten [`Dial9Context`] into your entry — anywhere; field order doesn't
+//! matter, see "Field order" below — and tag the fields you want in the
+//! dial9 trace with [`Emit`] (directly, or via
+//! `#[metrics(default_flags(Emit))]` at the struct level — struct-level
+//! defaults don't propagate across the flatten boundary, so `Dial9Context`'s
+//! own fields are unaffected either way):
+//!
+//! ```
+//! use dial9_tokio_telemetry::telemetry::metrique_integration::{Dial9Context, Emit, Interned};
```
rcoh, line 30: we should probably move this into dial9-util to decouple it from the tokio stuff

dial9-tokio-telemetry/src/telemetry/metrique_integration/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+//!
+//! Dial9 acts as a peer metrique sink: entries flowing through your
+//! existing EMF/JSON metrique pipeline are also recorded into the dial9
+//! trace, correlated by worker id, task id, and a monotonic time span. A
+//! single trace file then carries both tokio runtime telemetry and your
+//! application's own request-scoped metrics.
+//!
+//! # Opting a field in
+//!
+//! Flatten [`Dial9Context`] into your entry — anywhere; field order doesn't
```
rcoh, line 19: this describes opting a field in but the example below appears to show opting a field out

dial9-tokio-telemetry/src/telemetry/metrique_integration/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+//! let _m = RequestMetrics {
+//!     dial9: Dial9Context::capture(),
+//!     route: "/api/users".into(),
+//!     operation: "GetUser",
+//!     debug_blob: String::new(),
+//! };
+//! ```
+//!
+//! [`Interned`] routes a string-shaped field through dial9's string pool.
+//! `flags(skip(Emit))` excludes a field the struct-level default would
```
rcoh: struct-level default? ah...`default_flags(emit)`

I think we should probably have default all fields with explicit opt-out

dial9-tokio-telemetry/src/telemetry/metrique_integration/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+//!   wrapper's own inherited default (comma-join each element, then
+//!   `.string()`) before it reaches this sink — whether or not the entry is
+//!   boxed. There is no dial9-side workaround; encoding as a plain string
+//!   is the only representation this sink can actually observe.
+//! - **`Flex` fields are not supported.** No metrique `Value` impl
+//!   currently produces `FieldShape::Flex`, and `Flex<T>`'s own descriptor
+//!   support is `Unavailable` — flattening a `Flex` field anywhere in an
+//!   entry currently makes the *whole entry* look hand-written to this
+//!   sink (dropped, not just that field). Avoid flattening `Flex` on any
+//!   entry that also wants dial9 integration.
```
rcoh, line 100: maybe we can follow a followup ticket for this. I was thinking that for any un-modeled fields we could have a trailer with https://docs.rs/dial9-trace-format/latest/dial9_trace_format/types/enum.FieldType.html#variant.OptionalDynamicMap or DynamicMap then put all the other dynamic fields in there

dial9-tokio-telemetry/examples/metrique_integration.rs
```diff
@@ -0,0 +1,69 @@
...
+use dial9_tokio_telemetry::telemetry::metrique_integration::{Dial9Context, Dial9Stream, Emit};
+use metrique::CloseValue;
+use metrique::RootEntry;
+use metrique::local::{LocalFormat, OutputStyle};
+use metrique::unit_of_work::metrics;
+use metrique::writer::EntryIoStream;
+use metrique::writer::format::FormatExt;
+use metrique::writer::stream::tee;
+
+#[metrics(rename_all = "PascalCase", default_flags(Emit))]
```
rcoh, line 33: in the final API, we should structure this so it can be `dial9::Emit`, I think it will be clearer to read that way

dial9-tokio-telemetry/src/telemetry/metrique_integration/schema.rs
```diff
@@ -0,0 +1,438 @@
...
+/// role map. Returns `None` only if descriptors became unavailable between
+/// the `descriptor_cache_key` call and this one (not expected in practice —
+/// entries don't change shape mid-flight — but handled defensively).
+pub(super) fn build(entry: &impl Entry) -> Option<BuildResult> {
+    let Descriptors::Available(descs) = entry.descriptors() else {
+        return None;
+    };
+
+    let mut field_roles = HashMap::new();
+    // Fixed header prefix: worker_id/task_id/monotonic_ns_end always occupy
```
rcoh: any specific reason to do it this way? It seems like if we skipped this, then we wouldn't need any special-casing of dial9's internal fields

dial9-tokio-telemetry/tests/metrique_integration.rs
```diff
@@ -0,0 +1,405 @@
...
+use metrique::writer::core::{FieldShape, KnownShape};
+use metrique::writer::value::{Value, ValueWriter};
+use metrique::writer::{Entry, EntryIoStream, EntryWriter};
+use std::time::Duration;
+
+/// Decode every event across all captured segments into `(schema name,
+/// positional field values)`. The dial9 metrique sink registers a schema
+/// per distinct entry shape at runtime, so there's no compile-time struct to
+/// deserialize into — assert on raw positional values instead.
+fn decode_raw_events(segments: &[Vec<u8>]) -> Vec<(String, Vec<FieldValue>)> {
```
rcoh, line 26: you should be able to use the `derive(Deserialize)` machinery to clean up this test a bit

```
#[derive(Deserialize)
struct TheEventIExpect {
  field1: u64,
  field2: String
}
```

etc.

dial9-utils/src/metrique_integration/stream.rs
```diff
@@ -0,0 +1,469 @@
...
+/// ```
+///
+/// Entries with no descriptor (hand-written `Entry` impls), with `Emit`
+/// fields but no flattened `Dial9Context`, or with unsupported field shapes
+/// are diagnosed (see the module docs) and either dropped or recorded with a
+/// fallback header — they never panic the pipeline.
+#[derive(Debug)]
+pub struct Dial9Stream {
+    handle: Dial9Handle,
+    cache: HashMap<Vec<DescriptorId>, CachedDescriptor>,
```
rcoh, line 54: what (if anything) could cause this map to grow unbounded? is it strictly bounded by the number of distinct metrique record definitions? or is it possible a pathological metrique struct could cause this to balloon?

dial9-utils/src/metrique_integration/stream.rs
```diff
@@ -0,0 +1,469 @@
...
+        if emitted != 0 && emitted.is_multiple_of(DEBUG_SUMMARY_SAMPLE) {
+            tracing::debug!(
+                descriptors_seen = self.stats.descriptors_seen.load(Ordering::Relaxed),
+                descriptors_skipped = self.stats.descriptors_skipped.load(Ordering::Relaxed),
+                events_emitted = emitted,
+                events_dropped = self.stats.events_dropped.load(Ordering::Relaxed),
+                "dial9 metrique sink summary"
+            );
+        }
+    }
```
rcoh, line 177: a little meta, but we might as well emit this to the dial9 trace

rcoh (commented):
nice!!

jlizen:
Hey Ben, closing in favor of #723 , I didn't realize you already had started when I implemented. Mine is a bit further along, I think, since I fixed some of those things upstream in metrique and also did some perf tuning of the encoding path. Picking up some improvements from yours though. 

Feel free to DM me if you'd rather divide and conquer, there is definitely are some things that I'm punting.

Benjscho:
Sounds good!

PR #685, opened 2026-07-20
Add Android CPU Profiling Support

nickrobinson:
## Overview
This PR adds support for enabling CPU profiling on Android devices. Android requires special handling as compared to Linux because the Android runtime owns `SIGSEGV` through `libsigchain`. This PR adds the platform-specific signal and context handling needed for safe frame-pointer unwinding.

In addition this PR also adds a basic CI step to build dial9 against an `aarch64-linux-android` target.

## Testing
I've tested these changes by running the cpu-profiling examples on a connected Android device. In addition we have been running this at Ditto behind a feature flag on production Android devices for the past month. 

Screenshot 2026-07-20 at 5 53 32 AM


rcoh:
This is awesome!!

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
```diff
@@ -1218,65 +1218,30 @@ mod tests {
 
     #[test]
     fn trace_runtime_multiple_runtimes_unique_worker_ids() {
-        use std::collections::HashSet;
```
rcoh, line 1221: thanks for this

examples/android_build.sh
```diff
@@ -0,0 +1,112 @@
...
+        # The specific `examples/${EXAMPLE}.rs` file to be targeting
+        --example "${EXAMPLE}"
+        # We are cross-compiling to `aarch64-linux-android`
+        --target "${TARGET}"
+); "${CARGO_BUILD_CMD[@]}"
+
+# ADB
+
+adb push \
+    "target/$TARGET/$PROFILE/examples/$EXAMPLE" \
```
rcoh: is it intentional that `$PROFILE` is not defined anywhere in this script? The docs don't seem to mention it. Presumably there is a reasonable default for it?

rcoh (commented):
overall this looks awesome. If I'm reading correctly, there is a small bug-ish where if you explictly use the CTimer backend:
```rust
 use dial9::cpu::CpuProfilingConfig;

  #[tokio::main]
  async fn main() {
      let recorder = dial9::TracingRecorder::builder()
          .with_cpu_profiling(CpuProfilingConfig::with_ctimer_backend())
          .build()
          .unwrap();

      // ... run the runtime ...
  }
```

you will bypass the JVM/ART fixes you added and go directly to:
```
  pub(crate) fn start_ctimer_only(config: SamplerConfig) -> io::Result<Self> {
      Self::with_ctimer_process_wide(&config)   // registers the calling thread
  }
```

which registers the calling thread

examples/android_build.sh
nickrobinson: I had ripped this out, but missed a spot. We had a custom PROFILE to include line-tables that I removed. I'll get this cleaned up

Cargo.toml
```diff
@@ -35,3 +35,7 @@ assert2 = "0.4"
 
 [profile.release]
 debug = 1
+
+[patch.crates-io]
```
rcoh: ah; I missed this. This will only apply to our crate. I don't think patches apply transitively. I suspect if we open a release-plz PR on libc we can get the new version released though

nickrobinson: I got lucky with timing here :). It looks like libc cut `0.2.188` yesterday which contains the fixed. I've removed the patch and bumped libc.

rcoh (approved):

PR #690, opened 2026-07-21
feat(profiler): record effective profiling configuration

rcoh:
## Summary
- record the effective CPU profiling backend and event source
- record effective scheduler profiling intervals and related configuration
- cover effective metadata behavior in profiler tests

## Stack
Stacked on #689. PR 2 of 7.

## Validation
- profiler metadata tests
- cumulative all-target/all-feature clippy and 1074-test workspace suite

PR #691, opened 2026-07-21
feat(viewer): persist folded tracing spans

rcoh:
## Summary
- persist resolved tracing spans and sample enclosing-span memberships
- make legacy span pairing correct across recycled IDs and worker migration
- add the spans Parquet part and final format/order guarantees
- retain decoder, persistence, and parser-parity coverage with the data model

## Stack
Stacked on #690. PR 3 of 7.

## Validation
- viewer decoder/aggregate/parser tests
- slice validation: 215 viewer tests across three stress iterations
- cumulative workspace suite: 1074/1074

### DIFF
--- a/dial9-viewer/src/ingest/aggregate.rs
+++ b/dial9-viewer/src/ingest/aggregate.rs
@@ -39,8 +39,10 @@ pub(crate) const ORDER_VERSION: u32 = 1;
 /// Storage-format version baked into the output key path
 /// (`{output_prefix}/v{N}/…`). Bump when changing *what* we persist and we want
 /// a deliberate recompute; reads/writes then target a fresh empty tree that
-/// repopulates lazily. The old tree is abandoned and GC'd out-of-band.
-pub(crate) const SAMPLES_FORMAT_VERSION: u32 = 3;
+/// repopulates lazily. The value is a monotonic cache namespace, not a schema
+/// revision, so skipped values are expected. The old tree is abandoned and
+/// GC'd out-of-band.
+pub const SAMPLES_FORMAT_VERSION: u32 = 6;
 
 /// Default raw-trace segment duration, in seconds. A source file covers
 /// `[epoch, epoch + segment_duration)`; the [`Scope`] time filter pads by this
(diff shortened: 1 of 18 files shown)


dial9-viewer/src/ingest/aggregate.rs
```diff
@@ -40,7 +40,7 @@ pub(crate) const ORDER_VERSION: u32 = 1;
 /// (`{output_prefix}/v{N}/…`). Bump when changing *what* we persist and we want
 /// a deliberate recompute; reads/writes then target a fresh empty tree that
 /// repopulates lazily. The old tree is abandoned and GC'd out-of-band.
-pub(crate) const SAMPLES_FORMAT_VERSION: u32 = 3;
+pub const SAMPLES_FORMAT_VERSION: u32 = 6;
```
jasgin, line 45: Why from 3 to 6?

dial9-viewer/src/ingest/parquet_writer.rs
```diff
@@ -253,9 +293,249 @@ fn polls_schema() -> Arc<Schema> {
...
+    fn to_i64(val: u64) -> anyhow::Result<i64> {
+        i64::try_from(val).map_err(|_| anyhow::anyhow!("timestamp {val} exceeds i64::MAX"))
+    }
+
+    for span in spans {
+        span_uid_builder.append_value(span.span_uid)?;
+        span_type_uid_builder.append_value(span.span_type_uid)?;
+        kind_builder.append_value(span.kind);
+        name_builder.append_value(&span.name);
+        target_builder.append_option(Some(&span.target));
```
jasgin: nit: why does this need to be an Option?

dial9-viewer/src/ingest/aggregate.rs
rcoh: every time I change it I need to bump it; there are artifacts persisted in various buckets so I just need to keep it consistent. this is basically just a nonce in the cache key so we don't reuse artifacts in a different format

dial9-viewer/src/ingest/parquet_writer.rs
rcoh: forward compatibility for spans that don't have a target field. Only the spans derived from tracing have a `target` set

dial9-viewer/src/ingest/aggregate.rs
rcoh: Documented that this is a monotonic cache namespace rather than a schema revision, so skipped values are expected.

dial9-viewer/src/ingest/parquet_writer.rs
rcoh: Kept the Parquet field nullable for future non-tracing span kinds, but changed the current always-present write from `append_option(Some(...))` to `append_value(...)`.

jasgin (approved):

PR #692, opened 2026-07-21
feat(viewer): add generalized span analysis APIs

rcoh:
## Summary
- add the generalized `/api/span-stats` catalog, histogram, facet, and exemplar backend
- add span-filtered flamegraph requests and validation
- share fold-stream selection and coverage behavior across analysis APIs
- add region-aware aggregate reads and transactional part merging

## Stack
Stacked on #691. PR 4 of 7.

## Validation
- `cargo fmt --check`
- `cargo check -p dial9-viewer`
- focused server and aggregate suites: 79/79
- independent boundary review approved

### DIFF
--- a/dial9-viewer/src/server/span_stats.rs
+++ b/dial9-viewer/src/server/span_stats.rs
@@ -1,0 +1,2602 @@
+//! `/api/span-stats` endpoint: stream bounded span statistics over Server-Sent
+//! Events, grouped by span type. Discovers available span types and returns
+//! occurrence-weighted duration histograms, percentiles, and wall-time
+//! composition.
+//!
+//! Uses the same scope/resolve/fold pipeline as `/api/flamegraph`. Reads
+//! `spans/` part-files (written by the v5 fold) for each folded source file.
+mod spans_reader;
+
+#[cfg(test)]
+use spans_reader::parse_map_column;
+use spans_reader::{SpanStatsRow, SpansBatchReader};
+
+use std::collections::HashMap;
+use std::convert::Infallible;
+
+#[cfg(test)]
+use arrow::array::Array;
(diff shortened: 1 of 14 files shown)


dial9-viewer/src/server/span_stats.rs
```diff
@@ -0,0 +1,2615 @@
...
+    fn exemplar_matches(&self, elapsed_ns: i64) -> bool {
+        self.exemplar_min_ns.is_none_or(|min| elapsed_ns >= min)
+            && self.exemplar_max_ns.is_none_or(|max| elapsed_ns <= max)
+    }
+
+    /// Stream one spans part into a bounded staged accumulator, then commit it.
+    /// A malformed later row discards the stage, so no row from that part can
+    /// partially mutate the live aggregate.
+    fn merge_spans_part(&mut self, data: Vec<u8>) -> anyhow::Result<()> {
+        let mut staged = self.clone();
```
jasgin, line 698: This clone seems like it could be quite expensive. Is this an issue / could it be avoided?

rcoh: its not cloning the vec, just the accumulator. It should be possible to get rid of it with shenanigans but I optimized all this code quite a bit to get it to be really snappy on large quantities of data so I'm not super concerned

rcoh: Confirmed this clones the bounded aggregate, not just a lightweight shell. I retained it to preserve transactional behavior for malformed later rows without buffering the full decoded part; a per-part delta also has to preserve global cardinality and first-seen facet semantics. Documented that tradeoff at the callsite.

jasgin (approved):

PR #694, opened 2026-07-21
feat(viewer): add Span Explorer with bounded exemplar refresh

rcoh:
## Summary
- add Span Explorer catalog, histogram, composition, exemplar, and deep-link UI
- add bounded exemplar-only refresh with seed-only folding
- carry folded-set identity through server snapshots and client adoption
- instrument coupled stream phases and register all new JS tests in e2e

## Stack
Stacked on #693. PR 6 of 7.

## Validation
- viewer tests: 289/289
- Span Explorer JS tests: 560 passed
- flamegraph API/deep-link/diff/view-state and trace-analysis JS suites
- full e2e trace script passed on retry; the first generated capture exposed a duplicate-WorkerPark integrity flake and was discarded

PR #695, opened 2026-07-21
docs: define the diff-first unified analysis view

rcoh:
## Summary
- define the diff-first unified analysis model and delivery plan
- document generalized span analysis and its wire contract
- record the server-side diff and span-identity ADRs
- include the design guidance and concept mock without changing implementation

## Stack
Stacked on #694. PR 7 of 7.

## Validation
- `git diff --check`
- exactly 12 documentation/guidance/mock paths; no code
- final Git tree equals `split/source-backup` and untouched `diff-first-design-doc`: `f8f31b9080daf97fdb5f3836e43899612689a991`
- full binary diff between the final stack and both protected sources is empty

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,6 +71,51 @@ rate_limited!(Duration::from_secs(60), {
 ```
 Unguarded logging in loops causes log spam that degrades observability and can itself become a performance problem. One-time paths (startup, shutdown, per-thread init) are exempt.
 
+## Viewer deep links (`/api/object` and `source_key`)
+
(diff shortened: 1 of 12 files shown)


jv1i (approved):
Approach looks good to me!

From a first read (so I might've missed it) I wonder if something needs to be done for the case where the two sides cover different sets of files (different time windows or hosts), and one of them happens to be already folded/cached from an earlier query. Like if it starts uneven we'd want to avoid users reading a difference that's really just one side being further along. Probably easier to see once it's implemented.

PR #696, opened 2026-07-21
Add Axum integrations to dial9-util

rcoh:
## Summary

- add feature-gated Axum 0.7 and 0.8 server integrations to `dial9-util`
- spawn connection, graceful-shutdown, and Hyper internal tasks through dial9
- support caller-provided executors so handles from other dial9 versions can instrument tasks
- replace the metrics service's local `axum_traced` module with the shared integration

## Public API

```rust
dial9_util::dial9_axum::axum_0_8::serve(listener, service)
    .with_executor(move |future| {
        handle.spawn(future);
    })
    .with_graceful_shutdown(shutdown_signal())
    .await?;
```

The default executor resolves dial9 from the current runtime. A custom executor only exchanges a boxed standard `Future`, avoiding a public dependency on a specific dial9 handle version.

## Testing

- `cargo check -p dial9-util --no-default-features --features axum-0-7`
- `cargo check -p dial9-util --no-default-features --features axum-0-8`
- `cargo test -p dial9-util --all-features`
- `cargo check -p metrics-service`
- `cargo nextest run --stress-duration 20s` (2 iterations, 939/939 each)

One unrelated `background_symbolization` timing test failed during the initial full run and passed when rerun in isolation and in both stress iterations.


dial9-utils/src/dial9_axum/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+
+impl Encodable for ConnectionAccepted {
+    fn encode(&self, enc: &mut ThreadLocalEncoder<'_>) {
+        let remote_addr = enc.intern_string(&self.remote_addr);
+        enc.encode(&ConnectionAcceptedWire {
+            timestamp_ns: self.timestamp_ns,
+            remote_addr,
+        });
+    }
+}
```
jv1i, line 133: nit, not for this PR: AFAICT the event in the trace will end up as `ConnectionAcceptedWire` do we want those to just be `ConnectionAccepted` somehow, to match how we name non-manual encode impl events? (I know this is existing and we have other examples doing the same)

makes me wonder if we should eventually support the interning case with something like a derive attribute, since the manual impls are very similar (intern one field, copy the rest)

dial9-utils/src/dial9_axum/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+    timestamp_ns: u64,
+    remote_addr: String,
+}
+
+#[derive(TraceEvent)]
+struct ConnectionAcceptedWire {
+    #[traceevent(timestamp)]
+    timestamp_ns: u64,
+    remote_addr: InternedString,
+}
```
jv1i, line 122: depends on the use case but perhaps splitting remote_addr into ip/port like `TcpAcceptQueueEvent` does could make the interning pay off more? with the client port changing, each full addr only gets used twice (accepted/close) and adds a pool entry. interning just the IP could get us reuse across connections

dial9-utils/src/dial9_axum/axum_0_8.rs
```diff
@@ -0,0 +1,321 @@
...
+            executor,
+            _marker,
+        } = self;
+
+        Box::pin(async move {
+            let (signal_tx, signal_rx) = watch::channel(());
+            executor.execute(Box::pin(async move {
+                signal.await;
+                drop(signal_rx);
+            }));
```
jv1i, line 220: I think this spawn (alongside the use of future::pending) means everyone that skips `with_graceful_shutdown` gets a task their traces that never terminates right? if so perhaps we can avoid spawning? I believe axum handles that case without a spawn (Serve::run https://github.com/tokio-rs/axum/blob/axum-v0.8.9/axum/src/serve/mod.rs#L179-L193)

jv1i (approved):
Looks good! leaving some comments, everything minor (hopefully we get new axum release soon with the new serve API).

Tests: we could test that ConnectionAccepted/Closed make it to a trace.

dial9-utils/src/dial9_axum/mod.rs
rcoh: opened issue

dial9-utils/src/dial9_axum/mod.rs
rcoh: Addressed in eba5336a: TCP addresses are now recorded as an interned IP in remote_addr plus an optional remote_port, while non-socket listener addresses retain the generic fallback.

dial9-utils/src/dial9_axum/axum_0_8.rs
rcoh: Addressed in eba5336a: Serve without graceful shutdown now enters the accept loop directly and does not spawn a pending shutdown task. Both Axum versions have a regression test asserting no executor task is spawned in that path.

jv1i (approved):

PR #698, opened 2026-07-22
refactor!: tokio instrumentation as a source

### LINKED ISSUE #356: dial9 crate structure v2

jv1i:
closes #356, this is the last big piece of the crate refactoring.

### Summary

This PR makes Tokio attach to a built recorder instead of going through a separate traced-runtime builder flow.

```rust
let (recorder, runtime) = dial9::recorder(writer)
    .build()
    .attach_runtime(|t| {
        t.worker_threads(4);
    })?;
```

You can call `attach_runtime` multiple times. Each call adds another runtime and returns the recorder for the next call.

For `#[dial9::main]`, config now naturally returns `std::io::Result<dial9::AttachedRuntime>`:

```rust
fn my_config() -> std::io::Result<dial9::AttachedRuntime> { .. }
```

Shutdown order stays the same: drop runtime(s) first, then call `recorder.graceful_shutdown(timeout)`.

### Main changes

At a high level, this removes the old traced-runtime abstraction and keeps one clear path: build a recorder, then attach one or more runtimes to it.

- Removed traced-runtime API surface (`TracedRuntime`, `TracedRuntimeBuilder`, `build_traced`, `TokioAttachConfig`, and related methods/types).
- `recorder_or_disabled(writer, tokio_fn)` is now `recorder_or_disabled(writer)` returning a `RecorderBuilder`. Sources and pipelines chain onto it as usual; a writer that failed to open still builds a disabled recorder.
- `RecorderBuilder::build()` now starts recording. `build_and_start()` is removed. Use `.paused()` when needed.
- `Recorder::graceful_shutdown` now returns `()`.
- Per-runtime attach settings are in `TokioAttachOptions` (`runtime_name`, task tracking, task dump config, hooks, instrumentation toggle).
- Session-wide settings stay on `RecorderBuilder`.

### Why Tokio attaches after `build()`

`attach_runtime` returns a runtime. A builder method cannot, because `build()` would then have to return the recorder plus one runtime per attach, and that needs the second builder type this PR removes.

Sources differ because they return nothing and hold nothing: the flush thread owns the shared state and passes a `FlushContext` in on every cycle, so registering a source needs no state of its own. Tokio's hooks run on tokio's threads with nothing passed in, so they capture the recorder's handle when they are registered. `TokioRuntimesSource` still lands in the same source list, just pushed after `build()`.

The builder-taking primitive, `attach_tokio_builder(&recorder, &mut builder, options)`, is `pub(crate)`. The public surface is `attach_runtime`, which owns `Builder::build()`. That leaves two cases unexpressible: instrumenting a builder something else will build, and `build_local()` / `LocalRuntime`. Making the primitive public later is additive.

### Pipeline changes

Pipeline behavior also moved closer to the recorder itself, so non-Tokio setups get the same default processing without extra wiring.

- Default pipeline assembly moved into `RecorderBuilder::build()`.
- Sources can contribute processing stages through `Source::segment_processor`.
- CPU symbolization is now source-contributed.
- S3 uploader is wired as terminal stage configuration, so call order with sources no longer matters.
- `RecorderPipelineExt` remains for custom pipeline and S3 shortcuts.

### Runtime/worker behavior updates

Worker ID reservation now happens when runtime work actually starts. That is a better fit for caller-built runtimes and avoids guessing worker counts at attach time.

- Worker ID blocks are now reserved lazily on first poll, not at attach time.
- Runtime metadata (`runtime.{name}` worker mapping) appears after that runtime has actually run work.
- Worker ID block assignment follows worker startup timing, not attach call order.
- Wire format is unchanged.

### Fixes included

While doing the refactor, a few regressions and edge cases surfaced. These are fixed in the same PR.

- `#[dial9::main]` now uses `dial9::block_on`, so the root future is instrumented again.
- `RecorderPipelineExt` regained writer-mode type-state (`RecorderPipelineExt<M>`), so disk-only stages are rejected for memory writers at compile time.
- A thread driving multiple runtimes now registers worker mapping per runtime (TLS guard tracks worker ID, not just a bool).
- Added `dial9::spawn_in(runtime_handle, fut)` for instrumented cross-thread spawn into a specific runtime.

### Migration quick map

Most migrations are mechanical. This table covers the common before/after replacements.

| Before | After |
| --- | --- |
| `recorder(w).with_tokio(f).build()?` | `recorder(w).build().attach_runtime(f)?` |
| `TracedRuntimeBuilder::disabled()` | `dial9::recorder_disabled()` |
| `recorder_or_disabled(writer, \|t\| ..)` | `recorder_or_disabled(writer).build().attach_runtime(\|t\| ..)?` |
| `traced.block_on(fut)` | `dial9::block_on(&runtime, fut)` |
| `traced.tokio_handle(rt.handle())` | `dial9::spawn_in(rt.handle(), fut)` |
| `traced.graceful_shutdown(t)` | `drop(runtime); recorder.graceful_shutdown(t)` |
| `recorder(w).build_and_start()` | `recorder(w).build()` |
| config fn returning `TracedRuntimeBuilder` | config fn returning `std::io::Result<dial9::AttachedRuntime>` |


rcoh:
small nit, lets call it `attach_tokio_runtime`

dial9-core/src/recorder.rs
```diff
@@ -60,29 +86,94 @@ pub fn recorder<M: BufferMode>(writer: SegmentWriter<M>) -> RecorderBuilder<M> {
...
+}
+
+/// What the background worker runs over sealed segments.
+#[cfg(feature = "pipeline")]
+enum Pipeline {
+    /// dial9's own pipeline, assembled at build from the registered sources.
+    Default,
+    /// Exactly these processors, in this order.
+    Custom(Vec<Box<dyn crate::pipeline::SegmentProcessor>>),
+}
```
rcoh: do we use the `Default` variant of this for anything? it seems like its a `Vec<Box<...>>` either way?

rcoh (commented):
the only major issue I see is this:
```
The new session_handle: Dial9Handle field (line 30, populated at mod.rs:248 via RuntimeContext::new(runtime_name, handle.clone())) closes a strong-reference loop:

  Arc<SharedState>
    → sources: Mutex<Vec<Box<dyn Source>>>
      → TokioRuntimesSource
        → contexts: Arc<Mutex<Vec<Arc<RuntimeContext>>>>
          → RuntimeContext.session_handle: Dial9Handle
            → HandleInner.shared: Arc<SharedState>   // same SharedState
```

dial9-core/src/recorder.rs
jv1i: yup ended up not being needed, removed now

jv1i:
Thanks for the catch @rcoh on the cyclic references, it made me realize I had an even worse one with RuntimeContext and RuntimeMetrics, refactored that now and made sure teardown drops the sources (and added a test for it)

Renamed to `attach_tokio_runtime`.

Also added `Dial9Handle::track_current_thread()` to register threads for sources that need them like sched events.

rcoh (approved):

PR #702, opened 2026-07-23
feat(s3): Enable async S3 client construction

rcoh:
This adds a new method to `SegmentProcessor`, `initialize`:
```rust
    /// Initialize this processor on the worker's Tokio runtime before the
    /// pipeline begins processing segments.
    ///
    /// Drivers should call this once before [`process`](Self::process) and abort
    /// pipeline construction if it returns an error. Default: no-op.
    ///
    /// The same panic-safety contract as [`process`](Self::process) applies.
    fn initialize(&mut self) -> Pin<Box<dyn Future<Output = io::Result<()>> + Send + '_>> {
        Box::pin(std::future::ready(Ok(())))
    }
```

which we use to support:

```rust
    /// Variant of [`s3`](Self::s3) that constructs the S3 client asynchronously
    /// on the pipeline worker's Tokio runtime.
    ///
    /// The future is polled when the pipeline worker starts.
    #[cfg(feature = "worker-s3")]
    pub fn s3_with_client_future<F>(mut self, config: s3::S3Config, client_future: F) -> Self
    where
        F: std::future::Future<Output = aws_sdk_s3::Client> + Send + 'static,
    {
        self.processors.push(Box::new(
            S3PipelineUploader::new(config, None).with_client_future(client_future),
        ));
        self
    }
```

This enables constructing an S3 client in an async context:
```rust
let s3_config = S3Config::builder()
    .bucket("my-trace-bucket")
    .service_name("my-service")
    .build();
let custom_credentials_provider: aws_sdk_s3::config::Credentials = todo!();
let custom_endpoint = "https://s3.example.com";
let (recorder, runtime) = dial9::recorder(writer)
    .with_s3_uploader_client_future(s3_config, async move {
        let sdk_config = aws_config::defaults(aws_config::BehaviorVersion::latest())
            .credentials_provider(custom_credentials_provider)
            .endpoint_url(custom_endpoint)
            .load()
            .await;
        aws_sdk_s3::Client::new(&sdk_config)
    })
    .build()
```


dial9-tokio-telemetry/src/background_task/mod.rs
```diff
@@ -106,12 +106,29 @@ impl<Mode: BufferMode> PipelineBuilder<Mode> {
...
     }
+
+    /// Variant of [`s3`](Self::s3) that constructs the S3 client asynchronously
+    /// on the pipeline worker's Tokio runtime.
+    ///
+    /// The future is polled when the pipeline worker starts.
+    #[cfg(feature = "worker-s3")]
+    pub fn s3_with_client_future<F>(mut self, config: s3::S3Config, client_future: F) -> Self
+    where
+        F: std::future::Future<Output = aws_sdk_s3::Client> + Send + 'static,
```
Fluzko, line 125: So this is returning a `Client`, but not a `Result<Client>`... I think users have no other option than `panic!` if something goes wrong configuring/boostraping it, right?

Fluzko (approved):
Minor API definition comment, LGTM

rcoh:
yeah I considered that issue but I think in practice you can't fail there anyway. We catch the panic and produce an inert client

Issue #498, opened 2026-06-06
add tower layer / future wrapper


rcoh:
The tracing layer works well when your code is already instrumented with tracing but integrating it requires some care (and it requires wiring up a tracing subscriber which might not match how your application works).

We should add utility wrappers that emit span-like metadata but that can be integrated ad hoc (e.g. as a tower middleware, as a future wrapper, potentially even as a proc macro like instrument(??) )

This would make it much more turnkey to get application info into a dial9 trace.

Ideally we would emit a standardized format that can be automatically rendered as spans in the UI

PR #713, opened 2026-07-24
feat: add ad-hoc span wrappers: sync guard, future wrapper, tower layer

fmzbl:
addresses: #498 

adds point-of-use wrappers that get application spans into a dial9 trace without wiring up a tracing subscriber.

three shapes cover the common cases: a sync guard for blocking sections, a future wrapper for async work, and a tower layer (feature-gated) for per-request spans. they emit the same wire format the tracing layer already produces, so nothing changes in the viewer.

design in docs/design/ad-hoc-span-wrappers.md; example in dial9/examples/adhoc_spans.rs.


rcoh:
- needs to be in dial9-util
- this should be a 0-cost abstraction vs. hand constructing TraceEvents manually. As written it appears to have a global(?) schema mutex.

fmzbl:
moving it to dial9-utils would create a cyclic dependency to get the worker_id and would need much more code restructure, what should i do?

docs/design/ad-hoc-span-wrappers.md
```diff
@@ -0,0 +1,244 @@
...
+pub struct Instrumented<F>;                    // Future, transparent output
+```
+
+Tower layer (feature `tower`):
+
+```rust
+// Fixed name:
+Dial9SpanLayer::named("http_request")
+// Per-request span (attach fields via the macro):
+Dial9SpanLayer::new(|| dial9_span!("rpc", service = "auth"))
```
rcoh: is it possible to have the request be an input to this closure?

docs/design/ad-hoc-span-wrappers.md
```diff
@@ -0,0 +1,244 @@
...
+**Field names are macro identifiers; values are `Display`/`Debug`.** Names
+become wire-schema field names, so they are fixed per callsite by construction —
+exactly the low-cardinality property the schema story needs. Values are
+formatted to strings eagerly at construction. By default they are **interned**
+(pooled) on the wire, since a span re-emits its fields on every poll and
+low-cardinality values dedup well; a `~` prefix stores a value **inline**
+instead, for high-cardinality values (request ids) that would otherwise grow the
+string pool.
+
+**The sync guard emits one enter/exit pair; the future combinator emits one
```
rcoh: we actually don't need to emit one per poll. It would be helpful to instead emit _metadata_ on exit though. 

We can already reconstruct the polls from the existing telemetry by correlating on task.

In the general analytics case, what we are most curious about is the time distribution. So ideally for the future-instrument we could record:
- event on enter
- event when the future completes (& the event we emit when the future completes actually includes the idle time, active time, and the number of polls). The will simplify the analysis story as well in the future.

docs/design/ad-hoc-span-wrappers.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Non-goals
+
+- Replacing `Dial9TracingLayer`. Applications already on `tracing` should keep
+  using it; the two coexist in one process (see *Span ID space*).
+- Context propagation / implicit parenting. Contextual parents are unreliable
+  across tasks (see the comment in `tracing_layer.rs::on_enter`); the viewer
+  infers nesting from timestamp containment. We only support *explicit*
+  parenting, same as the tracing layer.
+- Late / deferred fields. Fields are captured at the call site when the span is
```
rcoh: I think a design that supported late fields would be potentially very powerful (and honestly critical to be useful).

My recommendation is we make it possible to declare fields as "late" then allow values to be set in them (basically using a oncelock)

something like:
```rust
let (span, slots) = dial9_span!(request_id = ~request_id, status_code = !!, client_id = !!);
async move { 
  slots.status_code.set(response.status_code);
  slots.client_id.set(request.client_id);
}
.instrument(s.span)


// possible impl:
macro_rules! dial9_span {
    // no slots: just hand back the span
    (request_id = ~$rid:expr $(,)?) => {
        tracing::info_span!("dial9", request_id = %$rid)
    };
    // one or more slots
    (request_id = ~$rid:expr $(, $slot:ident = !!)+ $(,)?) => {{
        struct __Slots { $($slot: Slot),* }
        let span = tracing::info_span!("dial9", request_id = %$rid, $($slot = tracing::field::Empty),*);
        let slots = __Slots { $( $slot: Slot::new(span.clone(), stringify!($slot)) ),* };
        (span, slots)
    }};
}
```

docs/design/ad-hoc-span-wrappers.md
```diff
@@ -0,0 +1,244 @@
...
+  `tower-service` deps). Mirrored as a passthrough `tower` feature on the `dial9`
+  umbrella crate.
+- `#[traceevent(name = <expr>)]`: additive attribute on the existing derive; no
+  effect on types that don't use it.
+- No trace-format changes: only new instances of the existing self-describing
+  span schemas. Old JS viewers render new traces unchanged.
+
+## Open questions for review
+
+1. Field value encoding: values are stringified (interned or inline) today,
```
rcoh: yes this would be nice. Is it possible we can use the existing TraceValue trait here? I haven;t looked

docs/design/ad-hoc-span-wrappers.md
```diff
@@ -0,0 +1,244 @@
...
+   matching the tracing layer. A typed `Value` (varint for ints, etc.) would
+   help future numeric aggregation in the viewer and remains a compatible schema
+   addition later.
+2. Tower `make_span` takes no arguments today (`Fn() -> Dial9Span`), so fields
+   are constants or captured from the environment, not read off the request. A
+   `Fn(&Req) -> Dial9Span` form would let callers pull request fields in, at the
```
rcoh: this is important

docs/design/ad-hoc-span-wrappers.md
```diff
@@ -0,0 +1,244 @@
...
+   cost of tying the closure to the request type. Deferred until there is a
+   concrete need.
+3. Module name: `span` (shipped) vs `spans` vs re-exports at crate root.
+4. Does the tower layer belong in this crate behind a feature (shipped), or in a
```
rcoh: a tower module in dial9-utils

dial9/examples/adhoc_spans.rs
```diff
@@ -0,0 +1,155 @@
...
+        let span = dial9_span!("pricing.compute_tax", order_id = order_id);
+        let _entered = span.enter();
+        compute_tax(&order)
+    }; // exit + close here
+
+    // 3. Explicit parenting across a spawn. The viewer nests spans by timestamp
+    //    containment, which breaks when a task outlives its parent, so link the
+    //    audit span explicitly. `id()` is the handle that crosses the boundary.
+    let charge_span = dial9_span!("payment.charge", order_id = order_id, tax_cents = tax);
+    let audit = Dial9Span::new("audit.emit").with_parent_id(charge_span.id());
```
rcoh: would this also work with the macro?

dial9/examples/adhoc_spans.rs
```diff
@@ -0,0 +1,155 @@
...
+        fn call(&mut self, order_id: u64) -> Self::Future {
+            Box::pin(async move {
+                checkout(order_id).await;
+                Ok(200)
+            })
+        }
+    }
+
+    // `make_span` runs per request, so it can name the span and attach fields.
+    let layer = Dial9SpanLayer::new(|| dial9_span!("http_request", route = "/checkout"));
```
rcoh: this wouldn't work in practice since you couldn't hardcode the route. might be worth either making this example more realistic or trying to use it in the metrics service

rcoh:
> moving it to dial9-utils would create a cyclic dependency to get the worker_id and would need much more code restructure, what should i do?

What do you mean about "much more code restructure"?

Worker id is nice-to-have but we can simply call that same code from dial9-util, it is just calling [worker_index](https://docs.rs/tokio/latest/tokio/runtime/fn.worker_index.html) and we can call the same thing from dial9-util, there's no need for a tokio-telemetry dependency.

We definitely want spans to work, even if you are not using tokio

dial9/examples/adhoc_spans.rs
fmzbl: yes

fmzbl:
done, moved it to dial9-utils

dial9-utils/src/tracing_layer.rs
```diff
@@ -59,8 +59,19 @@
...
 //! appropriate span filtering.
 
-use crate::telemetry::{Dial9Handle, clock_monotonic_ns, current_worker_id};
 use dial9_trace_format::TraceEvent;
+
+/// Emitted once when a span closes, so the viewer can recycle its id. Mirrors
+/// the `SpanCloseEvent` the ad-hoc span wrappers emit (same schema, distinct
+/// producer).
+#[derive(TraceEvent)]
+#[traceevent(wire_slot)]
```
rcoh, line 83: is there a reason we moved this into the top of the file inside the imports?

dial9-utils/examples/adhoc_spans.rs
```diff
@@ -0,0 +1,158 @@
...
+        }
+        .instrument(dial9_span!("settlement.run", batch = batch))
+        .await;
+        tokio::time::sleep(Duration::from_millis(5)).await;
+    }
+}
+
+#[cfg(feature = "tower")]
+async fn tower_demo() {
+    use dial9_utils::span::Dial9SpanLayer;
```
rcoh: I would put this in `dial9_utils::tower`

dial9-utils/Cargo.toml
```diff
@@ -67,11 +70,21 @@ axum-08 = [
...
+## combinator, and the sync guard). Independent of any async runtime.
+span = ["dep:dial9-trace-format"]
+## Tower middleware for ad-hoc spans (`span::Dial9SpanLayer`). Implies `span`.
+tower = ["span", "dep:tower-service", "dep:tower-layer"]
 
 [dev-dependencies]
 dial9-core = { workspace = true, features = ["pipeline", "test-util"] }
+# The span tests need a traced Tokio runtime; that lives in dial9-tokio-telemetry.
+# Dev-dependency only, so no dependency cycle.
+dial9-tokio-telemetry = { path = "../dial9-tokio-telemetry" }
```
rcoh: I think this may need a version number or `workspace = true`

dial9-utils/Cargo.toml
```diff
@@ -13,8 +13,10 @@ missing_debug_implementations = "warn"
 [dependencies]
 dial9-core = { workspace = true, features = ["pipeline"] }
 dial9-trace-format = { workspace = true, optional = true }
+pin-project-lite = "0.2"
 bon = "3"
-tokio = { version = "1", features = ["fs"], default-features = false }
+# `rt` is needed for `tokio::runtime::worker_index()` (the span `worker_id`).
+tokio = { version = "1", features = ["fs", "rt"], default-features = false }
```
rcoh: this should be `optional = true`

dial9-utils/Cargo.toml
```diff
@@ -67,11 +70,21 @@ axum-08 = [
...
+## Ad-hoc span wrappers (`span` module: the `dial9_span!` macro, the future
+## combinator, and the sync guard). Independent of any async runtime.
+span = ["dep:dial9-trace-format"]
+## Tower middleware for ad-hoc spans (`span::Dial9SpanLayer`). Implies `span`.
+tower = ["span", "dep:tower-service", "dep:tower-layer"]
 
 [dev-dependencies]
 dial9-core = { workspace = true, features = ["pipeline", "test-util"] }
+# The span tests need a traced Tokio runtime; that lives in dial9-tokio-telemetry.
+# Dev-dependency only, so no dependency cycle.
```
rcoh: this comment doesn't make sense; tokio-telemetry doesn't depend on dial9-util so there is no cycle to begin with

docs/design/ad-hoc-span-wrappers-example.rs
```diff
@@ -0,0 +1,165 @@
...
+fn main() {
+    let writer = DiskBuffer::single_file("/tmp/checkout-traces/trace.bin").unwrap();
+    let recorder = recorder(writer).build();
+    let (recorder, runtime) = recorder
+        .attach_tokio_runtime(|t| {
+            t.enable_all();
+        })
+        .unwrap();
+
+    runtime.block_on(serve());
```
rcoh: this should demonstrate graceful shutdown

docs/design/ad-hoc-span-wrappers-example.rs
```diff
@@ -0,0 +1,165 @@
...
+// the spans themselves live in `dial9-utils`.
+use dial9_tokio_telemetry::telemetry::{DiskBuffer, RecorderTokioExt, recorder};
+
+// The entire span surface. The macro + core wrappers are unconditional;
+// `Dial9SpanLayer` is behind the `tower` cargo feature.
+use dial9_utils::dial9_span;
+use dial9_utils::span::{Dial9Span, Dial9SpanLayer, Instrument as _, Span as _};
+
+// Note what is absent here: no tracing_subscriber::registry(), no
+// Dial9TracingLayer, no `tracing` dependency at all.
```
rcoh: not a useful comment

docs/design/ad-hoc-span-wrappers-example.rs
```diff
@@ -0,0 +1,165 @@
...
+//! subscriber wired up — the whole point of #498. Every span below reaches
+//! the dial9 trace directly and renders in the viewer's span lanes with zero
+//! extra wiring.
+
+use std::time::Duration;
+
+use axum::{Router, extract::Path, routing::post};
+// Runtime setup (buffer/recorder/attach) comes from the tokio telemetry crate;
+// the spans themselves live in `dial9-utils`.
+use dial9_tokio_telemetry::telemetry::{DiskBuffer, RecorderTokioExt, recorder};
```
rcoh: we should use these from `dial9`, not from `dial9_tokio_telemetry`

dial9-utils/examples/adhoc_spans.rs
```diff
@@ -0,0 +1,158 @@
...
+//! Note what is absent: no `tracing_subscriber::registry()`, no
+//! `Dial9TracingLayer`. Each span emits its events directly, and they render in
+//! the viewer's span lanes exactly like tracing-layer spans.
+//!
+//! Run with the `tower` feature to include the middleware section:
+//!
+//! ```sh
+//! cargo run -p dial9-utils --example adhoc_spans --features tower
+//! ```
+use dial9_tokio_telemetry::telemetry::{DiskBuffer, RecorderTokioExt, recorder};
```
rcoh: this can't use `dial9` because we are in this crate? I think I'd rather have the examples _in_ dial9 or use a dial9 circular dev dependency. I vaguely remember that not working though.

dial9-utils/src/span/mod.rs
```diff
@@ -0,0 +1,780 @@
...
+    timestamp_ns: u64,
+    worker_id: u64,
+    span_id: u64,
+    span_name: InternedString,
+    #[traceevent(unit = "ns")]
+    active_ns: u64,
+    #[traceevent(unit = "ns")]
+    idle_ns: u64,
+    poll_count: u64,
+    completed: bool,
```
rcoh, line 408: so this captures whether it was a clean exit or not?

dial9-trace-format-derive/src/lib.rs
```diff
@@ -23,20 +22,38 @@ fn derive_trace_event_impl(input: DeriveInput) -> Result<proc_macro2::TokenStrea
...
+    //   `concat!("SpanEnter:", file!(), ":", line!())`. Used to give generated
+    //   structs a name the viewer recognizes (e.g. `"SpanEnter:..."`).
     let mut wire_slot = false;
+    let mut name_override: Option<syn::Expr> = None;
     for attr in &input.attrs {
         if attr.path().is_ident("traceevent") {
             let _ = attr.parse_nested_meta(|meta| {
                 if meta.path.is_ident("wire_slot") {
                     wire_slot = true;
+                } else if meta.path.is_ident("name") {
```
rcoh, line 123: this error is ignored on line 37; (let _ = ...) so a malformed name fails silently

dial9-trace-format-derive/src/lib.rs
```diff
@@ -170,9 +187,18 @@ fn derive_trace_event_impl(input: DeriveInput) -> Result<proc_macro2::TokenStrea
...
+    let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
+
     Ok(quote! {
-        impl ::dial9_trace_format::TraceEvent for #name {
-            fn event_name() -> &'static str { #name_str }
+        // The generic params may be named after user fields (e.g. the ad-hoc
+        // span macro uses the field ident as its type param), so silence the
+        // casing lints on the generated impl.
+        #[allow(non_camel_case_types, non_snake_case)]
+        impl #impl_generics ::dial9_trace_format::TraceEvent for #name #ty_generics #where_clause {
```
rcoh, line 338: this is actually a (good) but decently complex change. Lets split out a separate PR so we can carefully review and test the implications of this. 

For example, allowing generics means that fields can have different types;

```rust
struct MyTraceEvent<T>
  field: T
}
```

but this is broken because we generate a single schema name.

We probably want to restrict this to just livetimes initially? (and not full generics for this reason).

In either case, it will be much easier to digest this and test it properly as a separate PR

dial9-utils/examples/adhoc_spans.rs
```diff
@@ -0,0 +1,181 @@
...
+//! Note what is absent: no `tracing_subscriber::registry()`, no
+//! `Dial9TracingLayer`. Each span emits its events directly, and they render in
+//! the viewer's span lanes exactly like tracing-layer spans.
+//!
+//! Run with the `tower` feature to include the middleware section:
+//!
+//! ```sh
+//! cargo run -p dial9-utils --example adhoc_spans --features tower
+//! ```
+use dial9_core::buffer::DiskBuffer;
```
rcoh: Juli moved S3 into its own crate, so this can use `dial9` now

dial9-utils/examples/adhoc_spans.rs
```diff
@@ -0,0 +1,181 @@
+//! Ad-hoc span wrappers: spans in a dial9 trace with no `tracing` subscriber.
+//!
+//! Note what is absent: no `tracing_subscriber::registry()`, no
```
rcoh: this can be written more clearly, "dial9 spans record spans directly to the trace file without using `tracing` as a bridge`.

rcoh (commented):
API is great, impl looks good to.

The macro changes are non trivial. I'd like to handle them in their own PR which we'll land first (so can really give them a careful consideration and review) then we will land the span PR.

dial9-utils/src/span/mod.rs
```diff
@@ -0,0 +1,780 @@
...
+#[derive(TraceEvent)]
+#[traceevent(name = "SpanExit:adhoc::runtime")]
+struct RuntimeExit {
+    #[traceevent(timestamp)]
+    timestamp_ns: u64,
+    worker_id: u64,
+    span_id: u64,
+    span_name: InternedString,
+    #[traceevent(unit = "ns")]
+    active_ns: u64,
```
rcoh, line 401: we need to update the viewer to handle these fields. you can file an issue to do it as a follow up. We should use the `dial9.role` metadata extensions to handle it

dial9-utils/src/span/mod.rs
```diff
@@ -0,0 +1,780 @@
...
+}
+
+#[derive(TraceEvent)]
+#[traceevent(name = "SpanExit:adhoc::runtime")]
+struct RuntimeExit {
+    #[traceevent(timestamp)]
+    timestamp_ns: u64,
+    worker_id: u64,
+    span_id: u64,
+    span_name: InternedString,
```
rcoh, line 399: this should use the new `dial9.role = span name` on this field

dial9-utils/src/span/mod.rs
```diff
@@ -0,0 +1,780 @@
...
+    parent_span_id: Option<u64>,
+    span_name: InternedString,
+}
+
+#[derive(TraceEvent)]
+#[traceevent(name = "SpanExit:adhoc::runtime")]
+struct RuntimeExit {
+    #[traceevent(timestamp)]
+    timestamp_ns: u64,
+    worker_id: u64,
```
rcoh: was there a reason for worker_id vs. task_id here? At a minimum this needs to be optional.

dial9-utils/src/span/mod.rs
```diff
@@ -0,0 +1,780 @@
...
+    span_id: u64,
```
rcoh: I would prefer a newtype for spanid to make sure we don't mix it up with other u64s if possible

dial9-utils/src/span/mod.rs
```diff
@@ -0,0 +1,780 @@
...
+
+#[derive(TraceEvent)]
+#[traceevent(name = "SpanEnter:adhoc::runtime")]
+struct RuntimeEnter {
+    #[traceevent(timestamp)]
+    timestamp_ns: u64,
+    worker_id: u64,
+    span_id: u64,
+    parent_span_id: Option<u64>,
+    span_name: InternedString,
```
rcoh, line 385: I think we should probably use the macro to also capture a `Location` of the span so its easily traceble to where it came from in the code

rcoh (commented):
a few more thoughts

dial9-utils/examples/adhoc_spans.rs
fmzbl: added it as a dev dep

dial9-utils/src/span/mod.rs
fmzbl: yes. true = ran to completion, false = dropped before finishing,

dial9-utils/src/span/mod.rs
fmzbl: done #759

dial9-utils/src/tracing_layer.rs
fmzbl: no, reason, accident, movining it back

dial9-utils/src/span/mod.rs
fmzbl: it comes from tokio::runtime::worker_index() with no runtime integration required, added it optional

dial9-utils/src/span/mod.rs
```diff
@@ -0,0 +1,809 @@
...
+        // type params live in separate namespaces), so the field keeps its
+        // concrete inferred type — a `u64` stays a `u64`/`Varint`, not a string.
+        // The wire schema name is per call site (`SpanEnter:<file>:<line>:<col>`),
+        // which the viewer groups on by its `SpanEnter:` prefix.
+        #[derive($crate::span::__rt::TraceEvent)]
+        #[traceevent(name = ::core::concat!(
+            "SpanEnter:", ::core::file!(), ":", ::core::line!(), ":", ::core::column!()
+        ))]
+        #[allow(non_camel_case_types, non_snake_case)]
+        struct __Dial9Enter<$($key: $crate::span::__rt::TraceField + ::core::clone::Clone + 'static),*> {
```
rcoh: so this is where we use the generic? Why did we use a trait bound? I think we _know_ what the type of the field is in the macro.

fmzbl:
i made it so we dont do generics that way anymore

dial9-utils/src/lib.rs
```diff
@@ -9,6 +11,24 @@
...
 pub mod dial9_axum;
 
+/// Ad-hoc span instrumentation (sync guard, future combinator, tower layer)
+/// that emits span events directly into a dial9 trace with no `tracing`
+/// subscriber. Construct spans with the [`dial9_span!`] macro.
+///
+/// Behind the `span` cargo feature so consumers that don't need spans don't
+/// compile `dial9-trace-format`/`pin-project-lite`.
+#[cfg(feature = "span")]
+#[cfg_attr(docsrs, doc(cfg(feature = "span")))]
```
rcoh: `cfg_attr` isn't needed anymore because doc auto_cfg is stabilized

dial9-utils/src/span/mod.rs
```diff
@@ -0,0 +1,814 @@
...
+/// Construct one with the [`dial9_span!`](crate::dial9_span) macro (which
+/// captures the call site and typed fields for you), or with [`Dial9Span::new`]
+/// for a name-only span whose name is chosen at runtime.
+///
+/// Timing is recorded when you attach the span to a future with
+/// [`Instrument::instrument`], or open it over a synchronous scope with
+/// [`enter`](Self::enter). A `SpanCloseEvent` is recorded when the span is
+/// dropped, telling the viewer the span is complete — so span types are
+/// intentionally **not** `Clone`: one identity, one close.
+pub trait Span: Sized {
```
rcoh: this trait should really never be implemented manually by user code right? We should clarify that and possibly give it a more precise name.

dial9-utils/src/span/mod.rs
```diff
@@ -0,0 +1,814 @@
...
+            .finish()
+    }
+}
+
+// ── The `dial9_span!` macro ───────────────────────────────────────────────────
+
+/// Construct a span, capturing the call site and typed fields.
+///
+/// This is the ergonomic, zero-cost way to create a span: it generates a
+/// dedicated (non-generic) [`TraceEvent`](dial9_trace_format::TraceEvent) type
```
rcoh: `(non-generic)` isn't relevant

dial9-utils/src/span/mod.rs
```diff
@@ -0,0 +1,814 @@
...
+/// per call site, so emitting is as cheap as a hand-written event and numeric
+/// fields stay numeric on the wire.
+///
+/// ```
+/// use dial9_utils::dial9_span;
+///
+/// // Just a name:
+/// let span = dial9_span!("load_config");
+///
+/// // Typed fields — `key: Type = value` keeps the type (Varint), `%` is
```
rcoh: `(Varint)`: suggests that all untyped fields are Varint which I assume isn't the case? presumably I could use bools and they would work?

dial9-utils/src/span/mod.rs
```diff
@@ -0,0 +1,814 @@
...
+/// let span = dial9_span!("validate", config = ?cfg);
+/// ```
+///
+/// An eager field is written `name: Type = value` and keeps its Rust type on
+/// the wire, so the type must implement
+/// [`TraceField`](dial9_trace_format::TraceField) (`u64`, `i64`, `bool`, `f64`,
+/// `String`, …). The type is required — it makes the generated event struct
+/// concrete rather than generic. Use `%`/`?` to render any `Display`/`Debug`
+/// value to an owned `String` without naming a type. The span name must be a
+/// `&'static str` expression (for a runtime name, use
```
rcoh: is there a technical reason we can't use an owned string here?

docs/design/ad-hoc-span-wrappers-example.rs
```diff
@@ -0,0 +1,176 @@
...
+    }; // exit + close here
+
+    // ── UX 4: explicit parenting across a spawn ──────────────────────────
+    // The viewer nests spans by timestamp containment, which is right for
+    // everything above. A spawned task can outlive this request, so
+    // containment breaks — link it explicitly instead. `span.id()` is the
+    // only handle that crosses the task boundary.
+    let charge_span = dial9_span!("payment.charge", order_id: u64 = order_id);
+    let audit = Dial9Span::new("audit.emit").with_parent_id(charge_span.id());
+    tokio::spawn(emit_audit_record(order_id).instrument(audit));
```
rcoh: this is a nice example. is everything in here covered in the _actual_ examples? Note that this is just an example in the design, we may want to remove it to avoid confusion

rcoh (approved):

PR #712, opened 2026-07-24
refactor: make viewer/cli feature off by default & update changelog

jv1i:
Drops `cli` from `dial9`'s default features, and updates readme and skill.md. 
Updates the changelog (changlog changes were in #701 but it didn't got merged).

PR #717, opened 2026-07-24
fix(release): build the dial9 bin with its feature, fix Windows paths

jv1i:
rc1 published fine but failed at buidling release binaries. None of this is visible on PRs: CI has no windows runner, and the release build command lives only in `release.yml`.

## Issues

- linux + macos: `-p dial9-viewer -p dial9 --features dial9/cli` doesn't satisfy `required-features` in a multi-package build, so cargo silently skipped the `dial9` bin. Regression from #712 (`cli` now off by default).
- **windows:** `build.rs` interpolates skill paths raw into `include_str!` literals, so backslashes land as invalid escapes. likely broken back in #515.
- windows, silent: the same backslashes break `TOOLKIT_FILES` (`starts_with("scripts/")`), so `agents toolkit` writes zero files, no error.
- binstall on windows: AFAIK never worked. `pkg-fmt = "tgz"` with no per-target override, but releases upload `.zip`. Verified against 0.3.13.

## Fixes

- Build `dial9` on its own with `--features cli`.
- Normalize to forward slashes in `build.rs`, at both sites.
- Add a binstall `zip` override for `x86_64-pc-windows-msvc`.

## Added

- Reusable `build-viewer-binaries.yml`. `release.yml` calls it, and it's dispatchable to rebuild any tag's assets without re-running a publish. It checks out the tag, not the release branch.
- `windows-cli-check` CI job, main-only and out of `ci-pass`.

## After merge

Dispatch it with `tag = dial9-viewer-v0.5.0-rc1` to backfill rc1's binaries.

Expect 4 green and windows red. The workflow comes from main, so the cargo fix applies, but it compiles the tag, which still has the old `build.rs`. Windows needs a release built after this merges.

### DIFF
--- a/.github/workflows/build-viewer-binaries.yml
+++ b/.github/workflows/build-viewer-binaries.yml
@@ -1,0 +1,115 @@
+name: Build viewer binaries
+
+# Builds the dial9-viewer and dial9 binaries for a tag and uploads them
+# to that tag's GitHub release.
+on:
+  workflow_call:
+    inputs:
+      tag:
+        description: "Release tag to build from and upload assets to"
+        required: true
+        type: string
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release tag to build for (e.g. dial9-viewer-v0.5.0-rc1)"
+        required: true
+        type: string
+
+permissions:
+  contents: write
+
+jobs:
+  build:
+    name: Build ${{ matrix.target }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+            archive: tar.gz
+          - target: aarch64-unknown-linux-gnu
+            os: ubuntu-24.04-arm
+            archive: tar.gz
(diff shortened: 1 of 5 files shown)


.github/workflows/build-viewer-binaries.yml
```diff
@@ -0,0 +1,115 @@
...
+# Builds the dial9-viewer and dial9 binaries for a tag and uploads them
+# to that tag's GitHub release.
+on:
+  workflow_call:
+    inputs:
+      tag:
+        description: "Release tag to build from and upload assets to"
+        required: true
+        type: string
+  workflow_dispatch:
```
rcoh, line 12: do we need workflow_dispatch?

rcoh (approved):

.github/workflows/build-viewer-binaries.yml
jv1i: Not needed for regular publishing, but for cases like this one where the crates published fine and only the binaries failed, we'd want a way to manually rebuild a tag's assets. We could just do another version bump instead, I thought it could come in handy for cases like this, can drop it later also (or now if we don't like it).

PR #718, opened 2026-07-24
Add durable deep links to the new trace viewer

rcoh:
## Summary
- make `/new/viewer.html` serialize and restore its complete durable analytical state
- add readable query params for selections, rails, layout, disclosures, analysis modes, embedded zoom, and parse-time ranges
- validate trace-dependent anchors after load while preserving legacy URL behavior
- document and contract-test the agent-constructible URL vocabulary
- refuse Copy Link for local/demo sources that cannot be reproduced from a URL

## State ownership contract
- settled analytical controls belong in the typed store (`ViewerViewSlice` or another durable slice), not component-local variables
- local component state is limited to transient interaction state, derivation caches, and widget lifecycle handles
- adding durable state requires its store type/default, URL projection and parser, README registry entry, and round-trip coverage
- contract tests pin the documented query table to `VIEWER_VIEW_QUERY_PARAMS`, while exhaustive store fixtures catch missing slice initialization

## Validation
- `npm run check:types`
- `npm test` — 1,860 tests passed; all legacy Node suites passed
- `npm run build`
- `cargo build -p dial9-viewer`
- post-rebase: 5 focused Vitest files / 44 tests passed


dial9-viewer/ui/src/lib/url/copy-link.ts
```diff
@@ -6,8 +6,8 @@
 // element, its inline look, and the copy mechanics.
 
 export interface CopyLinkOptions {
-  /** Run before reading the URL (pages pass the sync binding's flush). */
-  beforeCopy?: () => void;
+  /** Run before reading the URL; return false to cancel an unshareable copy. */
+  beforeCopy?: () => void | boolean;
```
Fluzko, line 10: why not narrow to:
```js
  beforeCopy?: () => void | false;
```
anyways the interface type it's a bit strange, mostly in a world where we have _falsy_ values, where both `void` & `false` are _falsy_ 

what about:

```js
 /** Run before reading the URL; return false to cancel an unshareable copy, otherwise true */
  beforeCopy?: () => boolean;
```

Fluzko (approved):

dial9-viewer/ui/src/lib/url/copy-link.ts
rcoh: Addressed in follow-up #737: supplied copy hooks now return an explicit boolean, and the duplicate status-bar contract plus flamegraph caller were updated as well.

PR #720, opened 2026-07-25
feat: seal active trace segments on demand

kev009:
## Summary

- integrate current-data checkpoints with recorder-installed dump triggers instead of adding a separate segment-rotation API
- make `DumpTrigger::dump_current_data()` synchronize in-flight writers, drain thread-local and flush-thread buffers, seal the active fragment, and dispatch that stable snapshot while recording continues in a fresh segment
- add `stop_recording_and_dump_current_data()` as the preferred bounded-capture endpoint: it takes one final source sample, closes the recording gate on the flush thread, drains in-flight events, and checkpoints the result
- keep the checkpoint control path bounded: the single-slot queue and single active checkpoint reservation return `WouldBlock` instead of allowing an unbounded backlog
- arm debounce only after a dump is accepted and, for recorder checkpoints, after snapshot reservation succeeds, so `WouldBlock` and disconnected dispatches do not suppress later retries
- protect current-data checkpoint segments from ring eviction until each segment reaches a terminal pipeline outcome, retaining protection across retryable disk and memory failures; disk mode restores the configured retention budget after success, terminal failure, panic, or a protection release after the flush thread has shut down, while memory mode bounds temporary overshoot to the protected snapshot plus the newly sealed checkpoint segment
- own the checkpoint reservation and protected segment set with one drop guard, so construction failures, failed dispatch, queued shutdown, and normal resolution all use the same cleanup path
- match recorder checkpoints by their exact protected segment snapshot rather than coarse wall-clock windows, preventing both boundary-crossing omissions and post-boundary inclusion even when a disk segment was previously cached as out of range
- preserve the existing memory-ring floor that keeps a lone oversized segment while sharing one eviction routine between sealing and checkpoint cleanup
- document that `dump_time_range()` remains best-effort and does not pin segments
- install the terminal `RetainProcessor` only for disk-backed on-trigger mode with no processing stages, so a checkpoint can be acknowledged without rewriting or deleting its raw segment; memory mode keeps its ring intact and reports `WorkerStopped` when no consumer is configured
- preserve `DiskBuffer::single_file` as a true no-rotation mode by rejecting non-empty current-data checkpoints with `Unsupported`
- preserve standalone `dump::channel()` behavior, keep the writer retryable after a strict pre-seal flush failure, and dispatch an already-sealed checkpoint even when opening its replacement writer fails
- count only events actually written through the enabled-gate encoder path, share its encode/flush body with the ordinary path, and pair producer epoch publication with an acquire read before the flush thread skips a busy buffer
- gate dump-trigger test fixtures on the `pipeline` feature so featureless `dial9-core` builds do not compile pipeline-only APIs
- harden checkpoint fixtures against expected concurrency: the S3 bounded capture emits a dedicated event in the final source sample, the retention-budget assertion tolerates a file disappearing during the cleanup pass it is observing, and the shared boundary source removes duplicate setup

If recording was disabled before `dump_current_data()`, sources are deliberately not sampled because a source flush could pull post-boundary state into the capture. Source-owned pre-disable data may remain buffered; callers ending a bounded capture should use `stop_recording_and_dump_current_data()`.

## Retention and concurrency

A recorder-installed current-data dump establishes its boundary on the flush thread, then holds an eviction reservation until the worker has claimed and processed every checkpoint segment. Only one protected checkpoint may be active. This prevents the checkpoint rotation itself from evicting the snapshot it was created to preserve.

Time-range dumps retain their existing semantics: lookback/lookforward coverage is best-effort, and matching segments can be evicted while the request is pending or processing.

## Validation

Current rebased PR branch:

- `git diff --check`
- `cargo check -p dial9-core --no-default-features`
- `cargo check -p dial9-core --all-features`
- `cargo clippy --all-targets --features __nonlinux_all_features -- -D warnings` (FreeBSD)
- `cargo test -p dial9-core --all-features` — 250 passed; doc tests passed
- `cargo nextest run` — 1,357 passed
- `cargo nextest run --stress-duration 20s` — one complete 1,357-test iteration passed
- `./scripts/test-shuttle.sh` — 25 passed, 1 intentionally ignored
- focused integration regression: disk no-pipeline current-data checkpoint retains its raw segment

### DIFF
--- a/dial9-core/src/buffer.rs
+++ b/dial9-core/src/buffer.rs
@@ -3,6 +3,8 @@ use dial9_trace_format::encoder::{Encoder, RawEncoder};
 use crate::clock::clock_pair;
 use crate::collector::Batch;
 use crate::format::{ClockSyncEvent, SegmentMetadataEvent};
+#[cfg(feature = "pipeline")]
+use crate::fs::CheckpointGuard;
 use crate::fs::{ActiveHandle, Fs, RemoveReason};
 use crate::primitives::fs;
 use crate::rate_limit::rate_limited;
@@ -842,15 +844,8 @@ impl<M: BufferMode> SegmentWriter<M> {
     /// recording gate, so a busy capture-stop request has no lifecycle side
     /// effects.
     #[cfg(feature = "pipeline")]
-    pub(crate) fn reserve_checkpoint(&self) -> std::io::Result<()> {
-        if self.fs.try_reserve_checkpoint() {
-            Ok(())
-        } else {
-            Err(std::io::Error::new(
-                std::io::ErrorKind::WouldBlock,
-                "another current-data checkpoint is still active",
-            ))
-        }
+    pub(crate) fn reserve_checkpoint(&self) -> std::io::Result<CheckpointGuard> {
+        CheckpointGuard::reserve(&self.fs)
     }
 
     /// Protect the current ring snapshot, then seal a non-empty active
@@ -859,7 +854,10 @@ impl<M: BufferMode> SegmentWriter<M> {
     /// the disk segment's pipeline pass; this prevents the checkpoint's own
     /// rotation from evicting data it is about to dispatch.
     #[cfg(feature = "pipeline")]
-    pub(crate) fn checkpoint_segments(&mut self) -> std::io::Result<Vec<u32>> {
+    pub(crate) fn seal_checkpoint(
+        &mut self,
+        mut checkpoint: CheckpointGuard,
+    ) -> std::io::Result<CheckpointGuard> {
         if matches!(self.state, WriterState::Finished) {
             return Err(std::io::Error::new(
                 std::io::ErrorKind::BrokenPipe,
@@ -882,14 +880,12 @@ impl<M: BufferMode> SegmentWriter<M> {
         } else {
             self.fs.protect_memory_checkpoint_segments(current_index)
         };
+        checkpoint.track(protected);
 
-        if self.has_real_events
-            && let Err(error) = self.rotate_strict()
-        {
-            self.fs.release_checkpoint_segments(&protected);
-            return Err(error);
+        if self.has_real_events {
+            self.rotate_strict()?;
         }
-        Ok(protected)
+        Ok(checkpoint)
     }
 
     /// Re-apply the disk budget after the worker finishes checkpoint segments
@@ -900,16 +896,6 @@ impl<M: BufferMode> SegmentWriter<M> {
             self.evict_oldest()?;
         }
         Ok(())
-    }
-
-    #[cfg(feature = "pipeline")]
-    pub(crate) fn release_checkpoint_segments(&self, indices: &[u32]) {
-        self.fs.release_checkpoint_segments(indices);
-    }
-
-    #[cfg(feature = "pipeline")]
-    pub(crate) fn finish_checkpoint(&self) {
-        self.fs.finish_checkpoint();
     }
 
     #[cfg(test)]
--- a/dial9-core/src/buffer.rs
+++ b/dial9-core/src/buffer.rs
@@ -771,8 +771,9 @@ impl<M: BufferMode> SegmentWriter<M> {
         }
         // If even the current file alone exceeds total budget, stop writing.
         // Protected checkpoint segments may temporarily occupy the pipeline's
-        // in-flight reserve; they become evictable as soon as the worker loads
-        // them and must not make the active writer look irrecoverably full.
+        // in-flight reserve; they remain protected across retries and become
+        // evictable only after a terminal pipeline outcome or dump cleanup.
+        // They must not make the active writer look irrecoverably full.
         if self.total_size() > self.max_total_size && self.closed_files.is_empty() {
             self.state = WriterState::Finished;
         }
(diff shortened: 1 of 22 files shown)


jv1i (commented):
Hi, before reviewing the implementation I want to understand the use case. is there a reason to seal on demand rather than going through `DumpTrigger`? if the intent is "capture everything up to now", could `dump_current_data()` cover the in-flight data itself, so callers don't need to know about segments? 

Also not sure we want a blocking method on Dial9Handle, since that's what people are holding inside async code.

For future PRs, if your need doesn't match an existing issue, would you mind filing one first? (or at least include a motivation for these changes) so that it's easier to determine if the proposed implementation is the best way to accomplish that.

kev009:
The consumer is an admin-controlled flight recorder with potentially large segments that take a while to rotate depending on load. At capture stop it needs everything through an exact stop boundary to become immediately downloadable, without waiting for periodic rotation.

I reworked this around DumpTrigger as you suggest: dump_current_data() now drains pending/in-flight events and seals the active fragment before dispatching the pipeline. Capture stop uses the async stop_recording_and_dump_current_data() operation, which closes the recording gate on the flush thread and then establishes the same checkpoint.

The public blocking Dial9Handle::rotate_segment API and SegmentRotation type are gone and segment rotation is now entirely an internal implementation detail.

kev009:
This grew large by LoC/churn unfortunately.. but a majority is tests/doc, the implementation is around 730LoC.

The change is larger than a segment rotation because an exact current-data dump crosses thread-local encoders, the flush thread, the worker, and both storage backends. The additional retention lifecycle is necessary to prevent the checkpoint’s own rotation from evicting the data being dumped, while the single reservation keeps backpressure and memory overshoot bounded.  The tests added caught and eventually fixed flakiness around these.

dial9-core/src/buffer.rs
```diff
@@ -605,21 +621,30 @@ impl<M: BufferMode> SegmentWriter<M> {
             return Ok(());
         };
 
-        // Best-effort flush. If the underlying file is gone the buffered bytes
-        // are already lost; proceed to rotate rather than erroring.
-        let _ = raw.flush();
+        if strict_flush {
+            raw.flush()?;
```
kev009: **[bug]** `rotate_with_flush_policy` replaces `state` with `Finished` before the strict flush / `BufWriter::into_inner`. On `raw.flush()?` or the strict `into_inner` error, the encoder and handle are dropped without restore or seal, so unflushed bytes are lost and `is_closed()` is true. The flush loop then `shared.disable()`s. That contradicts the nearby comment that a pre-rotation flush failure should leave the writer available for retry. `dump_current_data_returns_checkpoint_flush_failure` only injects a `flush_once` failure (`fail_after_flushes_for_test(0)`), and `explicit_rotation_propagates_final_flush_failure` never asserts liveness; the NotFound path is the one that recovers. Automatic rotation still ignores flush errors and seals; the checkpoint path is stricter about acknowledgement but more destructive.

**Suggestion:** On a strict flush/`into_inner` error, put `WriterState::Active { writer: raw, need_metadata }` back before returning the error so the checkpoint fails, recording stays enabled, and a later dump can retry. Keep `Finished` only after the encoder has been consumed by seal or a failure to open the replacement segment. Extend the existing rotation unit test to assert `!writer.is_closed()` after the injected flush failure.

dial9-core/src/encoder.rs
```diff
@@ -370,6 +370,74 @@ pub(crate) fn with_encoder(
...
+            register(TlBufferHandle {
+                buffer: Arc::downgrade(arc),
+                flush_epoch: buf.flush_epoch.clone(),
+            });
+        }
+        if !is_enabled() {
+            return;
+        }
+        f(&mut buf.thread_local_encoder());
+        buf.event_count += 1;
```
kev009: **[bug]** `with_encoder_if_enabled` always does `buf.event_count += 1` after `f()`. `ThreadLocalEncoder::encode` / `write_event` already increment the same counter, and the encoder exists specifically so a callback that declines to write (metrique dropping an entry) is not counted. Every production `Dial9Handle::record_event` / `with_encoder` path now goes through this helper, so real events are counted twice in `Batch::event_count`, flush metrics, and `TlDrainStats`. A no-op `with_encoder` marks the buffer as having pending events and can flush a header-only batch. Shared-state tests still use the old `record_encodable_event` shortcut, so they do not catch this.

**Suggestion:** Delete the extra `buf.event_count += 1` and match `with_encoder`. Cover `EventBuffer::record_encodable_event` / `with_encoder` with an event-count assertion (one encoded event => `batch.event_count() == 1`; declined `with_encoder` => no pending flush).

kev009:
@jv1i let me know if you think this is getting out of hand, I can abandon the approach if it is too much incidental complexity.

rcoh:
It makes sense why we're not going with the dump trigger machinery; given that, I think a force-flush API is a reasonable change to make. We would want the API to coordinate with the epoch increment loop so you can get a clean flush of all data. It should be fairly straightforward (without having to go through the dump, dump trigger, etc.) Did you even open issue to track this request from a UX side?

kev009:
@rcoh I created https://github.com/dial9-rs/dial9/issues/886.  I don't have any strong opinions on the UX or implementation.  Let me know if I should close this PR.

kev009:
Closing as I don't think I'm on the right track here, the created issue records the feature.

PR #721, opened 2026-07-25
feat: improve FreeBSD runtime telemetry

kev009:
## Summary

- use `pthread_getthreadid_np()` for real FreeBSD kernel thread IDs instead of process-local synthetic IDs
- keep the end-to-end `max_rss_bytes > 0` assertion reliable on FreeBSD by touching a bounded resident working set while waiting up to one second for a resource-accounting tick
- add focused conversion coverage for FreeBSD's `ru_maxrss` units and negative-value rejection
- add a non-blocking FreeBSD 15.1 CI job using `vmactions/freebsd-vm` pinned to `77ed28d336d03fe19a3f4f7266c1d2c4714dd79d` (`v1.5.2`)
- bound the VM job to 30 minutes and cache its prepared minimal Rust toolchain
- keep the VM workload focused and featureless: the FreeBSD TID test, Tokio park/unpark and process-resource integration tests, and process-resource self-profile tests with only their required features

The FreeBSD job uses `continue-on-error: true` and is not part of the aggregate `CI Pass` gate. It is intended to provide native-platform signal without making the VM action a required merge dependency.

Thread CPU time remains unsupported on FreeBSD. FreeBSD services `CLOCK_THREAD_CPUTIME_ID` with a syscall rather than its vDSO, and reading it on every Tokio worker park and unpark would add unsampled hot-path overhead. Supporting sampled CPU time requires explicit missing-sample semantics and is left for separate work.

## Validation

Native FreeBSD 16.0-CURRENT amd64, using the same focused commands as the workflow:

- `cargo test -p dial9-core freebsd_tid_matches_kernel_thread_id` — 1 passed
- `cargo test -p dial9-tokio-telemetry --test park_unpark_tid --test process_resource_usage` — 3 passed
- `cargo test -p dial9-perf-self-profile --features "process-resource test-util" process_resource` — 7 passed

Additional branch validation:

- `cargo clippy --all-targets --features __nonlinux_all_features -- -D warnings`
- `git diff --check`

The exact FreeBSD 15.1 and Rust 1.94.1 pairing runs in the pull request GitHub Actions job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,10 +90,9 @@ jobs:
             . "$HOME/.cargo/env"
             freebsd-version
             rustc --version
-            cargo test -p dial9-core --all-features \
+            cargo test -p dial9-core \
               freebsd_tid_matches_kernel_thread_id
             cargo test -p dial9-tokio-telemetry \
-              --features __nonlinux_all_features \
               --test park_unpark_tid \
               --test process_resource_usage
             cargo test -p dial9-perf-self-profile \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,8 @@ jobs:
   freebsd:
     name: FreeBSD 15.1
     runs-on: ubuntu-latest
-    timeout-minutes: 75
+    continue-on-error: true
+    timeout-minutes: 30
     env:
       RUST_BACKTRACE: 1
       RUSTFLAGS: "--cfg tokio_unstable -C force-frame-pointers=yes"
@@ -82,22 +83,22 @@ jobs:
           cache-after-prepare: true
           envs: "RUST_BACKTRACE RUSTFLAGS"
           prepare: |
-            pkg install -y node24 rustup-init
+            pkg install -y rustup-init
             rustup-init -y --profile minimal --default-toolchain "${{ env.STABLE_TOOLCHAIN }}"
             . "$HOME/.cargo/env"
-            rustup component add clippy
           run: |
             . "$HOME/.cargo/env"
             freebsd-version
             rustc --version
-            cargo test -p dial9-core --all-features
-            cargo test -p dial9-tokio-telemetry --features __nonlinux_all_features
+            cargo test -p dial9-core --all-features \
+              freebsd_tid_matches_kernel_thread_id
+            cargo test -p dial9-tokio-telemetry \
+              --features __nonlinux_all_features \
+              --test park_unpark_tid \
+              --test process_resource_usage
             cargo test -p dial9-perf-self-profile \
               --features "process-resource test-util" \
               process_resource
-            cargo clippy --all-targets \
-              --features __nonlinux_all_features \
-              -- -D warnings
 
   windows-cli-check:
     name: Windows CLI check
@@ -497,7 +498,7 @@ jobs:
   ci-pass:
     name: CI Pass
     runs-on: ubuntu-latest
-    needs: [fmt, clippy, build, freebsd, android-aarch64-build, feature-check, no-tokio-unstable, check-docs, trace-integrity, ui, asan, ecs-sim, shuttle, package, doctests]
+    needs: [fmt, clippy, build, android-aarch64-build, feature-check, no-tokio-unstable, check-docs, trace-integrity, ui, asan, ecs-sim, shuttle, package, doctests]
     if: always()
     steps:
       - run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,41 @@ jobs:
           toolchain: ${{ matrix.toolchain }}
           save-if: ${{ github.ref == 'refs/heads/main' }}
 
+  freebsd:
+    name: FreeBSD 15.1
+    runs-on: ubuntu-latest
+    timeout-minutes: 75
+    env:
+      RUST_BACKTRACE: 1
+      RUSTFLAGS: "--cfg tokio_unstable -C force-frame-pointers=yes"
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - name: Build and test in FreeBSD
+        uses: vmactions/freebsd-vm@77ed28d336d03fe19a3f4f7266c1d2c4714dd79d # v1.5.2
+        with:
+          release: "15.1"
+          usesh: true
+          copyback: false
+          cache-after-prepare: true
+          envs: "RUST_BACKTRACE RUSTFLAGS"
+          prepare: |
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -64,6 +64,41 @@ jobs:
           toolchain: ${{ matrix.toolchain }}
           save-if: ${{ github.ref == 'refs/heads/main' }}
 
+  freebsd:
+    name: FreeBSD 15.1
+    runs-on: ubuntu-latest
+    timeout-minutes: 75
```
jv1i: 75 min seems excessive, and this would be gating on every PR
could it go non-blocking (out of ci-pass) and/or main-only, like build-nightly?

.github/workflows/ci.yml
```diff
@@ -64,6 +64,41 @@ jobs:
...
+      - name: Build and test in FreeBSD
+        uses: vmactions/freebsd-vm@77ed28d336d03fe19a3f4f7266c1d2c4714dd79d # v1.5.2
+        with:
+          release: "15.1"
+          usesh: true
+          copyback: false
+          cache-after-prepare: true
+          envs: "RUST_BACKTRACE RUSTFLAGS"
+          prepare: |
+            pkg install -y node24 rustup-init
```
jv1i: is node24 used by something here? seems like it doesn't from a quick read

.github/workflows/ci.yml
```diff
@@ -64,6 +64,41 @@ jobs:
...
+            freebsd-version
+            rustc --version
+            cargo test -p dial9-core --all-features
+            cargo test -p dial9-tokio-telemetry --features __nonlinux_all_features
+            cargo test -p dial9-perf-self-profile \
+              --features "process-resource test-util" \
+              process_resource
+            cargo clippy --all-targets \
+              --features __nonlinux_all_features \
+              -- -D warnings
```
jv1i: I don't think we need to run clippy here since we already do all-targets --features in ubuntu, the one addition here is __nonlinux_all_features but I think we could add that elsewhere, and keep this freebsd specific.

perf-self-profile/src/process_resource.rs
```diff
@@ -273,9 +273,23 @@ mod unix {
         fn read_process_resource_usage_returns_metrics() {
             let snapshot = read_process_resource_usage()
                 .expect("getrusage should succeed for the current process");
+            // FreeBSD may report ru_maxrss == 0 until its first accounting
+            // tick, so validate a resource counter populated at startup.
+            #[cfg(target_os = "freebsd")]
+            assert!(snapshot.minor_faults > 0);
+            #[cfg(not(target_os = "freebsd"))]
             assert!(snapshot.max_rss_bytes > 0);
```
jv1i: this would drop max_rss_bytes assertion entirely for freebsd, could we do a little work (or retry briefly) and keep the real assertion instead of swapping the field permanently?

dial9-tokio-telemetry/tests/process_resource_usage.rs
```diff
@@ -43,6 +43,12 @@ fn traced_runtime_records_process_resource_usage() {
         "expected at least one process resource usage event"
     );
     assert!(metrics[0].timestamp_ns > 0);
+    // A newly started FreeBSD process may report ru_maxrss == 0 until the
+    // kernel's first resource-accounting tick, so validate a resource counter
+    // populated at startup instead.
+    #[cfg(target_os = "freebsd")]
+    assert!(metrics[0].minor_faults > 0);
+    #[cfg(not(target_os = "freebsd"))]
```
jv1i: same as the unit test, if we can wait out the first accounting tick, keeping max_rss_bytes asserted here would be better than switching to minor_faults

jv1i (commented):
Looks good! we might want to trim the CI job for this

.github/workflows/ci.yml
kev009: Done in ecc31fea. The FreeBSD job is now non-blocking (`continue-on-error: true`), removed from `ci-pass`, and reduced from 75 to 30 minutes. I kept it running on PRs so we retain pre-merge FreeBSD signal without gating merges.

perf-self-profile/src/process_resource.rs
kev009: Done in ecc31fea. The test keeps the real `max_rss_bytes > 0` assertion. On FreeBSD, sleeping alone did not advance `ru_maxrss`, so the bounded helper touches a resident working set while running through a likely statclock tick and retries for up to one second. No `minor_faults` substitution remains.

dial9-tokio-telemetry/tests/process_resource_usage.rs
kev009: Applied the same treatment in ecc31fea: the integration test primes FreeBSD RSS accounting with bounded page-touching work, then asserts the emitted event’s `max_rss_bytes > 0`. The assertion is no longer replaced with `minor_faults`.

.github/workflows/ci.yml
kev009: Good catch. The old broad telemetry command did indirectly reach `js_parser.rs`—the historical FreeBSD run failed there when Node was absent. In ecc31fea I narrowed the job to the FreeBSD-specific test binaries, so Node is no longer needed and has been removed.

.github/workflows/ci.yml
kev009: Done in ecc31fea. I removed FreeBSD clippy and the clippy component installation. The job now runs only the FreeBSD-specific TID and process-resource tests. Linux already runs all-targets/all-features clippy, and macOS already exercises `__nonlinux_all_features`.

.github/workflows/ci.yml
```diff
@@ -64,6 +64,42 @@ jobs:
...
+          prepare: |
+            pkg install -y rustup-init
+            rustup-init -y --profile minimal --default-toolchain "${{ env.STABLE_TOOLCHAIN }}"
+            . "$HOME/.cargo/env"
+          run: |
+            . "$HOME/.cargo/env"
+            freebsd-version
+            rustc --version
+            cargo test -p dial9-core --all-features \
+              freebsd_tid_matches_kernel_thread_id
```
jv1i, line 94: I think here we can either drop `--all-features` (I think I'd do that), or run the whole suite and drop the test name filter. As AFAICT this would build for all features but skip +200 tests

jv1i (commented):
I apologize, with "trim the CI job" I was thinking of removing clippy and nonlinux__all_features, not necessarily trimming the test contents to the specific tests (I should've been clearer).

But I think this is ok for now if we just drop those feature flags (since I think we don't need them?), and we can follow up with something like a main-only job that runs the whole suite on freebsd so we don't have to track individual tests.

jv1i (approved):

PR #722, opened 2026-07-25
feat(viewer)!: make S3 support optional

kev009:
## Summary

- gate `dial9-viewer`'s AWS SDK-backed S3 storage, bring-your-own
  credentials, and assume-role support behind a default-on `s3` feature
- support a local-only `dial9-viewer --no-default-features` build whose normal
  dependency graph contains no AWS SDK or S3 crates
- keep `dial9 --features cli` S3-capable by explicitly forwarding
  `dial9-viewer/s3`
- move S3-specific viewer implementation into feature-gated modules, with
  local-only counterparts for app assembly, credentials, and server routes
- reject S3-only configuration with explicit errors when S3 support is not
  compiled
- add CI coverage for the no-S3 viewer tests and dependency-graph boundaries

The uploader is intentionally out of scope. PR #742 moved it into
`dial9-destinations-s3`; this PR now changes only the viewer and its CLI
feature forwarding.

## Breaking change

Consumers using `dial9-viewer` with `default-features = false` must add
`features = ["s3"]` to retain the viewer's S3 APIs. Default `dial9-viewer`
builds and `dial9 --features cli` retain S3 support.

For a local-filesystem-only viewer without the AWS SDK, install or depend on
`dial9-viewer` with default features disabled.

## Validation

- `cargo test -p dial9-viewer --lib` — 361 passed
- `cargo test -p dial9-viewer --no-default-features --lib` — 331 passed
- local-only dependency guard found no `aws-*` or `s3s` packages; the
  S3-enabled `dial9` CLI graph includes `aws-sdk-s3`
- `cargo nextest run` — 1,190 passed
- `cargo nextest run --stress-duration 20s` — 1,190 passed, no flakes
- `cargo clippy --all-targets --features __nonlinux_all_features` — passed on
  FreeBSD
- `cargo fmt --check`
- default and no-default-feature `dial9-viewer` documentation builds
- `git diff --check`


dial9-tokio-telemetry/Cargo.toml
```diff
@@ -67,7 +67,7 @@ metrique = { version = "0.1.23", features = ["local-format"] }
 metrique-timesource = "0.1"
 shuttle = { version = "0.9.1", optional = true }
 dial9-core = { workspace = true, features = ["pipeline"] }
-dial9-utils = { workspace = true, optional = true }
```
jv1i: could dial9-utils be default = [] instead? then there's nothing to disable and this stays `{ workspace = true, optional = true }`. Since it'll be a grab-bag crate (s3 now, the axum bits in #696), we might not want a default single feature (otherwise #696's `dial9-utils = { features = ["axum-08"] }` would pull silently the whole AWS SDK)

if we do keep default = ["s3"], I think we can keep the workspace inheritance here by moving default-features = false up into [workspace.dependencies]?

side-note for myself: revisit whether .s3()/.symbolize() belong on PipelineBuilder or as per-crate ext traits like `RecorderPerfExt`

dial9-viewer/src/storage.rs
```diff
@@ -241,6 +241,7 @@ impl std::error::Error for StorageError {}
 ///
 /// Uses the structured error code (via `ProvideErrorMetadata`) rather than
 /// string matching, plus the HTTP status as a backstop.
+#[cfg(feature = "s3")]
```
jv1i: would moving the S3 items into storage/s3.rs and gating the module once work here? would be nice to avoid all the individual `cfg(feature = "s3")` in this file, same in server/mod.rs and lib.rs

dial9-viewer/src/server/config.rs
```diff
@@ -38,7 +38,16 @@ pub async fn get_config(State(state): State<AppState>) -> Json<ConfigResponse> {
...
-        supports_assume_role: state.role_assumer.is_some(),
+        supports_assume_role: {
+            #[cfg(feature = "s3")]
+            {
+                state.role_assumer.is_some()
+            }
+            #[cfg(not(feature = "s3"))]
+            {
+                false
+            }
```
jv1i: nit: I'd do what you did for `resolve_with_region` and have two cfg'd impls rather than inlining here

dial9-viewer/src/server/mod.rs
```diff
@@ -378,12 +395,12 @@ impl AppState {
...
     /// - malformed/incomplete/conflicting headers → 400
-    /// - [`CredSource::Static`] (and BYO enabled) → ephemeral S3 backend signed
+    /// - `CredSource::Static` (and BYO enabled) → ephemeral S3 backend signed
     ///   with the user's keys directly
-    /// - [`CredSource::AssumeRole`] (and an assumer wired) → assume the role with
+    /// - `CredSource::AssumeRole` (and an assumer wired) → assume the role with
     ///   the server's own identity via STS, then build the ephemeral backend from
     ///   the minted credentials
-    /// - [`CredSource::Default`] → the server's default backend
+    /// - `CredSource::Default` → the server's default backend
```
jv1i: could we somehow not regress on these doc links? at least for the default build

.github/workflows/ci.yml
```diff
@@ -171,6 +171,16 @@ jobs:
...
+      - name: Test viewer without S3
+        run: cargo test -p dial9-viewer --no-default-features --lib
+      - name: Check local-only CLI excludes AWS dependencies
+        run: |
+          cargo tree -p dial9 --no-default-features --features cli \
+            --edges normal --prefix none --format '{p}' > /tmp/dial9-cli-tree.txt
+          if grep -E '^(aws-|s3s)' /tmp/dial9-cli-tree.txt | grep -vE '^aws-lc-'; then
+            echo "AWS SDK/S3 dependencies found in the local-only dial9 CLI graph"
+            exit 1
+          fi
```
jv1i: any reason to exclude aws-lc-? AFAICT it only ever arrives via aws-smithy-http-client, so once the SDK is gone it should be gone too right?

dial9/Cargo.toml
```diff
@@ -49,7 +49,12 @@ bon = { version = "3", optional = true }
 tracing = { version = "0.1.44", optional = true }
 
 [features]
-default = []
+# Weak forwarding preserves the existing S3-enabled CLI when default features
+# are active without pulling the optional viewer into library-only builds.
+default = ["viewer-s3"]
+
+## S3 support for the optional viewer/analysis CLI.
+viewer-s3 = ["dial9-viewer?/s3"]
```
jv1i: I don't know if I love having viewer-s3 on by default when it's inert unless the viewer is on. 

we could consider either making s3 opt-in in the viewer, or split cli/cli-local here instead of forwarding through a default? cc @rcoh

jv1i (commented):
Thanks for the submission! Overall it looks good and I think it's a good idea, I'm on the fence regarding feature defaults and how it's wired, leaving some comments.

dial9/Cargo.toml
rcoh: we shouldn't default to including the viewer. _maybe_ the best way to deal with it is "if you want the viewer without s3, use the dial-viewer crate directly?"

dial9-tokio-telemetry/Cargo.toml
kev009: Done. `dial9-utils` now has `default = []`, and `dial9-tokio-telemetry` is back to `dial9-utils = { workspace = true, optional = true }`. The `worker-s3` feature still explicitly enables `dial9-utils/s3`. I also enabled all features for the crate’s docs.rs build so its published API docs do not become empty.

dial9-viewer/src/storage.rs
kev009: Done. The storage implementation moved to `storage/s3.rs` behind one module-level feature gate. The server’s S3-specific `AppState` implementation and routes similarly moved to `server/s3.rs`, with a small local-only counterpart. The S3 assembly used by `build_app` is also behind gated modules, leaving the function body free of inline cfg branches.

dial9-viewer/src/server/config.rs
kev009: Done. `config.rs` now calls `state.supports_assume_role()`. The S3 and local-only implementations live in their respective cfg-gated `AppState` modules; the local-only implementation always returns `false`.

dial9-viewer/src/server/mod.rs
kev009: Restored. The S3 implementation now uses real links to `CredSource::Static`, `CredSource::AssumeRole`, and `CredSource::Default`. The local-only implementation has separate documentation describing its actual behavior. Both default and no-default documentation builds were checked.

.github/workflows/ci.yml
kev009: Agreed—the exclusion was unnecessary. The guard now rejects the bare `^(aws-|s3s)` pattern across all targets. I also added a positive graph check so the `cli` feature cannot silently degrade to the supported local-only viewer configuration; that configuration would still compile successfully.

dial9/Cargo.toml
kev009: Took the direct-viewer approach. `dial9` remains `default = []`, so the facade never includes the viewer by default. Its `cli` feature explicitly enables the S3-capable viewer, preserving `cargo install dial9 --features cli`. I dropped `cli-local`; the documented local-only path is now `cargo install dial9-viewer --no-default-features`. CI verifies that direct viewer configuration is AWS-free across all targets.

kev009:
Verification is clean: viewer unit tests pass with and without S3 (351/321), the full suite passes 1129/1129, the 20-second stress run passes 1129/1129, and formatting, clippy, both dependency guards, and both documentation configurations pass. The remaining rustdoc warnings are pre-existing and unrelated to this PR.

jv1i:
Thanks for the updates @kev009 it looks good now. You'll need to solve conflicts, one thing we did separately is bring the s3 uploader into a destinations crate in https://github.com/dial9-rs/dial9/pull/742. So you no longer need to deal with the uploader-side of things in this PR, just the viewer.

kev009:
@yulnr thanks.  I squashed the rebase.

jv1i (approved):

dial9/Cargo.toml
```diff
@@ -57,8 +57,11 @@ tracing = { version = "0.1.44", optional = true }
 [features]
 default = []
 
-## The `dial9` viewer/analysis binary (`dial9 serve`, `dial9 agents`).
-cli = ["dep:dial9-viewer", "dep:anyhow"]
+## The `dial9` viewer/analysis binary (`dial9 serve`, `dial9 agents`) with S3.
+## This preserves the existing `cargo install dial9 --features cli` behavior.
+## For a local-only viewer without the AWS SDK, use the `dial9-viewer` crate
+## directly with its default features disabled.
+cli = ["dep:dial9-viewer", "dep:anyhow", "dial9-viewer/s3"]
```
jv1i: actually, is `default-features = false`s doing anything now that `cli` turns `s3` back on? I think we get the same result if we drop `default-features=false` and also the `dial9-viewer/s3` from `cli` feature?
not sure if intentional now or leftover from `cli-local`

rcoh (approved):
lgtm

rcoh:
I'm planning on resolving merge conflicts and merging this today

Issue #672, opened 2026-07-16
UI deprecation 1 - Ship new UI as opt-in, keep legacy default


Fluzko:
Goal: New UI reachable but never forced. Legacy stays default, every migrated page shows a persistent "Switch to new UI" affordance.

Status: Mostly built already (ui-switch.js: DEFAULT_UI="legacy", all 4 pages in NEW_UI_ENTRIES, sticky dial9-ui-preference, ?ui= override, bottom-right pill). This issue is the close-out gate before flipping, not fresh work.

Scope / guideline:
- Confirm all 4 surfaces (index, viewer, flamegraph, tokio_stats) migrated + registered, each new entry calls mount({ side: "new" }).
- Switch pill visible on both sides, sticky pref survives reload, ?ui=new / ?ui=legacy honored.
- Legacy + new built side by side in dist/ and embedded (rust-embed UiAssets).

Done when: switch works both directions on every page, preference persists, tests/ui_switch.test.ts green. This is the entry criterion for the next stage.

Next step: #673

PR #723, opened 2026-07-26
feat!: metrique → dial9 sink

### LINKED ISSUE #189: Integrate metrique events into dial9

jlizen:
Closes #189.

Dial9 can now record [metrique](https://docs.rs/metrique) unit-of-work entries into the trace as a peer of the user's EMF/JSON pipeline, putting per-request application metrics on the same timeline as runtime telemetry. Design record: `docs/design/metrique-integration.md`, updated in this PR to the shipped state.

### Summary

New `metrique-sink` feature in dial9-tokio-telemetry, re-exported as `dial9::metrique_sink`, built on metrique's entry-descriptor and field-flag system. An entry opts in by including a `Dial9Context`, which is also what supplies the worker, task, and start/end context; every field of an opted-in entry is then recorded, and `Skip` excludes individual ones. Entries without a context record nothing, so teeing the sink into an existing pipeline picks up only what you asked for. `Dial9Stream::tee(&handle, emf)` wires dial9 alongside the pipeline you already have. No dial9-specific macro or field newtypes.

Every dial9-owned name is `dial9.`-prefixed, on the wire and in the other formats. That removes a whole class of collision (an entry is free to carry its own `worker_id`) and makes the fields filterable, which is what `Dial9Stream::tee` does: it wraps the other sink so `dial9.worker_id` and the monotonic timestamps stay out of your EMF output. Filtering covers both the values and the descriptor, so a descriptor-aware downstream sink still sees a descriptor that matches the values it received. A general `drop_fields` in metrique itself would be the better long-term home, but starting in dial9.

For entries that cannot take a new field (a shared struct, or dial9 support that should be switchable from one place), `append_on_drop_dial9` wraps the entry instead of editing it, producing the same event. It derefs to the wrapped entry so call sites are otherwise unchanged.

The sink builds an encode plan (wire schema with `unit` annotations, per-field routing) the first time it sees a descriptor sequence; after that, each entry only re-derives its sequence id to find the cached plan (the identification cost tracked in awslabs/metrique#348). Value routing is purely positional, leaning on metrique's guarantee that descriptor segments enumerate fields in exactly `Entry::write` order. A walk whose callback count differs from the descriptor in either direction means that contract broke, and the event is dropped with a rate-limited warning rather than recorded with mis-attributed values. List fields (`Vec<T>`, slices) encode as typed `DynamicList` values through the structured `values()` callback, with `Interned` pooling each string element. `Flex` entries carry no descriptors and are skipped with a diagnostic. Unit annotations are normalized to the vocabulary the viewer already formats. Events carry the request start as their timestamp and a `dial9.duration_ns` header field rather than an absolute end timestamp: durations varint-encode in a fraction of the bytes and are what consumers want anyway.

Viewer span rendering for these events and a `Kpi` flag are follow-up scope; there is an in-flight UI refactor (#672-674) and I don't want to step on toes. The design doc's Future evolution section tracks the full list.

Measured overhead (criterion bench included): ~26ns for `Dial9Context::capture()` on the request path plus ~30ns for the close-time clock read, ~430-580ns per entry on the metrique flush thread in steady state (all-scalar payload up to one allocating string; boxed entries from a global sink add ~60ns), ~3.5ns per entry with a paused recorder or disabled handle. The steady-state number is roughly 2x what EMF-formatting the same fields costs (~315ns to `io::sink()`). The largest attributed chunk, ~150ns, is per-entry descriptor identification, which can't be hoisted today because the descriptor sequence is a property of the entry instance, not its type (`Option`-flattened children change it at runtime); awslabs/metrique#348 proposes a macro-precomputed sequence id that would reduce identification to a single `u64` read. The comparison is also not quite apples-to-apples: the dial9 event carries four context header fields the EMF bench doesn't, plus panic isolation around user `Value` impls. Note that unlike the tracing layer (whose ~300ns per span enter+exit lands on the worker thread, inside the request path), all of this runs in the background: the request path only pays the capture.

Also fixes a latent panic path in `ThreadLocalEncoder::write_event` (validation errors now drop the event with a rate-limited error instead of panicking the calling thread) and teaches trace-shape's known signatures about `QueueSampleEvent.active_tasks`, which the regenerated demo trace exposed. Two small cross-crate additions ride along: `Dial9Handle::is_recording` (dial9-core), so sources can skip per-entry work while recording is paused, and an identity fast path for repeated dynamic `Schema` handles in the trace-format encoder, which also benefits the tracing layer's per-span events.

### Testing

Integration tests round-trip entries through `Dial9Stream` (directly and via a real `BackgroundQueue`) into a disk trace and decode it back, covering the things you would expect: payload shapes and units, list encoding, the context header, and the failure paths (panicking values, descriptor/write mismatches, unsupported shapes).

The trace-integrity vitest suite asserts the demo trace contains metrique events with the prefixed context fields and normalized unit annotations, so a future demo regen that loses them fails CI. The demo trace is regenerated here for the new field names; the demo-pinned flamegraph anchors moved with it, verified as capture drift rather than a code regression by replaying the previous trace bytes against the current code.

That flamegraph anchoring is pretty annoying... you have to regenerate them every time you regenerate the trace file... but not trying to address it now.


jlizen:
Setting this to a draft while I fix: https://github.com/awslabs/metrique/issues/345

The workaround here is ugly and I see a decent way to fix it in metrique. That will simplify our sink implementation a good bit since we no longer need to walk the entry and cache seen fields.

dial9-tokio-telemetry/src/metrique_sink/plan.rs
```diff
@@ -0,0 +1,498 @@
...
+            // the element shape so garbage text is rejected up front.
+            validate_element(elem.get())?;
+            if interned {
+                (FieldType::PooledString, ValueKind::Str { interned: true })
+            } else {
+                (FieldType::String, ValueKind::Str { interned: false })
+            }
+        }
+        FieldShape::Flex { .. } => return Err("dynamic-key (Flex) fields are not supported"),
+        FieldShape::Optional(_) => return Err("nested optional shapes are not supported"),
```
jlizen: This restriction will go away as I refactor once the fix for.https://github.com/awslabs/metrique/issues/345 lands

jlizen:
Ok, fixed both of the bugs I ran into (flattened fields not interleaving in descriptor ordering, and not properly passing `values()` through wrappers). I'm going to land those in metrique so that I can refactor this PR properly. It will stay in draft in the meantime.

jlizen:
I cut over to the proper implementation with the upstream metrique fixes staged locally. It won't pass CI here since it expects versions that aren't released yet. I'll move from draft -> published once new metrique versions are cut and CI goes green.

jlizen:
Ok, metrique fixes are released upstream, CI should be able to go green now.

I also addressed some offline feedback from rcoh@ (as well as in #684 ):
- opt-out fields rather than opt-in, and opt-in entries in general by using `Dial9Context`
- add a `Dial9Stream::tee()` helper that filters out the `dial9.*` fields from other streams (we can upstream this functionality more generically into metrique later)
- add a `.append_on_drop_dial9()` helper that adds the `Dial9Context`, friendlier for conditional compilation of dial9

dial9-tokio-telemetry/src/metrique_sink/context.rs
```diff
@@ -0,0 +1,103 @@
...
+#[metrics(subfield)]
+#[derive(Debug)]
+pub struct Dial9Context {
+    /// Global dial9 worker id of the capturing thread
+    /// (`WorkerId::UNKNOWN` when off-runtime).
+    #[metrics(name = "dial9.worker_id", flags(Context))]
+    worker_id: u64,
+    /// Tokio task id of the capturing task, absent when not inside a task.
+    #[metrics(name = "dial9.task_id", flags(Context))]
+    task_id: Option<u64>,
```
rcoh: wdyt about splitting this into `tokio context`? fwiw, you really only need task_id & possibly thread id since you can infer worker id.

dial9-metrique/src/filter.rs
```diff
@@ -0,0 +1,136 @@
...
+use std::io;
+use std::time::SystemTime;
+
+use metrique_writer::core::descriptor::Descriptors;
+use metrique_writer::core::entry::SampleGroupElement;
+use metrique_writer::{Entry, EntryConfig, EntryIoStream, EntryWriter, IoStreamError, Value};
+
+/// The prefix every dial9-owned field name carries (see
+/// [`Dial9Context`](super::Dial9Context)).
+const DIAL9_PREFIX: &str = "dial9.";
```
rcoh, line 20: I guess the other alternative is using a forceflag. I am _slightly_ worried about conflicting with actual metrics beginning with `dial9.`

thoughts?

dial9-metrique/src/filter.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    fn sample_group(&self) -> impl Iterator<Item = SampleGroupElement> {
+        // Sample groups are keyed by name/value pairs the entry chooses;
+        // dial9's context fields are not sample-group members, so there is
+        // nothing to filter.
+        self.0.sample_group()
+    }
+
+    fn descriptors(&self) -> Descriptors<'_> {
```
rcoh, line 66: does descriptors get called a lot assuming it is cheap? or is that an expectation?

dial9-tokio-telemetry/src/metrique_sink/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+//! Requires the `metrique-sink` feature.
+//!
+//! # Usage
+//!
+//! An entry opts into the dial9 trace by including a
+//! [`Dial9Context`](crate::metrique_sink::Dial9Context). Every field of an
+//! opted-in entry is recorded; exclude individual fields with the
+//! [`Skip`](crate::metrique_sink::Skip) field flag:
+//!
+//! ```ignore
```
rcoh: `ignore` example. Can this be no-run?

dial9-tokio-telemetry/src/metrique_sink/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+//! into an existing pipeline therefore only records the entries you opt in.
+//!
+//! ## Opting in without touching the entry
+//!
+//! Adding a field is intrusive when the metrics struct is shared, owned by
+//! another team, or when dial9 should be switchable from one place.
+//! [`Dial9EntryExt`](crate::metrique_sink::Dial9EntryExt) attaches the same
+//! context from the outside, so the entry definition stays as it was:
+//!
+//! ```ignore
```
rcoh: same here

dial9-metrique/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+//! you would rather keep them.
+//!
+//! All dial9 encoding happens on the thread that drives the metrique
+//! pipeline (the `BackgroundQueue` flush thread for the standard setup).
+//!
+//! # Overhead
+//!
+//! Measured by `benches/metrique_sink_bench.rs`: `Dial9Context::capture()`
+//! costs ~26 ns on the request path, plus ~30 ns for the end-timestamp
+//! clock read when the entry closes. Encoding costs ~430-580 ns per entry
```
rcoh: higher than I would have expected

dial9-tokio-telemetry/src/metrique_sink/writer.rs
```diff
@@ -0,0 +1,483 @@
...
+    out
+}
+
+/// The distribution's only observation, or `None` when it is empty or
+/// carries more than one.
+fn single_observation(distribution: impl IntoIterator<Item = Observation>) -> Option<Observation> {
+    let mut iter = distribution.into_iter();
+    let first = iter.next();
+    if iter.next().is_some() { None } else { first }
+}
```
rcoh: interesting. Dropping data when it contains more than 1 observation is one possibility. I guess its reasonable. Seems like just taking the first one is probably _basically_ as useful. Worth considering.

dial9-trace-format/src/encoder.rs
```diff
@@ -168,6 +168,12 @@ pub struct Encoder<W: Write = Vec<u8>> {
     stack_pool: FxHashMap<Box<[u64]>, u32>,
     next_stack_pool_id: u32,
     schema_ids: FxHashMap<SchemaKey, WireTypeId>,
+    /// Identity fast path for dynamic [`Schema`] handles: wire ids keyed by
+    /// the address of the schema's shared `SchemaEntry` allocation. The
+    /// `Arc` is kept alive in the value so the address cannot be reused
+    /// while cached. Repeated `write_event` calls with the same handle skip
+    /// the name hash and deep schema comparison in `ensure_registered`.
```
rcoh: considering if we should split this into its own PR

jlizen:
From rcoh@ offline:
> I think we probably want it either in dial9-util or its own crate. I don't think we want anything other than tokio in dial9_tokio_telemetry now
> should be a pretty easy lift-and-shift. For task_id / worker_id, you can call it from tokio directly
> otherwise this will only be usable if you're using tokio

Good point, updating

dial9-metrique/tests/metrique_sink.rs
```diff
@@ -0,0 +1,1548 @@
...
+            format!("[{}]", rendered.join(","))
+        }
+        FieldValueRef::None => "<none>".to_owned(),
+        other => format!("{other:?}"),
+    }
+}
+
+/// Decode all `metrique:*` events from every trace segment file in `dir`
+/// (including the live `.active` segment).
+fn decode_metrique_events(dir: &std::path::Path) -> Vec<DecodedEvent> {
```
rcoh, line 57: can we use the `derive(Deserialize)` approach here?

dial9-trace-format/src/encoder.rs
jlizen: I tacked it on while trying to get the performance overhead lower on the flush thread. Open to splitting it out if you think it makes this unreadable? Or is the concern about being able to revert it atomically?

dial9-tokio-telemetry/src/metrique_sink/context.rs
```diff
@@ -0,0 +1,103 @@
...
+/// want them.
+//
+// Deliberately neither `Default` nor `Clone`: a defaulted context would
+// silently look like "valid context captured at monotonic time 0".
+#[metrics(subfield)]
+#[derive(Debug)]
+pub struct Dial9Context {
+    /// Global dial9 worker id of the capturing thread
+    /// (`WorkerId::UNKNOWN` when off-runtime).
+    #[metrics(name = "dial9.worker_id", flags(Context))]
```
rcoh: the viewer won't recognize these for correlation; should open an issue to track

examples/metrics-service/src/main.rs
```diff
@@ -261,6 +264,11 @@ fn main() -> std::io::Result<()> {
         },
     )?;
 
+    // Per-request metrique entries (routes::RequestMetrics) flow into the
+    // dial9 trace. A production service would keep its EMF stream and use
+    // `Dial9Stream::tee(recorder.handle(), emf_stream)` instead.
+    let metrics_join = ServiceMetrics::attach_to_stream(Dial9Stream::new(recorder.handle()));
```
rcoh: no reason not to also tee out EMF metrics IMO

dial9-metrique/src/lib.rs
jlizen: For the flush side, or the capture side?

dial9-core/src/encoder.rs
```diff
@@ -80,14 +85,25 @@ impl ThreadLocalEncoder<'_> {
...
+    /// (logged rate-limited; never panics the calling thread).
     pub fn write_event(
         &mut self,
         schema: &dial9_trace_format::encoder::Schema,
         values: &[dial9_trace_format::types::FieldValue],
-    ) {
-        self.encoder
-            .write_event(schema, values)
-            .expect("writing to Vec<u8> is infallible");
+    ) -> bool {
```
rcoh: should this return `Result<()>` instead of a bool to clarify?

I'm not positive we want the logging on this side anyway.

rcoh: note that we need to make sure to land this before 0.5.0 assuming we _do_ want to change this API

dial9-metrique/src/filter.rs
jlizen: `EntryWriter::value` don't carry flags so either way we would need to use the name in the value filter. I think probably the risk of collisions is low enough to be fine..?

dial9-metrique/src/filter.rs
jlizen: It's decently cheap and is generally called on a background thread, and only by descriptor-aware sinks (of which currently none exist).

THe more concerning thing is the extra overhead from `write()`, which I would expect to be a ~10s of nanos on the flush thread, I can add a bench for that.

dial9-tokio-telemetry/src/metrique_sink/writer.rs
jlizen: I'm fine with taking first one

dial9-metrique/src/filter.rs
jlizen: Actually, it's not bad at all, <1ns on the write path, nevermind.

jlizen:
Addressed review feedback:
  - Extracted the sink into a new tokio-free `dial9-metrique` crate (dial9-core deps only); `tokio` feature adds task-id capture, enabled by the facade iff `metrique-sink` + `tokio`
  - Context now records OS thread id (worker inferable) instead of worker id
  - Multi-observation distributions record the first observation instead of dropping the field
  - Doc examples are `no_run`; tests use typed serde decode

dial9-tokio-telemetry/src/metrique_sink/context.rs
jlizen: I don't think there's a ton of benefit to splitting it out specifically into a `TokioContext`, I instead gated the task id behind a `tokio` feature on `dial9-metrique`. Agreed on using TID over worker ID.

dial9-metrique/src/filter.rs
rcoh: ah because flags only go via `metric` gotcha. Possibly we could tag them as dial9 via entry descriptors?

dial9-metrique/src/lib.rs
rcoh: encoding at ~500ns per entry

dial9-trace-format/src/encoder.rs
rcoh: partially just that is slightly complex and we may want to test it independently, ensure it doesn't have resource leaks etc.

One example is that right now, if you accidentally call it with many Arc's to the same context, it leaks. Which is maybe OK? But also makes me a little worried since people often use dial9 in weird ways

dial9-metrique/src/context.rs
```diff
@@ -80,14 +80,24 @@ impl Dial9Context {
...
             monotonic_ns_start: clock_monotonic_ns(),
             monotonic_ns_end: MonotonicAtClose,
         }
     }
 }
 
+/// The calling thread's OS thread id, read once per thread and cached: a
+/// thread's tid never changes, and `gettid` is a real syscall on the
+/// capture (request) path otherwise.
+fn cached_tid() -> u64 {
```
rcoh: should we upstream caching the tid into dial9-core?"

dial9-core/src/handle.rs
```diff
@@ -77,6 +77,18 @@ impl Dial9Handle {
...
 
+    /// Whether this handle is connected to a live recorder that is currently
+    /// recording (i.e. not paused via [`disable`](Self::disable)).
+    ///
+    /// Cheaper than attempting a record: sources that do per-event work
+    /// before reaching [`with_encoder`](Self::with_encoder) can skip it
+    /// entirely while recording is off. The check is a relaxed atomic load;
+    /// racing a concurrent enable/disable is benign (the event lands or is
+    /// skipped, exactly as if it had arrived a moment earlier or later).
+    pub fn is_recording(&self) -> bool {
```
rcoh: should we make `is_enabled` tell the truth instead of adding a second method?

dial9-metrique/benches/metrique_sink_bench.rs
```diff
@@ -0,0 +1,256 @@
...
+use metrique::unit::Microsecond;
+use metrique::unit_of_work::metrics;
+use metrique_writer::{EntryIoStream, IoStreamError};
+use std::hint::black_box;
+
+#[metrics(rename_all = "PascalCase")]
+struct RequestMetrics {
+    #[metrics(flatten)]
+    dial9: Dial9Context,
+    #[metrics(flags(Interned))]
```
rcoh, line 32: designing the module structure so you can name this something like `dial9::Interned` would be nice for visual clarity

dial9-metrique/src/lib.rs
jlizen: ~180ns are from the descriptor identifications, we do that every entry due to optional flattened. https://github.com/awslabs/metrique/issues/348 would help avoid that

~185 ns are from metrique's generated write + our EntryWalk::value dispatch + ValueCapture + the one String alloc in our bench, the main optimization opportunity I see there would be the streaming decoding to save the string alloc, #724 would shave maybe ~60-120ns off of that

There is ~70ns on wire encoding, and then it goes down from there for smaller pieces.

dial9-trace-format/src/encoder.rs
jlizen: Fair, I"ll split it off

dial9-core/src/handle.rs
jlizen: this is a bit lossy but I think it's probably the right choice, as we queue up for 0.5

dial9-metrique/benches/metrique_sink_bench.rs
jlizen: Sure, I can add some re-exports

dial9-metrique/src/filter.rs
jlizen: That would be a fair bit more expensive since we would need to call `descriptors()` per entry and build the positional map, essentially run the sink machinery on the `tee()` side (which currently is basically free). If we had https://github.com/awslabs/metrique/issues/348 it would maybe be more tractable.

jlizen:
Updated to address feedback:

 -   split off identity fast path, I'll put it in its own PR (that costs ~75ns/entry on the flush path that we'll get back when the other PR runs)
  -  moved cached_tid to dial9-core
  -  write_event returns io::Result<()> with logging moved to callers (and added similar logging to the tracing layer) - this is a breaking change
  -  flush metrics-service metriques to EMF as well
  - changed `is_enabled` to also incorporate "is it connected" rather than only "Is dial9 initialized" (this is technically a breaking behavioral change)
  - re-export the flags to allow eg `dial9:Interned`

rcoh (approved):
Should we show append-on-drop dial9?

jlizen:
> Should we show append-on-drop dial9?

@rcoh you mean in the changelog?

rcoh:
Yeah (or anywhere else we document it)

jlizen:
I suppose, I don't really like the vibes of that approach compared to inlining it in the struct, but I can put both?

PR #727, opened 2026-07-27
fix(core): implement Debug for Recorder, FlushContext, and perf

jv1i:
Implements debug for some pub items that were missing it.


rcoh (approved):
There is a lint for missing debug impls; please add it

PR #732, opened 2026-07-28
feat(tokio): attach a runtime from a shared Dial9Handle

jv1i:
Some thread-per-core services build each core's runtime on that core's own thread, independently, for locality. (spotted `aws-sdk-s3-transfer-manager` doing this). Feeding one recorder into all of them needs attach that doesn't
consume `self`, today it does, and we'd rather no have users cloning the `Recorder` for it.

### Changes

`Dial9HandleTokioExt::attach_tokio_runtime(&self, builder, options) -> io::Result<Runtime>`, on `Dial9Handle`. Clone a handle once per thread (`recorder.handle().clone()`) and attach independently, no `Recorder` cloning or wrapping needed.

`RecorderTokioExt` and its consuming `Recorder::attach_tokio_runtime{,_with}` are gone, the handle method is the only attach. One method, always takes `TokioAttachOptions`. You build the `tokio::runtime::Builder` yourself, including `enable_all()` (previously seeded by the closure variant) and the flavor.

A `#[dial9::main]` config pairs the two itself now: `Ok((recorder, runtime))`. `recorder_from_env{,_with}` are unchanged.

Tests go through a local `common::attach(&recorder, workers, options)` helper for the builder boilerplate, examples and docs spell out the real calls.

### Why Handle

Attach doesn't own new `Recorder` infrastructure, it just wires a runtime into state a handle already exposes, same as `record_event`/ `track_current_thread`, and it's already convenient to use across threads.


dial9/Cargo.toml
```diff
@@ -138,6 +138,7 @@ linux-socket = [
 dial9 = { path = ".", features = [
     "tokio",
     "analysis",
+    "cpu-profiling",
```
jv1i, line 146: unrelated, dev dep needed for some docs

dial9-tokio-telemetry/src/telemetry/recorder/recorder_tokio.rs
rcoh: this needs to return self so we can manage both together?

this method could also take `&self`

rcoh: _slightly_ confusing to have attach_tokio_runtime_with and attach_tokio_runtime that behave slightly differenlty.

any way to unify them? its ok if not.

dial9/tests/recorder_attach.rs
```diff
@@ -290,3 +290,92 @@ fn on_recording_start_fires_at_enable() {
...
+        CORES * 6
+    );
+    assert!(
+        worker_ids.len() >= CORES,
+        "expected poll events from {CORES} distinct worker IDs, got {worker_ids:?}"
+    );
+    assert_eq!(
+        poll_starts, poll_ends,
+        "PollStart ({poll_starts}) != PollEnd ({poll_ends})"
+    );
```
rcoh: weird assertions. why would worker_ids.len() ever be more than CORES?

poll_starts * CORES is also a weird assertion; we can use `derive(Deserialize)` and parse the worker ids out of the events.

dial9/tests/recorder_attach.rs
```diff
@@ -290,3 +290,92 @@ fn on_recording_start_fires_at_enable() {
...
+    recorder.graceful_shutdown(Duration::from_secs(1));
+
+    #[derive(serde::Deserialize)]
+    struct WorkerIdField {
+        worker_id: u64,
+    }
+    const UNKNOWN_WORKER_ID: u64 = 255;
+
+    let bytes = std::fs::read(dir.path().join("trace.0.bin")).expect("sealed segment");
+    let mut decoder = Decoder::new(&bytes).expect("valid trace");
```
rcoh, line 351: the thing I _really_ want to see in this test is confirmation we are managing segment metadata properly and producing the runtime-to-worker mapping that we would expect.

rcoh (approved):
direction lgtm

PR #733, opened 2026-07-28
Decode annotated single-event spans

rcoh:
## Summary
- decode producer-independent single-event spans from `dial9.role` schema annotations in both Rust and JavaScript
- generalize over the three timing quantities — start, duration, and end (`end = start + duration`): a schema declares **any two** and the decoder derives the third, with the packed event timestamp counting as an end
- preserve producer-defined span type, dynamic span name, execution context, and non-structural attributes
- keep legacy tracing reconstruction unchanged and treat unannotated `metrique:*` events as ordinary custom events
- support both fat and columnar UI paths with poll-backed active-time and sample attribution

The format and validation contract are documented in `docs/design/single-event-spans.md`.

Rebased onto `main` now that #736 has merged.

### Realigned to the merged #736 format

#736 landed two format changes this branch's decoder predated:

1. the second timing quantity is `span.duration`, not `span.start` — the packed event timestamp is the span **end**, so `start = end - duration`
2. an absent `unit` annotation defaults to `ns` instead of being a validation error

Rather than swap one role for the other, both decoders now handle any two of start, duration, and end (the packed event timestamp). This covers both metrique's duration encoding and the older start encoding.

`dial9-viewer/ui/tests/core/single_event_span_timing.test.ts` covers the timing-resolution matrix and the schema validation rules, and mirrors the Rust-side tests so the two decoders cannot drift.

An earlier revision of this branch also added a `span.end` role. No producer emits it and it was not part of #736's spec, so it has been removed — the packed event timestamp is always the span end.

Also untracks `dial9-viewer/ui/node_modules`, an absolute-path symlink committed by mistake (`.gitignore` had `node_modules/` with a trailing slash, which matches only a directory).

## Verification
- `cargo nextest run` (1,328 passed)
- `cargo nextest run --stress-duration 20s` (1 iteration, 1,328 passed)
- `cargo fmt --all --check`
- `npm run test` (1,940 Vitest tests passed, 11 skipped; all legacy Node suites passed)
- `npm run check:types`
- `npm run build`
- `cargo build --release -p dial9-viewer`
- demo trace decodes 24,970 metrique single-event spans (p50 630ns, no negative durations); verified end-to-end in the Spans track and the Span Explorer (histogram, poll-backed time composition, exemplars)


dial9-viewer/ui/trace_parser.js
```diff
@@ -760,13 +1168,51 @@
...
         const memoryOverflows = state.memoryOverflows;
         const threadNames = state.threadNames;
-        const tidToWorker = state.tidToWorker;
         const runtimeWorkers = state.runtimeWorkers;
         const segmentMetadata = state.segmentMetadata;
         const taskDumps = state.taskDumps;
         const customEvents = state.customEvents;
         const spanEventSink = state.spanEventSink;
         const clockSyncAnchors = state.clockSyncAnchors;
+        if (singleEventSpan != null) {
```
jlizen, line 1177: This would "freeze" something as a standard event if its span data came in as a subequent annotation event. We should probably commit things only after `finalizeParse`?

jlizen: We also would have a problem if something arrived with `has_timestamp = false` but was classified as a regular event until subsequent annotation frames. We probably want to disable that discard until finalize as well.

jlizen: Would be good to have fixtures probing both the above

jlizen (approved):

dial9-viewer/ui/trace_parser.js
jlizen: looks like the rust parser has the same problem?

PR #736, opened 2026-07-29
Switch metrique to the generalized single-event span format

rcoh:
## Summary
- define the producer-independent single-event span schema convention
- keep the wire format annotation-agnostic and expose Dial9's convention constants from `dial9-core`
- emit metrique close times as packed event timestamps and annotated start/context fields
- add `dial9::SpanName` for selecting a dynamic span name field
- regenerate the demo trace and update its integrity/search fixtures

The packed timestamp is the close time so delta encoding remains in emission order. Span recognition is annotation-driven; decoders must not match the `metrique:` schema prefix.

This is PR 1 of 2. The generalized decoder in #733 is temporarily stacked on this PR for review and will be rebased onto `main` after this PR merges.

## Verification
- `cargo nextest run` (1,283 passed)
- `cargo nextest run --stress-duration 20s` (1,283 passed)
- `npm run test` in `dial9-viewer/ui` (1,883 passed, 11 skipped; legacy Node suites passed)
- regenerated demo trace with 14,821 scheduler samples


dial9-metrique/src/plan.rs
```diff
@@ -138,34 +142,41 @@ pub(crate) struct Plan {
...
     pub(crate) const ALL: [Header; 4] = [
         Header::ThreadId,
         Header::TaskId,
-        Header::Duration,
+        Header::Start,
         Header::WallClock,
     ];
 
-    fn name(self) -> &'static str {
+    pub(crate) fn name(self) -> &'static str {
```
jlizen, line 156: i recommend keeping the dial9. prefix to avoid collisions with userspace fields.

dial9-metrique/src/plan.rs
```diff
@@ -234,20 +303,54 @@ pub(crate) fn build_plan(
...
+                        if matches!(kind, ValueKind::Scalar(ScalarKind::Str { .. })) {
+                            if let Some(index) = field_index {
+                                annotations.push(FieldAnnotation::new(
+                                    index,
+                                    schema_extensions::ROLE_KEY,
+                                    roles::SPAN_NAME,
+                                ));
+                            }
+                        } else {
+                            invalid = true;
```
jlizen: I don't think invalid `SpanName` here (or other places) should disable the entire entry... can't we just fall back to the default display name in that place?

dial9-metrique/src/plan.rs
```diff
@@ -138,34 +142,41 @@ pub(crate) struct Plan {
...
         match self {
-            Header::ThreadId | Header::TaskId | Header::Duration | Header::WallClock => {
-                FieldType::OptionalVarint
-            }
+            Header::ThreadId | Header::Start => FieldType::Varint,
+            Header::TaskId | Header::WallClock => FieldType::OptionalVarint,
+        }
+    }
+
+    fn index(self) -> u16 {
```
jlizen: This means that `Header::ALL` is no longer source of truth, can't we just do `Header::ALL.into_iter().enumerate()` here?

dial9-metrique/src/writer.rs
```diff
@@ -110,26 +111,24 @@ impl<'p, 'enc> EntryWalk<'p, 'enc> {
             }
         }
 
-        // Timestamp: request start, or the flush-thread clock as fallback.
-        self.values[0] =
-            FieldValue::Varint(self.ctx.monotonic_start.unwrap_or_else(clock_monotonic_ns));
+        let thread_id = self.ctx.thread_id.ok_or(WalkError::MissingContext {
```
jlizen, line 119: Should we have a check for `start <= end` here?

dial9-metrique/src/writer.rs
```diff
@@ -110,26 +111,24 @@ impl<'p, 'enc> EntryWalk<'p, 'enc> {
...
+            field: Header::ThreadId.name(),
+        })?;
+        let start = self.ctx.monotonic_start.ok_or(WalkError::MissingContext {
+            field: Header::Start.name(),
+        })?;
+        let end = self.ctx.monotonic_end.ok_or(WalkError::MissingContext {
+            field: "event timestamp",
```
jlizen: nit: any reason not to have it be a a wire field name here like its siblings?

docs/design/single-event-spans.md
```diff
@@ -0,0 +1,172 @@
...
+captured at span start. `thread_id` is an OS thread ID.
+
+Tokio task ID is the preferred correlation key. A thread ID must be resolved
+against the time-local `WorkerPark`/`WorkerUnpark` timeline; it must not be
+treated as having one stable worker mapping for the whole trace. A supplied
+worker ID must not override a known block-in-place gap.
+
+### Attributes
+
+Fields carrying structural roles are not copied into the span's attributes.
```
jlizen: This is not currently true, we include `span.name`, I would carve that out from this rule.

docs/design/single-event-spans.md
```diff
@@ -0,0 +1,172 @@
...
+```text
+[start, end)
+```
+
+`start <= end` is required. Conversion to nanoseconds and all timestamp
+arithmetic must be checked.
+
+### Span name
+
+At most one string or pooled-string field may carry:
```
jlizen: or optional variant of either

docs/design/single-event-spans.md
```diff
@@ -0,0 +1,172 @@
...
+
+A schema is a single-event span schema when exactly one field carries:
+
+```text
+dial9.role = span.start
+```
+
+The event's packed timestamp is the span end. The annotated start field and
+the packed end timestamp are unsigned monotonic-clock values. The start
+field's `unit` annotation declares its scale; v1 supports `ns`, `us`, `ms`,
```
jlizen: We default to ns if not stated right? Needs definition.

docs/design/single-event-spans.md
```diff
@@ -0,0 +1,172 @@
...
+
+For metrique the value is `metrique`. This is the producer/instrumentation
+family, corresponding to the decoded span kind. The schema name remains the
+structural type discriminator.
+
+If the annotation is absent, the decoded type is `single-event`.
+
+### Execution context
+
+Optional integer fields may carry one of these roles:
```
jlizen: nit: Maybe rephrase to `These roles are integer fields and are optional`? Since they aren't actually modeled as optional on the wire

dial9-metrique/src/writer.rs
```diff
@@ -110,26 +111,24 @@ impl<'p, 'enc> EntryWalk<'p, 'enc> {
...
-        // Timestamp: request start, or the flush-thread clock as fallback.
-        self.values[0] =
-            FieldValue::Varint(self.ctx.monotonic_start.unwrap_or_else(clock_monotonic_ns));
+        let thread_id = self.ctx.thread_id.ok_or(WalkError::MissingContext {
+            field: Header::ThreadId.name(),
+        })?;
+        let start = self.ctx.monotonic_start.ok_or(WalkError::MissingContext {
+            field: Header::Start.name(),
+        })?;
+        let end = self.ctx.monotonic_end.ok_or(WalkError::MissingContext {
```
jlizen, line 125: We previously used one absolute timestamp + delta encoding to save a few bytes on the wire. Now we have two absolute. Is there a reason not to go to eg `end + duration` (and calculate start from that), even if we switch the packed timestamp over to close?

dial9-metrique/src/writer.rs
```diff
@@ -110,26 +111,24 @@ impl<'p, 'enc> EntryWalk<'p, 'enc> {
...
+            field: "event timestamp",
+        })?;
+
+        // Completed entries are emitted at close, so the packed timestamp is
+        // the close time. This keeps delta encoding in emission order.
```
jlizen: nit: there still are race conditions that put them out of order, maybe `This mostly keeps...`

dial9-metrique/src/plan.rs
```diff
@@ -257,6 +360,14 @@ pub(crate) fn build_plan(
                     });
                 }
                 Err(err) => {
+                    if span_name {
```
jlizen, line 354: this branch appears untested?

dial9-metrique/src/writer.rs
jlizen, line 119: All of these missing context cases appear unreachable in the public API, are these guards just defensive against an internal misimplementation? Might be worth a note if so.

I also would move this check the plan layer, probably, rather than the writer layer? This seems mostly redundant and we could just assert against the required roles (whereas we currently just track that >=1 role was seen).

dial9-metrique/tests/metrique_sink.rs
```diff
@@ -289,10 +336,160 @@ fn full_entry_round_trips() {
...
+        "test setup must reverse start and close order"
+    );
+    assert!(
+        events[0].timestamp_ns <= events[1].timestamp_ns,
+        "packed timestamps must follow close/emission order: {events:#?}"
+    );
+}
+
+#[test]
+fn packed_timestamp_is_captured_at_close_not_encoding() {
```
jlizen, line 398: Would it be worth asserting against more specifically that the close falls within a range of reading the clock before/after in addition to just that it is before encoding?

dial9-viewer/ui/tests/core/flamegraph_recipes.test.ts
```diff
@@ -144,24 +145,38 @@ describe("flamegraph recipes", { timeout: 60_000 }, () => {
...
-        longPolls.some((p) => s.timestamp >= p.start && s.timestamp <= p.end),
+        longPolls.some(
+          ({ poll, workerId }) =>
+            s.workerId === workerId &&
+            s.timestamp >= poll.start &&
+            s.timestamp <= poll.end,
+        ),
     );
-    expect(samples.length, `expected >0 samples for long polls`).toBeGreaterThan(0);
+    const attached = longPolls
```
jlizen, line 173: Do we want to preserve the assertion that there was at least one sample inside a long poll?

dial9-metrique/src/lib.rs
```diff
@@ -176,7 +181,29 @@ pub use stream::Dial9Stream;
...
+
+    /// Tokio task ID captured when the metrique entry starts.
+    pub const TOKIO_TASK_ID: &str = "dial9.tokio.task_id";
+
+    /// Optional wall-clock timestamp from metrique's timestamp field.
+    pub const WALL_CLOCK_NS: &str = "dial9.wall_clock_ns";
+}
+
+/// Value emitted under the `dial9.span.type` annotation for metrique spans.
+pub const SPAN_TYPE: &str = "metrique";
```
jlizen: Should this move into `dial9_core::schema_extensions`?

dial9-metrique/src/plan.rs
```diff
@@ -257,6 +360,14 @@ pub(crate) fn build_plan(
...
+                        invalid = true;
+                        tracing::error!(
```
jlizen: nit: probably these could be in a `invalidate()` helper

dial9-metrique/src/plan.rs
```diff
@@ -221,6 +281,15 @@ pub(crate) fn build_plan(
             // does: two fields with the same name would be indistinguishable
             // to trace consumers. First occurrence (headers included) wins.
             if fields.iter().any(|f| f.name() == full_name) {
+                if span_name {
```
jlizen: nit: a bit awkward to log twice here

dial9-core/src/schema_extensions.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub const SPAN_NAME: &str = "span.name";
+
+    /// OS thread ID captured at span start.
+    pub const THREAD_ID: &str = "thread_id";
+
+    /// Tokio task ID captured at span start.
+    pub const TOKIO_TASK_ID: &str = "tokio.task_id";
+
+    /// Tokio worker ID captured at span start.
+    pub const TOKIO_WORKER_ID: &str = "tokio.worker_id";
```
jlizen: Currently unused?

dial9-metrique/src/lib.rs
```diff
@@ -226,6 +253,37 @@ impl FlagConstructor for Interned {
...
+/// #[metrics(flags(dial9::Interned, dial9::SpanName))]
+/// operation: &'static str,
+/// ```
+///
+/// The field remains part of the event payload and receives the
+/// `dial9.role=span.name` schema annotation. [`Interned`] can be combined with
+/// this flag to pool repeated names. Without a `SpanName` field, viewers use
+/// the event's schema name.
+///
+/// Multiple `SpanName` fields, use on a non-string field, or combining it
```
jlizen: also enumerate colliding name?

jlizen (commented):
Refactor LGTM overall but some notes, main load-bearing stuff is the ~~`dial9.` prefixes~~, some spec gaps, and the duplicate enumeration with `Header::index()`.

I do also think it's probably a mistake to drop entire entries due to invalid spanname as opposed to logging the warning but preserving them without the display name.

dial9-metrique/src/plan.rs
jlizen: Ah nevermind, I see you prepend the prefix elsewhere

jlizen (approved):

PR #742, opened 2026-08-01
refactor(s3)!: extract the S3 uploader into a dial9-s3 crate

jv1i:
### Motivation

`dial9` currently depends on `dial9-utils` for exactly one thing, the S3 uploader. The rest of that crate is opt-in integration code customers reach for themselves, and `aws-sdk-s3`/`aws-config` aren't optional in it, so someone pulling it for the axum middleware compiles the AWS SDK too.

### Changes

S3 moves to its own `dial9-s3` crate: `s3.rs` and its two private helpers, the circuit breaker and the instance metadata. Types sit at the crate root, so it's `dial9_s3::S3Config`. Nothing in `dial9` depends on `dial9-utils`.

Through the facade S3 is a top level module `dial9::s3`, instead of `dial9::core::pipeline::s3`, it only lived there because it followed `S3PipelineUploader`, which really is a processor. `worker-s3` works the same as before.

`dial9-metrique` was missing from `release-plz.toml` and the CI `package` and `semver-checks` lists, so it's added there too.

### Publishing

Set the version of `dial9-s3`  at `0.5.0-rc0` just to match the other leaf crates, but the release PR will set everything at rc2 regardless. (dial9-s3 and dial9-metrique will need manual release before the rest)


jv1i:
oh semver check fails because I added dial9-s3 to it but it isn't yet in main, can keep it as is since it doesn't block CI Pass

rcoh (approved):
Maybe dial9-destination-s3? Or dial9-sink-s3? Whatever nomenclature we are currently using

PR #745, opened 2026-08-03
docs: design sampled task dump capture

rcoh:
## Summary

- Documents capture-time sampling for Tokio task dumps, with one worker-local decision before `trace_with` and no second emission sampler.
- Recommends a default budget of 10 captures/s/worker and records per-event inclusion probability for inverse-probability weighting.
- Defines task-scoped mixed CPU/async flamegraphs, source-owned segment metadata, API compatibility, and a runnable percentage-based profiling test app.

## Measurements

- Representative capture plus trim median: 7.65 us; planning cost: 10 us.
- At 10 captures/s/worker: 0.010% runtime capacity and 0.0428% of representative active CPU.
- Normal wake tracing: 0.0138% runtime capacity and 0.059% of representative active CPU.
- Combined planning estimate: 0.0238% runtime capacity and 0.102% of representative active CPU.

## Key design decisions

- Sampling covers every dial9-instrumented task on every attached worker; the initial mixed-flamegraph UI is single-task scoped to avoid zero-inclusion bias from uninstrumented tasks.
- `TokioRuntimesSource` emits task-dump configuration and per-worker sampling-active data as segment metadata.
- CPU weights use existing `cpu.profile.frequency_hz` segment metadata.
- The follow-on implementation includes an in-situ app whose function names encode their expected CPU, lock-wait, and timer-wait percentages.

## Validation

Docs-only change. `git diff --cached --check` passed before commit; no Rust or trace-format implementation changed.


rcoh:
Follow-up from wdyt review: I analyzed the current demo trace by grouping TaskDumpEvents on (task_id, timestamp_ns). Of 12,241 captures, 12,233 (99.935%) have multiple callchains; 11,053 (90.295%) are an I/O wait plus a persistent graceful-shutdown notification. This makes equal weight splitting actively misleading. Commit 4d7b4b33 now specifies deterministic representative-stack selection: demote proven timeout/cancellation control branches, prefer a uniquely application-specific branch, and emit one synthetic [awaiting any of N] stack when peer select branches remain ambiguous. Every capture receives its inverse-probability idle weight exactly once, while all raw callchains remain inspectable.

rcoh:
Split the reusable test workload into `docs/design/telemetry-integration-test-app.md`.

The new design specifies:
- one self-describing trace covering CPU profiles, spans, and task dumps;
- relative weights encoded in stable function-name tokens, registered by trace events so missing stacks remain detectable;
- trace-contained config, measurement windows, span topology, and task-dump branch roles, with no sidecar manifest;
- one normalized observation model validated independently through the local JS path and the Rust decode/Parquet/query path;
- a canonical fixture plus fresh-capture CI workflow and extension rules for future telemetry features.

The task-dump design now treats this app and both conformance adapters as an explicit prerequisite for mixed flamegraphs. Commit: `9b207f7e`.

rcoh:
Trimmed the telemetry integration-test design to a tracer bullet in `7c4d08cd` (455 lines to 187).

The prerequisite is now only:
- one instrumented task;
- two weighted CPU leaves and two weighted timer waits;
- one root span with CPU/wait children;
- expectation and measurement-marker events;
- one fresh trace checked through local JS and aggregate Rust/Parquet parsing.

Deferred until a feature proves it needs them: scenario registries, contract versioning, normalized result frameworks, canonical fixtures, server/S3 setup, multiple tasks, noise threads, branch matrices, worker coverage, and capture-rate sweeps. Task-dump flamegraph work can add the smallest required weighted-wait or branch case to this app later.

rcoh:
Reworked the integration-test design in `46411b91` around the actual core idea: the program and trace are an executable specification of the expected trace.

The harness now has no fixture-specific expectation table. Stable names encode domains and weights; expectation events register missing-item detection, stack/span relationships, and active-span attribution; markers define the measurement window.

The minimal workload is also genuinely mixed at two levels: both the cycle span and its nested inner span contain CPU work and async waits. This declares CPU `1:3`, wait `1:2`, whole-cycle CPU:wait `4:3`, and inner-subtree `3:2` relationships for later mixed-flamegraph validation.

noxware (approved):

PR #750, opened 2026-08-03
chore(metrique): drop version from tokio-telemetry dev-dep

jv1i:


dial9-metrique/Cargo.toml
```diff
@@ -25,7 +25,7 @@ tokio = ["dep:tokio"]
 [dev-dependencies]
 dial9-core = { workspace = true, features = ["test-util"] }
 dial9-metrique = { path = ".", features = ["tokio"] }
-dial9-tokio-telemetry = { version = "0.5.0-rc1", path = "../dial9-tokio-telemetry" }
+dial9-tokio-telemetry = { path = "../dial9-tokio-telemetry" }
```
rcoh, line 28: should this be workspace = true? I think if there is no version you can't publish it.

rcoh (approved):
lgtm but I think you need a version to have it be publishable

dial9-metrique/Cargo.toml
jv1i: not needed for dev-deps, but it made me realize it wasn't in the workspace deps at all, updated now. 
once we release and even out the versions we can just use workspace = true here

Issue #68, opened 2026-03-09
Display SegmentMetadata in trace viewer


rcoh:
The JS trace parser currently skips over SegmentMetadata events (wire code 11) without storing them. The viewer should parse and display segment metadata (service name, host, etc.) in the UI.

See the TODO in `trace_viewer/trace_parser.js` added in PR #66.

rcoh:
As part of this, can we also look into capturing segmentmetadata in the backend analysis as well?

PR #751, opened 2026-08-03
feature(viewer): display segment metadata table

noxware:
# Changes
Adds a clickable "Metadata" chip that popups a table with all the metadata.

# Related
#68


rcoh:
this button is next to two other identical buttons that are not clickable; there is no indication that it is clickable. Can you update the UI so its more obvious that you can interact with this?

Alternatively we have a "global table" on the left side. You could add a third column "metadata"?

noxware:
@rcoh the simplest change would be just re-styling the button a bit to make it look more like a button.

But adding this as a new tab in the left sidebar also makes sense. That sidebar has URL state, so you could share a URL with the metadata table visible, like `rail=metadata`.

Well, you could do it with the button popover as well by encoding a `popover=metadata`.

My only insecurity regarding the sidebar is that it contains things that are sortable, interactable with `p` / `n` hotkeys, and that control the "main view" viewport. Metadata would be the first unrelated tab.

Any preference?

rcoh:
lets move it to the table on a metadata column; p/n probably doesn't make that much sense _yet_ but you can imagine a future where you can click on a segment metadata and that will help you jump to other segments with the same metadata / different metadata (e.g. commit id)

noxware:
Made the change. But data there looks a bit "tight". Since the point of the change was to get better visibility of things, I'm letting it wrap for now. But I think it would look better if we don't wrap and use a tooltip even if we need to hide part of a key/value. wdyt?

rcoh:
yeah that looks pretty bad. Lets just ship your v0 commit but make it look clickable (maybe merge it with the `i` icon that shows other segment metadata?

rcoh:
there is a longer task to make that left-panel more usable. right now the tasks are way too narrow too, but that is a longer task to make it draggable etc. etc.

rcoh (approved):

noxware:
Missed one commit. Created https://github.com/dial9-rs/dial9/pull/761 for it.

PR #752, opened 2026-08-03
refactor: move tracing layer to dial9-utils

### LINKED ISSUE #725: move the TracingLayer out of `tokio_telemetry` and into `dial9-util` behind a flag
While we are there, we should switch it from tracking `worker_id` to (optionally) tracking https://docs.rs/tokio/latest/tokio/task/fn.try_id.html which is stable when the function is available.  This will require updating the correlation tooling for spans but the end result will be a major simplification.

Today, we use worker id to do a complex back-correlation to task id but by simply storing task id, we will simplify significantly

noxware:
# Changes

- Moves `tracing_layer` to `dial9-utils`. Nothing is left in `dial9-tokio-telemetry` or the `dial9` facade. No re-export or anything. These are kept as `dev-dependencies` for testing.
- Start sending `dial9.tokio.task_id`, instead of `worker_id`, for spans.
- Readers still recognize `worker_id` as a fallback for compatibility with old traces, but `dial9.tokio.task_id` is preferred.

# Note

The `worker_id` to `task_id` (now `dial9.tokio.task_id`) change that was suggested in the issue, was actually necessary to keep `dial9-utils` independent, or it would need `dial9-tokio-telemetry` for its TLS bookkeeping.

# Other

Closes #725


noxware:
i still need to verify some stuff in the viewer after the merge, but rust side is complete

dial9-tokio-telemetry/src/lib.rs
```diff
@@ -6,7 +6,6 @@
     unreachable_pub
 )]
 #![deny(unused_must_use, unsafe_op_in_unsafe_fn)]
-#![cfg_attr(docsrs, feature(doc_cfg))]
```
rcoh, line 9: I believe we still want this attribute in general to produce `doc_cfg` annotations in docs.rs

dial9-utils/src/tracing_layer.rs
rcoh, line 99: we can file this as a todo but we should probably persist _either_ a task_id or a thread id. A thread_id would let us track spans in the blocking pool (we can track them as is, we just don't have any way of seeing what thread they are on)

dial9-utils/README.md
```diff
@@ -3,21 +3,54 @@
...
 
+## Tracing span events (opt-in)
+
+Connect `Dial9TracingLayer` to a `tracing_subscriber` registry:
+
+```rust
+use dial9_utils::tracing_layer::Dial9TracingLayer;
+use tracing_subscriber::prelude::*;
+
+tracing_subscriber::registry()
```
rcoh, line 42: there is also a pattern where you set `dial9: true` on the spans you want. We should either demonstrate that or link to an example probably

rcoh (approved):
lgtm! couple of small inline comments. good to merge after you complete validations

dial9-utils/src/tracing_layer.rs
```diff
@@ -98,13 +101,13 @@ fn build_callsite_schemas(meta: &'static tracing::Metadata<'static>) -> Callsite
 
     // Base fields present on all span events
     let mut enter_fields = vec![
-        FieldDef::new("worker_id", FieldType::Varint),
+        FieldDef::new("task_id", FieldType::OptionalVarint),
```
rcoh: lets call this `dial9.tokio.task_id` to align with the metrique layer and reduce chance of conflict with user-defined fields

dial9-tokio-telemetry/src/lib.rs
noxware: Done. Commit `beab7cf`.

dial9-utils/README.md
noxware: Documented at `a03c6e5`.

dial9-utils/src/tracing_layer.rs
noxware: Addressed at `5675228`. Note:

This looks a little weird due to the surrounding fields not having the same convention:

```rust
const TOKIO_TASK_ID_FIELD: &str = "dial9.tokio.task_id";
FieldDef::new(TOKIO_TASK_ID_FIELD, FieldType::OptionalVarint),
FieldDef::new("span_id", FieldType::Varint),
FieldDef::new("parent_span_id", FieldType::OptionalVarint),
FieldDef::new("span_name", FieldType::PooledString),
```

Also, the previous `task_id` name used to exist in `trace_shape.rs`. And other events like `PollStartEvent` still name this `task_id`.

So, are you ok with the impact of the rename?

dial9-utils/src/tracing_layer.rs
noxware: You mean `task_id` and `thread_id` exchangeable? Or sending both if both are available? Makes sense to file it as an iterative improvement after merging the refactor.

noxware:
The same CI failures are happening in the unrelated https://github.com/dial9-rs/dial9/pull/751 PR after merging `main`.

https://github.com/dial9-rs/dial9/actions/runs/31060375868/attempts/1

Update: Passed after rerunning the failed jobs. It may be flaky.

dial9-utils/src/tracing_layer.rs
rcoh: they are not interchangable. If you are not running on Tokio, you won't have a task id.

noxware: Should we send `thread_id` always? Or only if there is no `task_id`?

noxware:
`dial9-tokio-telemetry` from where this comes, requires at least tokio `1.52`, while `dial9-utils` was requiring any `1.x` version.

`try_id` requires at least `1.41`, so we must ensure that as a minimum.

I specified `1.52` as the minimum in `dial9-utils` to match `dial9-tokio-telemetry`.

PR #754, opened 2026-08-04
feat: Properly Handle Multi-Runtime Events

### LINKED ISSUE #697: Flush emits aggregated queue depth across all runtimes instead of per-runtime samples
`TokioRuntimesSource::flush` sums `global_queue_depth()` and `active_tasks()` across all runtimes and emits a single `QueueSampleEvent`.
  This loses per-runtime granularity; consumers can't tell which runtime is backlogged.
  
  It should emit one `QueueSampleEvent` per runtime, tagged with runtime identity.

rcoh:
Fixes #697 

This changes both the supply and parse side to better support projects using multiple runtimes. Instead of a `QueueSampleEvent`, producers now produce `RuntimeMetricsEvent` that includes the Runtime name.

I tweaked the example app to have two runtimes and improved the viewers rendering of multi-runtimes; there is now a queue per runtime showing the global queue and active task quantity.


dial9-viewer/src/trace_shape.rs
```diff
@@ -129,6 +129,7 @@ const BUILTIN_SCHEMAS: &[&str] = &[
     "WorkerParkEvent",
     "WorkerUnparkEvent",
     "QueueSampleEvent",
+    "RuntimeMetricsEvent",
```
jv1i, line 132: I think we need to update `feature_enabled` fn in simulator to have this name too

jv1i (approved):
LGTM. I think the skill scripts need the same fallback you added in queue-model.ts. red_flag_scan and zoom still read just `spans.queueSamples`. perhaps they can share a helper for that

PR #758, opened 2026-08-06
trace-format: support borrowed event structs in #[derive(TraceEvent)]

rcoh:
## Summary

Enable code like:

```rust
#[derive(TraceEvent)] 
struct Request<'a> {
  ts: ...
  operation: &'a str,
}
```

This also cleans up the `'static` requirement which was always bothering me anyway.

## Test plan

- [x] New round-trip integration test: borrowed event encodes and decodes correctly
- [x] Byte-identical event frames test: borrowed and owned twins produce same wire bytes
- [x] Snapshot tests for borrowed/mixed/wire_slot+lifetime expansions
- [x] Error-case tests: two lifetimes rejected, type params rejected
- [x] All existing 172 tests pass unchanged
- [x] Full workspace compiles (downstream crates verified)
- [x] clippy + fmt clean


dial9-trace-format/src/encoder.rs
```diff
@@ -505,11 +505,8 @@ impl<W: Write> Encoder<W> {
     /// hashmap (registering the schema if needed) and populate the slot cache
     /// so the next call for the same type takes the fast path.
     #[cold]
-    fn resolve_dynamic_wire_id<T: TraceEvent + 'static>(
-        &mut self,
-        slot: usize,
-    ) -> io::Result<WireTypeId> {
-        let key = SchemaKey::RustType(TypeId::of::<T>());
+    fn resolve_dynamic_wire_id<T: TraceEvent>(&mut self, slot: usize) -> io::Result<WireTypeId> {
+        let key = SchemaKey::RustType(TypeId::of::<T::Id>());
```
fmzbl: testing this I found that a wrong `type Id` silently corrupts the trace on a hand-written
  `impl TraceEvent`. before, this key came from the compiler (`TypeId::of::<T>()`), so a hit was
  always the same type and needed no check. now `Id` is hand-written, so a hit can be a different
  type and nothing verifies it, the events get encoded under the other type's schema and their own
  schema never reaches the wire.

  maybe adding a 'debug_assert_eq' since it's cheap? or making it a compile error, though that may
  take more work.

rcoh: I think we can document that you need a unique type, are you saying if you impl TraceEvent and set a non-unique id?

fmzbl: yeah exactly, i said that cause getting it wrong corrupts silently, the events just get encoded under the other type's schema, but idk, maybe docs are enough.

rcoh:
> yeah exactly, i said that cause getting it wrong corrupts silently, the events just get encoded under the other type's schema, but idk, maybe docs are enough.

I'm not sure if this will cause other complications/be worth it but you can do

```
trait TraceEvent<Id=Self> { ... }
```

It may complicate other code but it makes it hard to misuse

rcoh:
Ah, this is a good option: https://docs.rs/typeid/latest/src/typeid/lib.rs.html#198-222

fmzbl:
oh cool, yea, i tested it and i think it would work

fmzbl (approved):

fmzbl:
leaving the approve

PR #760, opened 2026-08-06
feat(trace-format-derive): generics, field roles, strict attribute parsing

fmzbl:
Split out of #<span PR> so it can be reviewed on its own, per review feedback. Lands first;
  the span PR rebases onto it.

  Three changes to `#[derive(TraceEvent)]`:

  - **Generics/lifetimes** are carried onto the generated impl, so an event struct can have
    typed generic fields. Note the wire event name is per type, not per monomorphization: a
    generic event must have one instantiation per `name`, which is how the caller uses it
    (per-call-site generated structs with a `name = concat!(..., file!(), line!(), ...)`).
    Documented on the macro, plus a snapshot test.
  - **`role = "..."`** field attribute, emitting a `dial9.role` schema annotation. Vocabulary
    stays in `dial9_core::schema_extensions::roles`.
  - **Strict attribute parsing**: `parse_nested_meta` errors were being dropped with `let _ =`,
    so a malformed attribute was silently ignored. They now propagate, and unknown keys are
    rejected. All keys of one attribute are parsed in a single pass, since a pass that skips a
    key chokes on its unconsumed value.

  Also adds `TraceField for &str`, needed for borrowed fields on generic events.


dial9-trace-format-derive/src/lib.rs
```diff
@@ -170,9 +243,26 @@ fn derive_trace_event_impl(input: DeriveInput) -> Result<proc_macro2::TokenStrea
...
+    // typed, generically-parameterized fields).
+    //
+    // Caveat for type parameters: `event_name()` is per *type*, not per
+    // monomorphization, so all instantiations of a generic event share one wire
+    // schema. That is sound only when a name maps to a single set of field
+    // types, which is how `dial9_span!` uses it: each call site declares its
+    // own struct with a call-site-unique `name = ...`, so it is instantiated
+    // exactly once. Two instantiations with different field types under one
+    // name would register conflicting schemas for that name.
+    let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
```
rcoh: Why do we actually need type generics btw? I don't think I ever understood that

rcoh (commented):
Why do we need generic types? I'm really worried about allowing this because just because it's sound for tracing it's a soundness whole we are exposing in the API

dial9-trace-format-derive/src/lib.rs
fmzbl: i didnt think it would hurt, im integrating your new branch with mine and correcting that

dial9-trace-format-derive/src/lib.rs
```diff
@@ -113,21 +175,37 @@ fn derive_trace_event_impl(input: DeriveInput) -> Result<proc_macro2::TokenStrea
...
+            annotation_tokens.push(quote! {
+                ::dial9_trace_format::schema::FieldAnnotation::new(
+                    #idx,
+                    #ROLE_ANNOTATION_KEY,
+                    #role,
+                )
+            });
+        }
+        // `kind = "..."` is emitted as a "kind" schema annotation telling the
+        // viewer how to chart the field (gauge / counter / updown-counter).
```
jv1i, line 233: Should we have a SUPPORTED_ROLES list here and validate against it, like units and kinds? Otherwise (AFAICT) a typo in a role would go through `StructuralRole::from_annotation` as None and the schema goes turns into `NotSpan`

jv1i:
re: CI failure - I think we need a rebase + `cargo insta accept` (snapshot here is prior to #765)
while we're at it, should we add a positive snapshot for `name =  ...`?s

rcoh (approved):

PR #763, opened 2026-08-06
feat: build without tokio_unstable

### LINKED ISSUE #364: Support compiling w/o tokio_unstable
> we actually could (at a loss of fidelity) make it optional today. You wouldn't get poll information in your trace unless you spawned your future with our future wrapper (we would need a small bit of additional work to have the wrapped future emit poll start/stop events). You would still get CPU profiling, application events, and worker parks/unparks.
> 
> I'm not sure Tokio's disposition on stabilizing those callbacks. I'm not even sure if they should be stabilized or if we should pursue a more wholesale wrapper-future API.
> 
> I think "dial9 without tokio unstable" is something we absolutely should support 

 _Originally posted by @rcoh in [#363](https://github.com/dial9-rs/dial9-tokio-telemetry/issues/363#issuecomment-4390776518)_

jv1i:
closes #364.

Currently building dia9-tokio-telemetry without tokio_unstable is impossible due to a lot of compilation errors. This PR makes tokio instrumentation possible without unstable, with more narrow data.

### Changes

Tokio's four task-scoped hooks (`on_before_task_poll`, `on_after_task_poll`, `on_task_spawn`, `on_task_terminate`) and `TaskMeta` sit behind `#[cfg(tokio_unstable)]`. Without them the poll events come from `TracedFuture`, which only wraps futures going through `dial9::spawn`, `spawn_in`, `block_on` and `spawn_with`, so coverage is those tasks rather than every task on the runtime. 
Task spawn/terminate has no wrapper stand-in: the task id doesn't exist until first poll, so a spawned-but-never-polled task would go unrecorded, and `Drop` runs on threads the recorder doesn't own. The task list is built from polls either way, and the viewer already leaves lifetime blank when the events are absent.

`worker_index()` is unstable too, and it backs `current_worker_id()`, so park/unpark, wake events and tracing-layer spans lose worker identity without a fallback. Threads claim an id on first touch instead. `worker_ids` becomes a `BTreeSet<u64>`, the map key stopped meaning anything once there's no tokio slot and `metadata_entry` only ever read the values.

`local_queue` has no stable source, so it records 0. A `tokio.unstable` segment metadata key says which mode produced the trace, and the viewer reads it as two flags.

### Viewer

The viewer now reads which mode produced a trace, so missing data is never drawn as zero. It notes the reduced fidelity in the trace info menu, explains an empty Tasks tab instead of reporting "no tasks in this trace", and drops the per-worker series from the Queue depth track rather than drawing a flat zero line that would read as "the local queues are empty". The lane tooltip shows `-` for local queue for the same reason.

Reduced fidelity
Screenshot 2026-08-07 at 12 58 40 AM

Local queue showing `-`
Screenshot 2026-08-07 at 12 59 32 AM

`diagnose_setup.js` reports it as a finding, which should also prevent `no-task-polls` from blaming `Runtime::block_on` for a missing flag.

### CI

New CI job builds and tests without the flag.

### Why one id at a time

The fallback reserves worker ids singly rather than a `num_workers()` block. `block_in_place` hands a worker's core to a blocking-pool thread, which then runs the worker loop and fires the park hooks, so it reads as a new worker here and takes its own lane. A pre-sized block would run past its end and collide with the next runtime's ids.

### Tests

Some tests moved to `dial9::spawn` where they were about worker ids or decoding and raw `tokio::spawn` was incidental. Some existing tests are now gated to the flag, in three groups:

- assert on `TaskSpawnEvent` / `TaskTerminateEvent`, which now only tokio's hooks write
- need a poll census over every task on the runtime: `end_to_end_trace_matches_workload_and_metrics`, `attached_runtime_metadata_reaches_sealed_segment`, `overhead_bench_validates`
- register the task-meta hooks themselves, so the API they call is gone (`tokio_hooks.rs`)

### Misc. / follow ups

I considered adding a wrapper stand-in for task spawn/terminate as well (spawn dated on the spawning thread, terminate from the wrapper's `Drop`), but it didn't seem worth the extra field on every future, so I'm leaving that up for a follow up.


CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- dial9 now builds without `--cfg tokio_unstable` ([#364](https://github.com/dial9-rs/dial9/issues/364)).
+  With the flag nothing changes. Without it, poll events come from dial9's future wrapper instead of tokio's hooks, so they cover tasks spawned through `dial9::spawn`, `spawn_in`, `block_on` and `spawn_with` rather than every task on the runtime.
+  Task spawn/terminate events and per-worker queue depth have no stable source and are unavailable. 
+  Traces carry a `tokio.unstable` metadata key recording which mode produced them, and the viewer reports the reduced coverage.
```
rcoh: We probably now or in the future want this to be descriptive of what's in the trace rather than just unstable/stable 

For example we might eventually have fine grained control, sampled poll events, etc.

This could be additive metadata but worth thinking about

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
rcoh, line 99: Is this Todo done?

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -29,10 +30,11 @@ pub(crate) struct RuntimeContext {
     /// `Dial9Handle::current()`, the tracing layer, and `dial9::spawn` resolve.
     pub session_handle: Dial9Handle,
     /// Base worker ID for this runtime, reserved on the first worker resolve.
+    #[cfg(tokio_unstable)]
     pub worker_id_base: OnceLock<u64>,
-    /// Maps worker_index → global worker_id within this runtime.
+    /// Global worker IDs within this runtime.
     /// Populated lazily the first time each worker thread resolves its identity.
-    pub worker_ids: RwLock<HashMap<usize, u64>>,
+    pub worker_ids: RwLock<BTreeSet<u64>>,
```
rcoh: Why BTree instead of HashSet?

Side note RWLocks always worry me. RWLock is probably not actually useful here

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -70,16 +72,54 @@ thread_local! {
...
         f(slot.get_or_insert_with(|| tokio::runtime::Handle::current().metrics()))
     })
 }
 
-/// Local queue depth for `worker_index` on the current worker's runtime.
-fn local_queue_depth(worker_index: usize) -> usize {
-    worker_metrics(|m| m.worker_local_queue_depth(worker_index))
+/// This thread's global worker ID, claiming one if it has none yet.
+///
+/// Without `worker_index()` there is no slot to offset from, so IDs follow
```
rcoh, line 108: We actually used to have some super hacky code (before I added the new worker index function) to back calculate this.

This is probably fine too.

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -473,37 +528,68 @@ impl Encodable for TaskSpawn {
     }
 }
 
+/// Poll-start event for a task the `TracedFuture` wrapper is about to poll.
+///
+/// Stands in for tokio's `on_before_task_poll` hook when that is unavailable.
+/// Resolves the worker id through [`claim_thread_worker_id`], so a poll that
+/// lands before this thread's first park still gets a real lane.
+#[cfg(not(tokio_unstable))]
+pub(crate) fn make_wrapper_poll_start(
```
rcoh: I wonder if we should put this in TracedFuture to make it a hit clearer.

Or could we get RuntimeContext into dial9::spawn and unify these?

rcoh (commented):
Overall looks great

dial9-tokio-telemetry/src/telemetry/recorder/mod.rs
jv1i: Not as I was picturing it when I wrote that, I'll follow up on that after this PR

CHANGELOG.md
jv1i: I made that more individual now, while still keeping a `tokio.unstable` so that it still easy for disgnosis/agents to tell why those downgrades happen. We now have: `tokio.poll_coverage`, `tokio.local_queue` and `tokio.task_events`. We can improve how we report this in the UI later, but I agree on making this descriptive and additive

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
jv1i: BTreeSet because `metadata_entry` emits a sorted CSV and the old code was already sorted by hand, so ti just avoids that manual step, but no big deal. 

Good idea on the RWLock, there wasn't a reason for it I think, switched it to the `primitives::sync::Mutex`, should get all of this on shuttle later.

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
jv1i: good ideas! I had an actual issue here, without the flag a worker only joined its runtime on first park instead of first poll. 

what I did now is have both live in `enroll_thread`, and I merged the constructors, `make_poll_start` / `make_poll_end` take `Option<&RuntimeContext>`. 

didn't put it in `TracedFuture` because `spawn_in` builds the future before there's a runtime context, so a poll-time lookup is needed either way and the field would be on top of it rather than just instead. (now only a thread's first poll pays for the lookup)

dial9-tokio-telemetry/src/telemetry/recorder/recorder_tokio.rs
```diff
@@ -379,6 +383,8 @@ impl Dial9HandleTokioExt for Dial9Handle {
         // Publish attach state only after a successful build.
         // `TokioRuntimesSource` picks the runtime up from the registry on its
         // next flush.
+        #[cfg(not(tokio_unstable))]
```
rcoh: this is this `not(tokio_unstable)`?

In general it would be good to do the best to keep code paths uniform as much as possible and explain how the divergence happens and when

dial9-tokio-telemetry/src/telemetry/format.rs
```diff
@@ -196,6 +196,7 @@ pub struct TaskSpawnEvent {
 
 #[derive(TraceEvent)]
 #[traceevent(wire_slot)]
+#[cfg_attr(not(tokio_unstable), allow(dead_code))]
```
rcoh, line 201: weird/surprising that we have TaskTerminate with this flag but not `TaskSpawn`. worth explaining

dial9-tokio-telemetry/src/traced.rs
```diff
@@ -261,7 +300,52 @@ impl<F: Future> Future for WakeTraced<F> {
 
         let traced_waker = make_traced_waker(this.waker_data.clone());
         let mut traced_cx = Context::from_waker(&traced_waker);
-        this.inner.poll(&mut traced_cx)
+
+        #[cfg(tokio_unstable)]
```
rcoh: I might consider making this a configuration option you can pick for the future "emit poll events" vs. "don't emit poll events"

that will help avoid having differing code paths and it also means that you have a bit more control. 

A final option to consider is that in `on_poll_start` we set a TL basically saying "we are recording this already!" and look for that when deciding whether we should record our own events.

rcoh (approved):

rcoh (approved):

PR #768, opened 2026-08-07
feat: JoinSet extension trait

noxware:
# Changes
- Adds a `JoinSetExt` sealed extension trait that is implemented by Tokio `JoinSet`s, giving them two traced methods, `spawn_traced` and `spawn_traced_on`.
- Add a couple of very specific tests to ensure `#[track_caller]` propagates correctly through an external helper function.

# Note

I exposed `JoinSetExt` from `dial9_tokio_telemetry::telemetry::JoinSetExt` and from `dial9::JoinSetExt` but not from `dial9_tokio_telemetry::JoinSetExt` as it seems to expose very specific stuff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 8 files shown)


rcoh (approved):
lgtm!

PR #771, opened 2026-08-11
fix(viewer): preserve aggregate flamegraph inspection in URL

### LINKED ISSUE #740: fix(viewer): New viewer for flamegraphs does not preserve inspection in URL
The old viewer will add a URL parameter `inspect=` which makes it possible to deep link into a specific flamegraph view, however, the new flamegraph viewer lacks this behavior

noxware:
## Changes

Some degree of support for `inspect=` was already there. The missing part was in the `api=1` version, which this PR adds.

## Verification

You will need the `dev-server` to test this specific place.

From the root:

```shell
(cd dial9-viewer/ui && npm run build)
cargo run -p dial9-viewer --bin dev-server --features dev-server
```

Then go to `http://127.0.0.1:3000/new/flamegraph.html?api=1&bucket=demo-traces&credential_mode=ambient&prefix=traces` (notice the `api=1`).

There, right click a frame and select `Inspect frame`. It should open the butterfly (split) view. There should be an `&inspect=` in the URL, and if you navigate again to that URL, you should be in the same place. 

## Attention

It should be possible to test this without manually writing the URL, but there are a couple of issues preventing this. Look at the key hardcoded in the dev-server:

```
"traces/2026-04-09/1900/demo-service/local/host-0/abcd/1744224000-0.bin.gz"
```

The date in the second part, and the timestamp in the filename do not match. This is easy to fix, but would not make this testable yet.

The other issue is that `local/host-0`, in the URL, is not properly parsed. Removing `local/`, together with fixing the timestamp would make this testable but, `local/` is actually a valid key, as there are tests using `us-east-1/i-0abc123` as their instance path. And fixing the parser would require a lot more changes. 

So, I will leave things as they are here, and continue investigating the proper fix in other issue/PR.

## Other

Closes #740


noxware:
There are a couple of not so easy to fix issues related to this one that I have to split out.

jv1i (approved):

noxware:
Created https://github.com/dial9-rs/dial9/issues/788 with some of the findings mentioned in `Attention`. But not the `dev-server` timestamp mismatch.

PR #772, opened 2026-08-11
refactor(viewer): promote the Vite UI and remove the legacy viewer

### LINKED ISSUE #769: Remove previous viewer and make new one the default

noxware:
## Changes

- Removes the old viewer.
- Promotes the new viewer as the canonical one.
- Update comments across the codebase referencing the old viewer.
- Port or remove tests no longer relevant for the new viewer.
- Keeps some very tiny stubs for `/new` urls, in case someone have them bookmarked. 

## Other

Closes #769


jv1i (approved):
Looks good!
Seems like journeys and readout-schema still use the pre-promotion ids (#btn-next-poi, #lanes-container, #poi-counter, #sort-by-worst), and AFAICT they no longer exist, so I think J1-J4/J7/J8 and the viewer readout schema go dead here. 

Should we retarget those now or in a follow-up?
Similarly behavior-diff and some of things under `/parity` won't be needed anymore.
for /parity, the a11y/contract/inventory checks in might be worth keeping, but maybe under another name?

Either way this PR looks good, I'm fine with following up on those as long as we don't lose track of them.

noxware:
Good catch 👍  I will make a separate PR with that, just to move this to done and make further review steps easier and focused

Issue #764, opened 2026-08-07
Improve shuttle concurrency-test coverage


tiny-ac:
Improve Shuttle coverage for concurrent components that currently lack it.
- Make sure shuttle tests actually run as part of the CI check
- Ensure coverage reports are properly generated
- Look for alternatives to primitives that shuttle doesn't cover (e.g. `tokio::sync::Notify`, `tokio::sync::mpsc`).
  - Defer to a follow-up/document gaps if a fix turns out to be too risky or complex for this pass

If any real concurrency bugs turn up while adding this coverage, file them as separate issues rather than fixing them inline here.

Follow up PRs
- [x] test/shuttle-shared-state-coverage (https://github.com/dial9-rs/dial9/pull/851)
- [x] test/shuttle-worker-time-coverage (https://github.com/dial9-rs/dial9/pull/857)
- [ ] test/shuttle-runtime-context-attach (https://github.com/dial9-rs/dial9/pull/873)
- [ ] test/shuttle-dump-trigger-debounce (https://github.com/dial9-rs/dial9/pull/877)
- [ ] test/shuttle-tracing-layer-cache-coverage (https://github.com/dial9-rs/dial9/pull/878)
- [ ] test/recording-teardown-panic-gap (https://github.com/dial9-rs/dial9/pull/897)
- [ ] ci/shuttle-zero-tests-guard 
  - push last, once others are merged

PR #781, opened 2026-08-12
test: improve shuttle concurrency test coverage

tiny-ac:
#764 

- Adds Shuttle coverage to dial9-core (macro adoption across pre-existing scenarios, plus new coverage for DiskFs claim-dedup and Source::flush panic injection)
  - New shuttle_test! macro (dial9-core/src/primitives.rs) pairs check_pct and check_uncontrolled_nondeterminism
    - supports should_panic/expected=/replay=/determinism_only/default budget shorthand (verify_faults_triggered is also supported by the macro but unused by anything in this PR)
- `test-shuttle.sh` now fails loudly if a crate's shuttle filter matches 0 tests instead of silently reporting success
- Mutation-tested new tests
- Filed issues
  - https://github.com/dial9-rs/dial9/issues/782
  - https://github.com/dial9-rs/dial9/issues/783


dial9-core/src/primitives.rs
```diff
@@ -130,6 +130,242 @@ macro_rules! define_thread_local {
...
+///     fn my_scenario() { /* ... */ }
+/// }
+/// ```
+///
+/// Add `, should_panic` after `depth = $depth` to document a known bug
+/// instead of asserting correctness. Add `, expected = "..."` after
+/// `should_panic` (or `flaky_sigabrt_determinism_only`) to pin the specific
+/// panic message, like `#[should_panic(expected = "...")]` -- otherwise the
+/// test passes on any panic, including an unrelated one. Add `, replay =
+/// "<schedule>"` to also check in a `replay_known_failure` test that pins
```
tiny-ac: - `pct`/`determinism` pick a different random schedule every run, so a known bug's exact reproduction isn't pinned down anywhere
- `replay = "<schedule>"` checks in one captured failing schedule as its own `replay_known_failure` test, so that specific failure reproduces every run instead of waiting on `pct` to stumble into it again
- Doesn't replace the random search, `pct`/`determinism` keep running as before

https://github.com/dial9-rs/dial9/issues/782#issuecomment-5296746729

dial9-core/src/pipeline_shuttle_tests.rs
```diff
@@ -120,101 +120,227 @@ impl Source for MockSource {
...
+    }
 }
 
-#[test]
-fn pct_real_pipeline() {
-    shuttle::check_pct(test_core_pipeline, 10000, 3);
+// Companion to the scenario above: a TL-buffer write hits the same
+// unguarded-flush gap. Kept separate because the live buffer left behind at
+// panic time is what makes `determinism` crash (16/30 with the write vs
+// 0/30 without) -- hence `flaky_sigabrt_determinism_only`.
```
tiny-ac, line 289: The SIGABRT here isn't a real production crash risk. It's internal to shuttle 0.9.1

`check_uncontrolled_nondeterminism` runs a schedule twice (record, then replay), and if a task's still holding a `ThreadLocalBuffer` when the first panic tears down shuttle's own `EXECUTION_STATE`, that buffer's `Drop` panics again mid-unwind and Rust aborts.

`check_pct` doesn't do the double run, so it never hits this.

A standalone repro outside shuttle of the same panic shape exits clean (code 0)

```rust
use std::sync::{Arc, Mutex};
use std::thread;

 fn main() {
    let sources: Arc<Mutex<Vec<i32>>> = Arc::new(Mutex::new(vec![1, 2, 3]));
    let sources2 = sources.clone();
    let handle = thread::spawn(move || {
        let mut g = sources2.lock().unwrap();
        for _ in g.iter_mut() {
            panic!("simulated Source::flush panic");
        }
    });
    let join_result = handle.join();
    println!("join() returned: is_err={}", join_result.is_err());
    println!("main thread still alive, proving no process abort occurred");
    let lock_result = sources.lock();
    match lock_result {
        Ok(_) => println!("mutex NOT poisoned (unexpected)"),
        Err(_) => println!("mutex IS poisoned, as expected -- but process is still alive"),
    }
}
```

 Output:

 ```
 thread '<unnamed>' panicked at real_panic_probe.rs:19:13:
 simulated Source::flush panic
 join() returned: is_err=true
 main thread still alive, proving no process abort occurred
 mutex IS poisoned, as expected -- but process is still alive
 ```

dial9-core/src/collector.rs
```diff
@@ -133,3 +133,111 @@ mod tests {
...
+            );
+        }
+    }
+
+    // Deterministic: eviction order is a data-structure property, not a
+    // race (the scenario above only checks aggregate counts, not order).
+    // Still needs shuttle's harness since `BoundedQueue` uses `shuttle::sync::Mutex`.
+    crate::shuttle_test! {
+        default, determinism_only;
+        fn shuttle_collector_eviction_is_fifo() {
```
tiny-ac: Structures being documented FIFO. Swapped `pop_front()` for `pop_back()` in eviction paths (evicts newest instead of oldest — LIFO):

```rust
let evicted = if q.len() >= self.capacity {
    q.pop_back()   // was q.pop_front()
} else {
```

Every existing accounting test (`shuttle_eviction_accounting`, `shuttle_collector_eviction_accounting`, `shuttle_handoff_no_loss`) stayed green under LIFO eviction.

Added tests to catch this kind of issue.

dial9-core/src/worker/shuttle_tests.rs
```diff
@@ -0,0 +1,220 @@
...
+        .collect()
+}
+
+crate::shuttle_test! {
+    num_iters = 2_000, depth = 3;
+    // Multiple writer threads seal segments into an in-memory `Fs`
+    // concurrently with a worker thread draining+processing them (no real
+    // Tokio runtime — see module doc). Every sealed segment must reach the
+    // processor exactly once.
+    fn shuttle_worker_handoff() {
```
tiny-ac: Races multiple writer threads sealing segments against a worker draining them via `process_open_segments()` directly (bypassing `run_continuous()`'s `tokio::select!`/`sleep`, which shuttle can't model). Checks exactly-once handoff under real interleaving instead of sampling it.

dial9-core/src/worker/shuttle_tests.rs
```diff
@@ -0,0 +1,220 @@
...
+    // `run_triggered`'s outer `tokio::select!`/`sleep_until` loop (shuttle
+    // has no live Tokio reactor). Multiple writers seal segments concurrently
+    // with one on-demand dump request; it must resolve exactly once
+    // regardless of whether any segments landed in its window first.
+    //
+    // `dump_current_data_at_for_test`/`set_epoch_secs_for_test` pin the
+    // clock values instead of reading real time (required for determinism
+    // replay), so this verifies the resolve-exactly-once race, not real
+    // clock-reading.
+    fn shuttle_dump_resolves_exactly_once() {
```
tiny-ac: Dropped the response send in `resolve_dump` to test coverage
```rust
// let _ = tx.send(result);   // removed
```

dial9-core/src/shared_state.rs
```diff
@@ -532,3 +532,92 @@ mod tests {
...
+    }
+
+    crate::shuttle_test! {
+        default;
+        // Shuttle counterpart of `concurrent_record_and_drain_preserves_event_count`:
+        // writer threads record events while a drainer thread concurrently bumps
+        // the drain epoch and intrusively drains, exploring interleavings
+        // exhaustively instead of sampling them. Every recorded event must reach
+        // the collector exactly once, regardless of schedule.
+        fn shuttle_concurrent_record_and_drain() {
```
tiny-ac: Core intrusive-drain invariant (`drain_all_tl_buffers` racing a writer's self-flush), previously only proptest-sampled. Kept the proptest alongside.

Injected bug: removed `self.collector.accept_flush(batch)` inside `drain_all_tl_buffers`. A batch gets flushed and counted, but never reaches the collector. Both `pct`/`determinism` failed immediately.

`test_core_pipeline` stayed green against the same injection. Its writer threads always finish and `join()` before the drain, so `ThreadLocalBuffer`'s own `Drop` self-flushes first.

 `drain_epoch`/`flush_epoch` use `Ordering::Relaxed` throughout, and shuttle only correctly models `SeqCst`. Keeping prop test because of that.

dial9-core/src/pipeline_shuttle_tests.rs
```diff
@@ -329,64 +455,57 @@ fn run_erroring_pipeline(fault: fs::FaultPolicy) -> u64 {
...
-        "rate limiting failed under probabilistic fs faults: observed {total} \
-         WARN/ERROR events, expected <= 10."
-    );
+// The fault is armed on the test thread but read on a spawned thread, this
+// proves it crosses that boundary. No `pct`: the spawning thread parks on
+// `.join()` immediately with no other work in between, so the two threads
+// are never simultaneously runnable.
+crate::shuttle_test! {
+    default, determinism_only;
+    fn fs_fault_visible_across_threads() {
```
tiny-ac, line 459: `primitives::fs::FAULT` is backed by real `std::thread_local!`, not the shuttle-swapped macro, so a fault armed on the test thread stays visible to a separately-spawned thread under shuttle.

Injected the opposite (swapped `FAULT` to the shuttle-aware macro): failed immediately, `"fault armed on the test thread was not observed on a spawned thread"`.

tiny-ac:
I was having issues with the Benchmark check so I temporarily forced the action to install valgrind instead of getting it from the cache. That seems to have fixed it

https://github.com/dial9-rs/dial9/actions/runs/32092487633/job/95577338772

dial9-core/src/fs/mod.rs
```diff
@@ -455,6 +455,17 @@ impl Fs {
             Fs::Disk(_) => panic!("set_seal_secs_for_test is memory-only"),
         }
     }
+
+    /// Test-only: override the creation epoch of a queued memory slot --
```
rcoh: I'd really like to avoid this sort of thing if possible. the best way to test these sorts of things is by having a mock-clock that is used.

dial9-core/src/worker/shuttle_tests.rs
```diff
@@ -0,0 +1,208 @@
...
+//!
+//! `shuttle_worker_handoff` drives the real `run()`/`run_continuous()` path
+//! (not the test-only `process_open_segments()` shortcut): this works for
+//! the in-memory `Fs` backend because `wait_for_more`'s in-memory branch
+//! only awaits `Fs::wait_for_wakeup()` (a plain `tokio::sync::Notify`, no
+//! reactor/timer needed) rather than `tokio::time::sleep` -- that real timer
+//! is only on the disk-backend branch, which this module never uses.
+//!
+//! `shuttle_dump_resolves_exactly_once` still can't drive `run_triggered()`
+//! directly: its outer loop selects on `tokio::time::sleep_until`, a real
```
rcoh: we should use https://docs.rs/shuttle-tokio-impl-inner/latest/shuttle_tokio_impl_inner/time/fn.sleep_until.html

dial9-core/src/worker/shuttle_tests.rs
rcoh: we should really try to avoid duplicating logic into tests and focus on how we can make the code testable (or by using shuttle-tokio) instead.

rcoh (commented):
I like all these focused tests. One thing I'd like to see eventually is a way to run as close to possible as _full_ dial9 under shuttle. It seems like it should be close to being possible.

this PR is already pretty big. Lets pick a couple of focused things we want in this PR to kick things off; maybe the macro and one or two shuttle tests? Then iterate from there.

I'd really like to avoid having any application logic in the tests. its very easy for it to get out of sync. Instead we should be thinking about how we can make the code easier to test.

dial9-core/src/collector.rs
rcoh: I don't totally understand what this comment is saying. What `pop_front` are you referring to in this comment?

dial9-core/src/pipeline_shuttle_tests.rs
```diff
@@ -120,101 +120,222 @@ impl Source for MockSource {
...
-#[test]
-fn determinism_check() {
-    shuttle::check_uncontrolled_nondeterminism(test_core_pipeline, 10000);
+// ── Panic injection ─────────────────────────────────────────────────
+
+// `flush_sources` has no `catch_unwind`, so the panic propagates out of the
+// flush thread's main loop and kills it.
+crate::shuttle_test! {
+    num_iters = 500, depth = 3, should_panic,
+    expected = "PanickingSource intentionally panics for shuttle coverage",
```
rcoh: can you change this to `expect_panic` or another name that clarifies that this is testing for failure?

dial9-core/src/pipeline_shuttle_tests.rs
rcoh: ah interesting. is there a way we could do cleaner teardown? perhaps not.

dial9-core/src/pipeline_shuttle_tests.rs
rcoh: is there a reason this PR doesn't fix `primitives::fs::FAULT`?

rcoh (commented):
nice!

dial9-core/src/pipeline_shuttle_tests.rs
tiny-ac: @rcoh 
This test just pins down the choice that was made here https://github.com/dial9-rs/dial9/blob/c771b2aab2e276212f68494ce1ad2a219530a0c4/dial9-core/src/primitives.rs#L202-L204

Using the crate's shuttle-aware `thread_local!` instead would silently stop fault injection from reaching any spawned thread. This test was added to catch that. Swapping `std::thread_local! -> shuttle::thread_local!` causes this test to fail

```
  // swap: std::thread_local! -> shuttle::thread_local!
  shuttle::thread_local! {
      static FAULT: Cell<FaultPolicy> = const { Cell::new(FaultPolicy::None) };
  }
```

Image

dial9-core/src/pipeline_shuttle_tests.rs
tiny-ac: @rcoh 
I don't think there's a clean workaround on our side. It'd need to be patched [here (Shuttle)](https://github.com/awslabs/shuttle/blob/91e3cae8f73c2be3934960c46f8494a7ca981867/shuttle-engine/src/runtime/execution.rs#L162-L202)  the [state cleanup](https://github.com/awslabs/shuttle/blob/91e3cae8f73c2be3934960c46f8494a7ca981867/shuttle-engine/src/runtime/execution.rs#L201) doesn't run on panicking branches.

Running this test multiple times gives different results, that's why it's ignored
> RUSTFLAGS="--cfg shuttle" cargo test -p dial9-core --features _shuttle --lib -- test_source_panic_does_not_lose_tl_buffer_write --ignored

- We get this error:
> `ExecutionState::with` panicked because `ExecutionState` is not set. Are you accessing a Shuttle primitive outside of a Shuttle test?

https://github.com/awslabs/shuttle/blob/91e3cae8f73c2be3934960c46f8494a7ca981867/shuttle-engine/src/runtime/execution.rs#L438
Image

So the [Drop here](https://github.com/dial9-rs/dial9/blob/c771b2aab2e276212f68494ce1ad2a219530a0c4/dial9-core/src/encoder.rs#L289) executes with EXECUTION_STATE already unset and panics

- We get a SIGABRT
Image

- Same test, no changes, successful run
Image

dial9-core/src/pipeline_shuttle_tests.rs
rcoh: ah understood.

rcoh (approved):

PR #784, opened 2026-08-13
refactor(tokio): record events through Dial9Handle

jv1i:
Closes a `SharedState` TODO in `register_hooks`.

Every tokio hook closure cloned an `Arc<SharedState>` and built its event by hand. `SharedState` is dial9-core's internals, and the tokio layer should record through the handle it already holds.

### Changes

Recording moves onto `RuntimeContext` as `record_worker_park`, `record_poll_start` and friends, so each hook body is one call. The six `Arc<SharedState>` clones go with them.

`Dial9Handle` gains `record_event_with` which takes in a closure.

Renamed `RuntimeContext::session_handle` to `recorder_handle` (leftover from when we had `SessionHandle` type)

### DIFF
--- a/dial9-core/src/handle.rs
+++ b/dial9-core/src/handle.rs
@@ -223,6 +223,20 @@ impl Dial9Handle {
         }
     }
 
+    /// Record an event that is only built when recording is on.
+    ///
+    /// Reach for this over [`record_event`](Self::record_event) when building
+    /// the event costs something you would rather not pay while recording is
+    /// paused, such as a clock read or a lookup. `make` runs only if the event
+    /// will be recorded.
+    pub fn record_event_with<E: Encodable>(&self, make: impl FnOnce() -> E) {
+        if let Some(inner) = &self.inner {
+            inner
+                .shared
+                .if_enabled(|buf| buf.record_encodable_event(&make()));
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


dial9-core/src/handle.rs
```diff
@@ -223,6 +223,20 @@ impl Dial9Handle {
         }
     }
 
+    /// Record an event that is only built when recording is on.
+    ///
+    /// Reach for this over [`record_event`](Self::record_event) when building
+    /// the event costs something you would rather not pay while recording is
+    /// paused, such as a clock read or a lookup. `make` runs only if the event
+    /// will be recorded.
+    pub fn record_event_with<E: Encodable>(&self, make: impl FnOnce() -> E) {
```
rcoh, line 232: I would probably prefer record_if_enabled

rcoh (commented):
Nice this is great

rcoh (approved):

dial9-core/src/handle.rs
jv1i: Went with record_event_with because record_event is also gated by enabled, so record_if_enabled would imply the sibling isn't. Merging as it is now to unlock the rest but happy to rename before release if we have another option. We could have a record_event ungated, then record_if_enabled and record_if_enabled_with but I don't know if there's a need for ungated usage at this level.

PR #785, opened 2026-08-13
refactor(tokio): drop SharedState from the future wrappers

jv1i:
Depends on #784, similar TODO, for the other two holders.

`TracedHandle` was `struct TracedHandle { shared: Arc<SharedState> }` a newtype over the thing we are removing, so it collapses into `Dial9Handle` outright. The `Option` still carries its meaning: `None` is "not connected to a recorder, spawn plainly". `TaskDumped` holds a handle for the same reason. Both record through `record_event_with`.

`current_runtime_ctx` takes a `&Dial9Handle` and resolves the state itself.

### emit

`TaskDumped::emit` recorded every callchain under a single `if_enabled`. `record_event_with` records one event, so the gate is per event now. The poll fast path already returns early when recording is paused, so trimming still only happens while recording, and the per-event gate keeps `disable` an exact boundary.

### DIFF
--- a/dial9-core/src/shared_state.rs
+++ b/dial9-core/src/shared_state.rs
@@ -162,11 +162,11 @@ impl SharedState {
 
     /// Check whether recording is currently enabled.
     ///
-    /// Prefer [`Dial9Handle::record_event_with`](crate::handle::Dial9Handle::record_event_with)
-    /// for event-recording paths — it builds the event only when it will be
-    /// recorded, so a paused recorder pays nothing. Use `is_enabled()` only for
-    /// control-flow decisions that don't directly record events (e.g.
-    /// deciding whether to wrap a waker in wake-tracking polls).
+    /// For recording paths prefer
+    /// [`Dial9Handle::record_event_with`](crate::handle::Dial9Handle::record_event_with),
+    /// which builds the event inside the check. Use `is_enabled()` for
+    /// control-flow decisions that don't record, such as whether to wrap a
+    /// waker in wake-tracking polls.
     pub fn is_enabled(&self) -> bool {
         self.state.load(Ordering::Relaxed) == State::Enabled as u8
     }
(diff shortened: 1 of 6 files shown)


dial9-core/src/shared_state.rs
```diff
@@ -162,9 +162,9 @@ impl SharedState {
 
     /// Check whether recording is currently enabled.
     ///
-    /// Prefer [`if_enabled`](Self::if_enabled) for event-recording paths — it
-    /// provides an [`EventBuffer`] that makes it structurally impossible to
-    /// record without checking first. Use `is_enabled()` only for
+    /// Prefer [`Dial9Handle::record_event_with`](crate::handle::Dial9Handle::record_event_with)
+    /// for event-recording paths — it builds the event only when it will be
+    /// recorded, so a paused recorder pays nothing. Use `is_enabled()` only for
```
rcoh: Should eventually do a pass through all the public docs to deslop them e.g. "pays nothing"

dial9-core/src/shared_state.rs
```diff
@@ -162,9 +162,9 @@ impl SharedState {
...
     /// control-flow decisions that don't directly record events (e.g.
     /// deciding whether to wrap a waker in wake-tracking polls).
     pub fn is_enabled(&self) -> bool {
```
rcoh, line 170: This public function should also be made private?

rcoh (approved):

dial9-core/src/shared_state.rs
jv1i: good idea, removed that now and I'll do a general pass for the release.

dial9-core/src/shared_state.rs
jv1i: yup, #787 (next PR) is making it private

PR #787, opened 2026-08-13
refactor(core): seal SharedState behind test-util

jv1i:
Last of the series. #784 moves recording onto the handle, #786 did the future wrappers, this one closes out the remaining reach-throughs and takes the type off the public API.

Three things still went through `SharedState`: the source registry, worker-id reservation, and a connectivity check. Each gets a `Dial9Handle` method.

### Source registry

`runtime_registry` and `register_runtime_metrics` both open-coded the same thing, find my source of type `T` (or insert one), via `with_sources_vec`/`with_sources_mut` plus a manual `downcast_mut`. `Dial9Handle::with_source` and `with_source_or_insert` do it once.

`push_source` gets a handle method too. It was already advertised as an extension point in `Dial9Handle::shared`'s own doc, so it needed a real home regardless. The memory profiler was the one production caller.

### Worker IDs

`Dial9Handle::reserve_worker_ids`. `claim_worker_id` checks the handle up front so the `OnceLock` init stays infallible, making the init fallible instead would let two first-resolving threads each reserve a block and leak the loser's.

### is_connected

Three call sites wanted "connected to a recorder, regardless of pause", spelled `shared().is_some()`. `is_enabled`'s docs pointed at that spelling and the memory profiler's `install` doc already described its argument as a "disconnected handle", so the concept was there without a name.

### Visibility

`shared_state`, `SharedState`, `Dial9Handle::shared` and `Recorder::shared` move behind `test_util_pub!`, the macro already in `lib.rs` for this. They stay `pub` under `test-util` and go `pub(crate)` otherwise. Both `#[doc(hidden)]` markers are deleted.

### DIFF
--- a/dial9-core/src/handle.rs
+++ b/dial9-core/src/handle.rs
@@ -1,8 +1,17 @@
 use crate::encoder::{Encodable, ThreadLocalEncoder};
 use crate::primitives::sync::Arc;
 use crate::shared_state::SharedState;
+use crate::source::Source;
 use crate::thread::ThreadTrackingGuard;
+use std::any::Any;
 use std::cell::RefCell;
+
+/// First registered source of type `T`, if any.
+fn find_source<T: Source>(sources: &mut [Box<dyn Source>]) -> Option<&mut T> {
+    sources
+        .iter_mut()
+        .find_map(|source| (&mut **source as &mut dyn Any).downcast_mut::<T>())
+}
 
 crate::primitives::thread_local! {
     /// Per-thread [`Dial9Handle`], populated via [`set_tl_handle`] and cleared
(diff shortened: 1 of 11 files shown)


dial9-core/src/handle.rs
```diff
@@ -210,6 +219,70 @@ impl Dial9Handle {
...
+    pub fn is_connected(&self) -> bool {
+        self.inner.is_some()
+    }
+
+    /// Reserve `count` consecutive global worker IDs.
+    ///
+    /// IDs are unique across every runtime attached to this recorder, so each
+    /// runtime can hand its workers a block of its own. `None` on a disabled
+    /// handle, which has no counter to draw from.
+    pub fn reserve_worker_ids(&self, count: u64) -> Option<u64> {
```
rcoh: It feels like this isn't the right solution. If you _are_ tokio (and the tokio source) maybe that source should have its own global? 

I think just pushing tokio specific methods into the public surface is not really the right solution.

(Similarly worker ids probably shouldn't be in shared state anyway)

Clearly multiple TokioSource's need to coordinate. They should either do that via their own global or a more general purpose API for reading off metadata

jv1i: Agreed on worker IDs, shouldn't have placed that on core at all actually. 
I moved the counter onto TokioRuntimesSource, dropped reserve_worker_ids from the handle, and next_worker_id out of SharedState.

On coordination: there's only one TokioRuntimesSource per recorder (with_source_or_insert is find-or-insert), so it's just runtimes coordinating through one source. I went for typed source access (instead of a global, mostly bc it seemed simpler for teardown) with_source / with_source_or_insert so that crates can reach the source they registered, "reach your own source" which is probably useful for others.

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -115,13 +117,12 @@ fn worker_metrics<R>(f: impl FnOnce(&RuntimeMetrics) -> R) -> R {
...
 #[cfg(not(tokio_unstable))]
 fn claim_thread_worker_id(ctx: &RuntimeContext) -> Option<u64> {
-    let shared = ctx.recorder_handle.shared()?;
     CLAIMED_WORKER_IDS.with(|claimed| {
         let mut claimed = claimed.borrow_mut();
         if let Some((_, id)) = claimed.iter().find(|(owner, _)| *owner == ctx.id) {
             return Some(*id);
         }
-        let id = shared.reserve_worker_ids(1);
+        let id = ctx.worker_id_counter.fetch_add(1, Ordering::Relaxed);
```
rcoh: I'd probably prefer to keep this as a function on `ctx`

rcoh (approved):

PR #797, opened 2026-08-17
feat(s3)!: adopt Hive-style S3 key layout

### LINKED ISSUE #788: Viewer aggregate queries misparse slash-containing instance paths
## The root issue

When uploading to S3, the default key looks like:

```
{prefix}/{YYYY-MM-DD}/{HHMM}/{service_name}/{instance_path}/{boot_id}/{epoch_secs}-{index}.bin.gz
```

However:

- The uploader supports `instance_path` values containing `/`. Its tests include an instance such as `us-east-1/i-0abc123`.
- Legacy traces may lack `boot_id`.

This makes the boundary between `instance_path` and `boot_id` ambiguous when parsing a key by path-component count alone.

For example:

```
.../service/us-east-1/i-0abc123/file.bin.gz
```

could be a legacy trace with:

- `instance_path = us-east-1/i-0abc123`
- no `boot_id`

but the current UI parser interprets it as:

- `instance_path = us-east-1`
- `boot_id = i-0abc123`

Similarly, with the current layout:

```
.../service/us-east-1/i-0abc123/boot-id/file.bin.gz
```

the viewer cannot recover the complete instance path using fixed component positions.

## Impact in the viewer

The Browser and `api=1` currently assume that the instance path occupies a single key component.

As a result, an instance such as `us-east-1/i-0abc123` is exposed and filtered as only `us-east-1`. The remaining component is either misinterpreted as `boot_id` or left outside the parsed host value, depending on which viewer parser is used.

This prevents the full instance path from being discovered, selected, and filtered correctly.

### Related service-name mismatch

The uploader does not validate `service_name`, so it can technically contain `/`, even though the viewer rejects slash-containing service filters.

The Browser lists S3 objects using `.../{service}/` as a prefix. As a result, querying `service=s1` also returns keys produced with `service_name=s1/s2`, while `s1/s2` cannot be queried directly.

This is not the main focus of this issue. Unlike `service_name`, slash-containing `instance_path` values are explicitly supported and covered by uploader tests.

## Mitigation

Without changing the S3 key format, the viewer can mitigate this for the default and legacy layouts by matching the complete expected instance path instead of extracting it from a fixed position.

For example, given:

- `service = service`
- `instance = us-east-1/i-0abc123`

after consuming the known service and instance prefixes, the remaining path should have one of these forms:

- `{filename}`: legacy trace without `boot_id`
- `{boot_id}/{filename}`: current trace layout
- Any other shape: not one of the supported layouts, so discard it

Like service matching, this is still prefix-based and can produce false positives. For example, querying `instance=a` can accept a legacy key for `instance=a/b` by interpreting `b` as `boot_id`.

> **Note:** Prefix matching should add `/` at the end to avoid matching `s10` with `s1`.

> **Alternative:** Stop supporting the legacy layout without `boot_id` in these viewer features. The parser could then take the last segment as the filename, the previous one as `boot_id`, and everything between the service name and `boot_id` as `instance_path`. The downside is that legacy keys cannot always be detected and rejected. Their last instance segment may instead be silently interpreted as `boot_id`. This also assumes `boot_id` never contains `/`, which is true for generated IDs but is not validated by the uploader.

### LINKED ISSUE #789: Change default S3 key layout to be `k=v` (hive style)
> I think we should change to hive-style keys in 0.5: `s3://.../date=2026-08-14/service=....` This will make them much easier to deal with and make them easier to evolve going forward. 

 _Originally posted by @rcoh in [#788](https://github.com/dial9-rs/dial9/issues/788#issuecomment-5296701803)_

The current key layout is inflexible and requires positional knowledge.

noxware:
## Summary

Changes the default S3 trace segment key layout to:

```text
{prefix}/version=1/date={YYYY-MM-DD}/service={service}/time={HHMM}/instance={instance}/boot={boot_id}/{epoch}-{index}.bin[.gz]
```

For example:

```text
traces/version=1/date=2026-08-20/service=payments%2Fapi/time=1937/instance=cluster%2Fworker-7/boot=abcd-42/1786736220-3.bin.gz
```

Partition values use Hive-compatible escaping, removing ambiguity when values contain `/`, `%`, or `=`.

The viewer reads both historical and versioned keys without permanently doubling S3 LIST requests.

## Review guide

The commits follow the migration order:

1. **`feat(viewer): read versioned segment object keys`**
   Adds parsing and historical compatibility.

2. **`feat(viewer): discover mixed S3 key layouts`**
   Updates Browser listing, flamegraphs, and Tokio Stats.

3. **`feat(s3): write versioned segment object keys`**
   Switches the uploader and updates fixtures.

4. **`docs(s3): describe the versioned key layout`**
   Documentation only.
   
 ## Other
   
Closes #788
Closes #789


dial9-core/src/segment_object_key.rs
```diff
@@ -0,0 +1,382 @@
+//! Dial9 segment object key layout and Hive path escaping.
```
rcoh: generally stuff related to S3 and parsing should not be in core

dial9-destinations-s3/src/s3.rs
```diff
@@ -88,9 +89,12 @@ where
 /// # Default key layout
 ///
 /// ```text
-/// {prefix}/{YYYY-MM-DD}/{HHMM}/{service_name}/{instance_path}/{boot_id}/{epoch_secs}-{index}.bin.gz
+/// {prefix}/date={YYYY-MM-DD}/time={HHMM}/service={service_name}/instance={instance_path}/boot={boot_id}/{epoch_secs}-{index}.bin.gz
```
rcoh: should we still have epoch-index as the final key component?

dial9-core/src/segment_object_key.rs
noxware: Provisional location. I was planning to ask about this.

This is mainly string encoding/decoding shared by `dial9-viewer` and `dial9-destinations-s3`.

My proposal would be to expose it from `dial9-destinations-s3`. The responsible of uploading, responsible of providing the decoder :P

But of course, that crate is currently focused in uploading and has a lot of dependencies. But would be better than making a shared published crate just for this or duplicating the file, I think.

dial9-destinations-s3/src/s3.rs
noxware: Mmm... good question. If the data there is very important, then, probably.

Some ideas:

```
.../start={epoch}/{index}.bin.gz
.../start={epoch}/index={index}/trace.bin.gz
.../start={epoch}/index={index} (no file name)
```

(basically the same idea, just moving the line)

Having both, `epoch` and `date/time` is weird but I suppose we will need to keep them for the efficient `LIST` call. Idk, maybe we can remove them, encode `start` as an ISO 8601 datetime after `service`, using up to the `MM` precision as a prefix for lookup. Idk if s3 supports prefixing out of the conventional `/` boundary or if there is a reason for us to put `date/time` before `service`.

dial9-core/src/segment_object_key.rs
noxware: > My proposal would be to expose it from dial9-destinations-s3

Although, that would still increase the a public API surface just to share a bit of code. So I'm actually not that convinced :/

noxware: I will split and keep this private in each crate. The viewer will hold the decoder and the s3 crate will hold the encoder. Although, currently the viewer needs the escaping logic as well, I may just duplicate that as it's not much code.

dial9-destinations-s3/src/s3.rs
```diff
@@ -188,7 +196,7 @@ impl S3Config {
     ///
     /// If a custom `key_fn` is set, delegates to it. Otherwise uses the
     /// default time-first layout:
-    /// `{prefix}/{date}/{HHMM}/{service}/{instance}/{boot_id}/{epoch_secs}-{index}.bin.gz`
+    /// `{prefix}/date={date}/time={HHMM}/service={service}/instance={instance}/boot={boot_id}/{epoch_secs}-{index}.bin.gz`
```
rcoh: I think we should restrict opaque prefixes to `[a-zA-Z0-9-]` (in practice this will always be `dial9-traces`, there aren't many reasons to override it)

rcoh (commented):
Could we simplify parsing and remove ambiguity by making `version=1` the first path component after the default prefix? We could potentially even 

0. Disallow `k=v` components in the configured prefix. In practice the prefix is `dial9-traces`. I would even consider fully removing the setter and waiting for someone to ask. I think consistency here is probably more useful and I don't know many practical reasons to override.

1. Change the new layout to:

`{prefix}/version=1/date=D/service=S/time=T/instance=I/boot=B/file`

Browser browse requests are service-scoped: service discovery happens first, followed by `/api/browse` for one selected service. This ordering enables:

- Service-name discovery with one delimiter LIST per UTC day.
- Layout discovery per `(day, service)`: `legacy`, `v1`, or `both`.
- Querying both layouts only on a service’s crossover day.
- A stable version anchor for future key-layout changes.

`/api/services` could return an opaque layout hint that `/api/browse` consumes. `/api/browse` should retain discovery/fallback behavior when the hint is absent, for direct links and API callers. Host-count metadata may need to become lazy because `instance` remains below `time`.

The current implementation generates and queries both historical and Hive prefixes unconditionally. This permanently doubles many LIST fan-outs and causes a service-scoped 24-hour range to exceed the existing prefix cap.

2. Clearly document that overriding `S3Config::key_fn` forfeits built-in browser discovery unless the custom function emits a supported versioned layout.

As a later follow up, we could provide a restricted an API to allow adding K/V pairs to the hive-style layout but that seems unnecessary for now.

rcoh (approved):

dial9-viewer/src/ingest/refine.rs
```diff
@@ -512,52 +577,114 @@ fn time_scoped_prefixes(source_prefixes: &[String], scope: &Scope) -> Vec<String
...
+            let encoded_service = crate::segment_object_key_codec::hive_escape(service);
+            append_buckets(day.from, day.to, window.granularity, |date, time| {
+                output.push(join_prefix(
+                    base,
+                    &format!("version=1/date={date}/service={encoded_service}/time={time}"),
+                ));
+            });
         }
-        prefixes
+    }
```
noxware, line 655: This can blow up S3 LISTs for queries without a service.

Main did one LIST per time bucket. This does one per service per bucket, so 2 hours across 100 services is about 12.5k LISTs with no cap.

Should we reject unscoped queries or return an error when they would generate too many LIST requests?

noxware:
@rcoh The 4 commits I've kept force pushing are almost the same as the previous diff, in an easier to review format. Each is (more or less) one unit of work. And I just added modifications directly to them as if they were pseudo-PRs.

The main important thing is: The first day V1 appears for a service, even if it has legacy mixed, will assume the next days will be V1 only, to avoid further expensive checks for legacy entries. This is a discovery x performance tradeoff.

Talking about performance, please check my comment on the for loop. Are requests without explicit service important?

rcoh:
I think this is good to merge. We can always iterate on the viewer later;
the only thing that needs to be perfect right now is the format we emit for
0.5

noxware:
Queued merge, with https://github.com/dial9-rs/dial9/issues/809 as follow up.

PR #798, opened 2026-08-18
Make the tabs for dial9 Tasks/Issues tables resizable

### LINKED ISSUE #795: make the tasks tab expandable

skanjalkar:
Closes #795 
- The issues/tasks rail is drag-resizable at its right edge (min 220px, max 60vw), persisted and shareable via a new rail-width URL param.
- Both rail tables get per-column resize: drag a header divider to resize just that column (first drag seeds all columns from their rendered widths and switches to fixed layout; the table scrolls horizontally inside the rail when columns outgrow it). Double-click a divider resets to content-fit.
- Resized headers center in their column; shrinking floors at each header label's measured width so it never clips.
- Widths persist in the dial9.viewer.trackPrefs blob and ride in links via new additive task-cols / issue-cols params (README URL-contract tables + ledger updated).

Screenshot from 2026-08-18 13-27-46

Its a little rough around the edges:
image

as you see here, the left tab goes over the subtabs. I think that's finem just a little rough around the edges.


rcoh:
While we are here, I think we should update how the task spawn location is elided so we delete the _beginning_ on overflow instead of the end. The beginning is almost always less relevant than the end

rcoh:
Otherwise, this looks nice!

skanjalkar:
I think that does it?

image

rcoh:
looks great

rcoh (approved):

PR #802, opened 2026-08-19
fix(viewer): no whole-trace blue box when opening toolbar analyses

### LINKED ISSUE #796: clicking "flamegraph" or "blocking calls" causes a blue highlight over the entire trace
presumably, this happens because those buttons simulate a "select all" on the trace. In any case, this is confusing for users as well as making the UI harder to see.

skanjalkar:
The Flamegraph / Blocking Calls / Heap toolbar buttons open a whole-trace analysis by retaining selection.sidebarRange = [minTs, maxTs], and the selection overlay boxed that range  - 

image

tinting the entire trace blue, which reads as a rendering bug.

After this change it no longer does that.

image

Closes #796


rcoh (approved):
LGTM thanks!

PR #804, opened 2026-08-19
docs: cleanup doc fillers, fix broken links

jv1i:
Docs cleanup, mostly removing some filler/slop from public docs and fixing some broken links (unlinking in a few cases where links wouldn't work or be necessary).

Also adds docs.rs metadata for `dial9-core`, `dial9-trace-format` and `dial9-metrique.


dial9-macro/src/lib.rs
```diff
@@ -225,9 +225,9 @@ fn expand_main(args: MainArgs, input: ItemFn) -> Result<TokenStream2, syn::Error
 /// Instrument an async main function with dial9 telemetry.
 ///
 /// This macro is a **replacement** for `#[tokio::main]`, not a complement —
-/// do not use both attributes on the same function. Your `config` yields a [`dial9::Recorder`] and its
+/// do not use both attributes on the same function. Your `config` yields a [`dial9::Recorder`](../dial9/struct.Recorder.html) and its
```
rcoh, line 228: does linking to html work?

rcoh (approved):

dial9-macro/src/lib.rs
jv1i: they do, I copied it copied from how tokio does it with their macro

PR #805, opened 2026-08-20
feat(core): opt-in process-global Dial9Handle

### LINKED ISSUE #699: Feature request: Global handle for non-Tokio threads
## Problem

`Dial9Handle::current()` only works on Tokio worker threads. For synchronous hot loops (no Tokio runtime on the profiled thread), there's no way to access the handle without manually threading it through every function signature.

## What I'm doing today

```rust
use std::sync::OnceLock;

static HANDLE: OnceLock<TelemetryHandle> = OnceLock::new();

// After TracedRuntime is created:
HANDLE.set(traced.guard().handle()).unwrap();

// In the hot loop:
if let Some(h) = HANDLE.get().filter(|h| h.is_enabled()) {
    record_event(MyEvent { ... }, h);
}
```

Every non-Tokio user has to roll their own `OnceLock` wrapper.

## Proposal

A global handle set during `TracedRuntime` construction, accessible from any thread:

```rust
// Works from any thread, returns disabled handle if not initialized:
let h = Dial9Handle::global();
```

Or equivalently, a fallback in `current()` that checks the global when no thread-local is set.

### Optional: `record!` macro

Separately, a convenience macro that wraps the null-check pattern would be nice but is trivial to write once the global exists:

```rust
dial9::record!(MyEvent { timestamp_ns: clock_monotonic_ns(), value: 42 });
```

jv1i:
closes #699

## Context 

`Dial9Handle::current()` only resolves on threads a dial9 runtime owns, so code in a synchronous hot loop had no way to reach telemetry without passing a handle through every call site.

## UPDATE

Went with the review. No `Dial9Handle::global()`: `Recorder::install_global_handle()` is the opt-in, and `current()` falls back to that handle when the calling thread has none of its own, so there is one resolver instead of two.

Runtime ownership moved into the Tokio source as a thread-local flag set by the runtime's thread hooks.

Everything below is the original proposal and the discussion that got here.

## Changes
A new `Recorder::install_global()` publishes the recorder's handle process-wide, and `Dial9Handle::global()` reads it from any thread. This is cleared when the recorder stops, so a stale global can't outlive it.

The slot is an `ArcSwapOption`, so reads are lock-free. Reaching the global from an arbitrary call site is the point of installing one, and a mutex there would have been the only contended thing in an otherwise thread-local recording path.

```rust
let (recorder, rt) = dial9::recorder_from_env()?;
recorder.install_global();

std::thread::spawn(|| {
    Dial9Handle::global().record_event(MyEvent { .. });
});
```

## Not making `current()` fall back to it

Up for discussion, my reasoning:

The issue offered a fallback in `current()` as an equivalent alternative. I'm holding off, because as we have it now, `current()` means "handle for this thread" (beside users, we also have `dial9::spawn` and the tracing layer use it to mean "is this thread runtime-owned")
Left alone, a fallback would make spawn wrap futures on runtimes dial9 isn't attached to, emitting wake events with `WorkerId::UNKNOWN`, and would record spans on every non-runtime thread in the process.
So we could merge "this thread" and "global" into a single `current()` with a fallback but it might not be as intuitive, and we'd have to make a decision on having spawn instrumentation and tracing spans be handle-scoped rather than runtime scoped.

Either way it's the reversible direction: adding the fallback later is additive, taking it away once people rely on it isn't.

`global()` is explicit, and `current()`, `try_current()` and spawn instrumentation are all untouched.

`dial9::record_event()` does fall back to the global, so it works from any thread once you opt in. 
I think recording from a free fn (not from ::current) will likely read as global so we might as well have it cover that case.


rcoh:
A good discussion of trade offs. I think I would rather scrap `::global()` entirely as an API and document that `::current()` falls back to global _iff_ `install_global()` has been called.

For Tokio-specific behavior, I recommend having a separate mechanism to track whether we are tracing _specific_ runtime threads; knowing whether a specific runtime is tracked is really a concern of the Tokio dial9 source and not dial9 itself.

jv1i:
Got it, updated to scrap global and made `::current()` fall back to global, I think that turns out cleaner. 
Renamed `install_global` to `install_global_handle`, and added explicit marking of threads as traced directly in tokio-telemetry

dial9-core/src/handle.rs
```diff
@@ -123,23 +136,27 @@ impl Dial9Handle {
...
             .with(|cell| cell.borrow().clone())
+            .or_else(global_handle)
             .unwrap_or_else(Self::disabled)
     }
 
     /// Return the [`Dial9Handle`] installed for the current thread,
     /// or `None` if no dial9 runtime has claimed this thread.
     ///
-    /// Prefer [`current`](Self::current) instead.
+    /// Does not consider the process-global handle. Prefer
```
rcoh: does this intentionally behave differently from `current`? Seems like it should be the exact same but without the `Self::disabled` fallback?

dial9-core/src/handle.rs
```diff
@@ -328,8 +345,41 @@ pub fn clear_tl_handle() {
...
+/// recorder's live handle.
+pub(crate) fn set_global_handle(handle: Dial9Handle) {
+    let displaced_live_other = GLOBAL_HANDLE.load().as_ref().is_some_and(|installed| {
+        !installed.shared.is_stopped()
+            && !handle
+                .shared()
+                .is_some_and(|s| Arc::ptr_eq(s, &installed.shared))
+    });
+    if displaced_live_other {
+        tracing::warn!(target: "dial9", "replacing a live process-global Dial9Handle");
```
rcoh: should be rate limited

dial9-core/src/recording.rs
```diff
@@ -121,6 +121,37 @@ impl Recorder {
         &self.handle
     }
 
+    /// Publish this recorder's handle as the process-global one, so
+    /// [`Dial9Handle::current`] resolves it on threads no dial9 runtime owns.
```
rcoh: `dial9 runtime` is not a defined term. This comment would be clearer if it was positively phrased as something like When set, `Dial9Handle::current` will resolve on all threads in this process. When not set, only threads explicitly tracked (e.g. via a Tokio runtime being tracked by dial9) will resolve.

dial9-core/src/recording.rs
```diff
@@ -121,6 +121,37 @@ impl Recorder {
...
+    ///
+    /// Subsequent installs replace the previous one. The global is cleared when
```
rcoh: should it return something to inform the caller that it replaced?

dial9-utils/src/tracing_layer.rs
```diff
@@ -410,11 +410,7 @@ where
     }
 
     fn on_close(&self, id: span::Id, _ctx: Context<'_, S>) {
-        let Some(handle) = Dial9Handle::try_current() else {
-            return;
-        };
-
-        handle.record_event(SpanCloseEvent {
+        Dial9Handle::current().record_event(SpanCloseEvent {
```
rcoh: this was actually on purpose so that we didn't fire off and construct events when no one was listening. 

This should use:
```
   if !handle.is_enabled() {
            return;
        }
```

dial9/src/lib.rs
```diff
@@ -68,9 +68,11 @@ pub mod s3 {
 
 use crate::core::{Encodable, current_handle};
 
-/// Record an event on the calling thread's current handle.
+/// Record an event on the handle [`Dial9Handle::current`] resolves: this
+/// thread's, or the process-global one.
 ///
-///  A no-op when no recorder is installed on the thread (the handle is disabled).
+/// A no-op when neither is installed. Opt into the global with
+/// [`Recorder::install_global_handle`](dial9_core::recording::Recorder::install_global_handle).
```
rcoh: a little wordy and meandering

rcoh (commented):
There are a few weird soundness problems from allowing one handle to replace another. 

And you can really only have one global handle anyway.

I recommend not allowing replacing the global handle; only bad things can happen if someone actually did that.

dial9-core/src/handle.rs
rcoh: I think maybe this should be `try_current_thread`? (and keep this behavior)?

dial9-core/src/handle.rs
```diff
@@ -134,17 +134,17 @@ impl Dial9Handle {
...
+    /// Return the [`Dial9Handle`] to record through: the calling thread's, or
+    /// the process-global one.
     ///
-    /// Threads claimed by a dial9 runtime (via [`set_tl_handle`], cleared by
-    /// [`clear_tl_handle`]) get that runtime's handle. Any other thread gets the
-    /// process-global handle, if one was installed with
-    /// [`Recorder::install_global_handle`](crate::recording::Recorder::install_global_handle).
-    /// With neither, an inert handle whose methods are all no-ops, see
-    /// [`Dial9Handle::disabled`].
+    /// Threads a dial9 runtime owns get that runtime's handle. Any other thread
```
rcoh: runtime isn't really meaningful here.

dial9-core/src/handle.rs
```diff
@@ -134,17 +134,17 @@ impl Dial9Handle {
...
+    /// gets the process-global handle, once
+    /// [`Recorder::install_global_handle`](crate::recording::Recorder::install_global_handle)
+    /// has been called. With neither you get an inert
+    /// [`disabled`](Self::disabled) handle and recording is a no-op.
```
rcoh: should we allow multiple runtimes to exist in a production setting? not clear that we should

rcoh:
Naming wise, maybe we have `try_current_thread` as the TL one? `Dial9Handle::current()` does TL -> global -> disabled?

There is a lot of weird stuff that can happen if you have local handles and a global handle though. At a minimum, Tokio shouldn't run through the global handle since if the global handle is different than the one set on the runtime, weird and unexpected things happen.

dial9-utils/src/tracing_layer.rs
jv1i: restored

dial9-core/src/recording.rs
jv1i: this now returns a Result and rejects replacement

dial9-core/src/handle.rs
jv1i: removed this warn now entirely since we no longer replace

rcoh (approved):

PR #814, opened 2026-08-22
fix(perf-self-profile): detect missing frame pointers before install

### LINKED ISSUE #779: Probe for frame pointers on memory profiling install
Currently, if you install memory profiling without frame pointers, you can get segfaults from reading garbage memory. This is in contrast to cpu profiling, where it gracefully degrades. This caused confusion in the wild when a user put their rustflags in their `Cargo.toml` rather than `config.toml` (prompting https://github.com/dial9-rs/dial9/pull/778)

Ideally we should instead return a `InstallError::MissingFramePointers` on `MemoryProfiler::install()` and let the user decide to crash out or skip.

I'm not sure the best way to do this, perhaps some sort of heuristic-based frame-count threshold with a chain of `#[inline(never)]`? But this seems false positive prone given that `rbp` is callee-saved and there is prebuilt std that has frame pointers.

In the passed we've tried to use address window matching but we have seen flakiness in our tests, so that doesn't seem great.

The heavier overhead option would be actually doing symbolization matching... it's only on `install()` so maybe the hit is acceptable. Though, for stripped binaries, we would anyway need a fallback to eg frame-count threshold.

tiny-ac:
closes #779

- add a frame-pointer self-test run at MemoryProfiler::install() time
  - detection uses frame count
  - log a warning whenever the self-test fails
  - configurable behavior through `fail_on_missing_frame_pointers`:
    - return `InstallError::MissingFramePointers` on failure (default) - `fail_on_missing_frame_pointers(true)`
    - proceed (return Ok) - `fail_on_missing_frame_pointers(false)`
- add CI job verifying the self-test catches a build without -C force-frame-pointers=yes

Tests
- add coverage for the hard-fail and fail_on_missing_frame_pointers(false) paths
- Ran
  - cargo check -p dial9-perf-self-profile / -p dial9 --features memory-profiling
  -  cargo test -p dial9-perf-self-profile --lib (28 passed)
  -  cargo fmt --check, clippy clean
  -  cargo test -p dial9 --features memory-profiling --test memory_profiling_missing_frame_pointers -- --ignored, in Docker (2 passed)
  - cargo nextest run --stress-duration 20s (1301 passed, 0 failed)

### DIFF
--- a/dial9/tests/memory_profiling_missing_frame_pointers.rs
+++ b/dial9/tests/memory_profiling_missing_frame_pointers.rs
@@ -1,19 +1,47 @@
 #![cfg(feature = "memory-profiling")]
 #![cfg(target_os = "linux")]
-//! Negative test for the frame-pointer self-test in `MemoryProfiler::install()`.
+//! Tests for the frame-pointer self-test in `MemoryProfiler::install()`: the
+//! default hard-fail behavior, `fail_on_missing_frame_pointers`'s opt-out
+//! (install succeeds instead of failing), and that a failed self-test always
+//! logs a warning regardless of that setting.
 //!
-//! This only exercises the intended failure mode when the binary is built
+//! These only exercise the intended failure mode when the binary is built
 //! WITHOUT `-C force-frame-pointers=yes` — every other job in
-//! `.github/workflows/ci.yml` builds with it, so running this under a normal
-//! build would (correctly) fail the assertion below. `#[ignore]`d so normal
-//! `cargo test`/`cargo nextest run` skip it; run explicitly (`-- --ignored`)
-//! from a job whose `RUSTFLAGS` omits the flag. See the
-//! `frame-pointer-detection` CI job.
+//! `.github/workflows/ci.yml` builds with it, so running these under a normal
+//! build would (correctly) fail. `#[ignore]`d so normal `cargo test`/
+//! `cargo nextest run` skip them; run explicitly (`-- --ignored`) from a job
+//! whose `RUSTFLAGS` omits the flag. See the `frame-pointer-detection` CI job.
 
-use dial9::memory::{InstallError, MemoryProfiler};
+use dial9::memory::{InstallError, MemoryProfiler, MemoryProfilingConfig};
 use dial9::{Dial9HandleTokioExt, TokioAttachOptions, recorder};
+use std::sync::{Arc, Mutex};
 
 mod common;
+
+#[derive(Clone, Default)]
+struct SharedBuf(Arc<Mutex<Vec<u8>>>);
+
+impl std::io::Write for SharedBuf {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        self.0.lock().unwrap().extend_from_slice(buf);
+        Ok(buf.len())
+    }
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
+    }
+}
+
+fn capturing_subscriber() -> (SharedBuf, impl tracing::Subscriber) {
+    let buf = SharedBuf::default();
+    let subscriber = tracing_subscriber::fmt()
+        .with_writer({
+            let buf = buf.clone();
+            move || buf.clone()
+        })
+        .with_ansi(false)
+        .finish();
+    (buf, subscriber)
+}
 
 #[test]
 #[ignore = "only meaningful when built without -C force-frame-pointers=yes"]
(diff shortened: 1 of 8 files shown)


dial9/tests/memory_profiling_missing_frame_pointers.rs
```diff
@@ -0,0 +1,38 @@
+#![cfg(feature = "memory-profiling")]
```
rcoh, line 1: sorry if I'm missing it, but this PR should definitely have a test that actually replicates the segfault behavior observed in the ticket when framepointers are not enabled.

The unwinder shouldn't segfault when frame pointers are missing; it uses a sigsev handler to catch segfaults due to reading garbage memory. Can you figure out what is going on?

tiny-ac: The issue seems to be these unsafe lines here
https://github.com/dial9-rs/dial9/blob/a36f1a141e3f31ed9574537ff8a4f95fdfa4737b/perf-self-profile/src/unwinder.rs#L219-L220

It's actually documented here that the SIGSEGV handler doesn't cover these reads https://github.com/dial9-rs/dial9/blob/a36f1a141e3f31ed9574537ff8a4f95fdfa4737b/perf-self-profile/src/unwinder.rs#L180-L185

Tried to reproduce this on main with recursive functions and building without the flag, it just silently gave bad data instead of segfaulting.

Still looking at whether there's a way to get real regression coverage here rather than synthetic reproduction.

rcoh: did you try using the memory profiler without framepointers enabled?

tiny-ac: Ran an experiment locally using Docker on version 0.3.15 of dial9 (built a small standalone crate for this)

  A — without the flag:

> docker run --rm -it --platform linux/amd64 -v /Users/agostinacamacho/S/dial9-memory-experiment:/work -e CARGO_TARGET_DIR=/tmp/ct -e RUSTFLAGS="--cfg tokio_unstable" rust:1.94.1-bookworm bash -c "cd /work && cargo run --release"
(It doesn't get to print `memory profiler installed; starting workload`)
image


  B — with the flag:

> docker run --rm -it --platform linux/amd64 -v /Users/agostinacamacho/S/dial9-memory-experiment:/work -e CARGO_TARGET_DIR=/tmp/ct2 -e RUSTFLAGS="--cfg tokio_unstable -C force-frame-pointers=yes" rust:1.94.1-bookworm bash -c "cd /work && cargo run --release"

image

Currently trying to get a symbolized stack trace to confirm the exact crash site

perf-self-profile/src/memory_profiling/profiler.rs
```diff
@@ -106,6 +106,13 @@ pub enum InstallError {
     AlreadyInstalled,
     /// The SIGSEGV handler used by the FP unwinder failed to install.
     Unwinder(std::io::Error),
+    /// The frame-pointer self-test run at install time indicates this
+    /// binary was not built with `-C force-frame-pointers=yes`; allocation
+    /// stacks would be near-empty and useless for profiling.
+    MissingFramePointers {
```
rcoh, line 113: add a `#[non_exhaustive]` here

perf-self-profile/src/unwinder.rs
```diff
@@ -125,6 +127,136 @@ impl Unwinder {
...
+    /// `-C force-frame-pointers=yes`.
+    ///
+    /// Recurses a fixed depth through this module's own `#[inline(never)]`
+    /// function on a dedicated thread, then inspects the captured stack —
+    /// by symbol name where available (layout-independent), falling back
+    /// to a raw frame-count threshold otherwise.
+    ///
+    /// Returns `None` if the self-test itself couldn't run (thread spawn
+    /// or panic) — inconclusive, not evidence of missing frame pointers.
+    pub fn self_test_frame_pointers(&self) -> Option<FramePointerSelfTest> {
```
rcoh: this should return a result so yo ucan see what the actual cause of failure was instead of throwing it away

perf-self-profile/src/unwinder.rs
```diff
@@ -125,6 +127,136 @@ impl Unwinder {
...
+///
+/// `#[non_exhaustive]` so new fields can be added without breaking callers.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct FramePointerSelfTest {
+    /// A match count if `matched_by_symbol`, otherwise a raw frame count.
+    pub frames_captured: usize,
+    /// Minimum `frames_captured` needed to pass.
+    pub expected_min: usize,
+    pub matched_by_symbol: bool,
```
rcoh: do these fields actually need to be public? As a user, I really just want to know if FP is enabled or not? Doesn't necessarily seem like I would have anything useful to do with this info

perf-self-profile/src/unwinder.rs
```diff
@@ -125,6 +127,136 @@ impl Unwinder {
...
+
+    /// `None` means nothing resolved at all (e.g. no symbolizer, or a
+    /// stripped binary) — callers should fall back to the frame-count
+    /// check, not read it as "no match". `Some(n)` is the match count,
+    /// which may be 0.
+    #[cfg(any(
+        target_os = "linux",
+        all(target_os = "android", target_arch = "aarch64")
+    ))]
+    fn verify_by_symbol(frames: &[u64]) -> Option<usize> {
```
rcoh: this can actually be quite slow because it will be the first pass of the symbolizer (hundreds of milliseconds).

We should avoid calling this in the startup path of an application as it will significantly increase startup latency

perf-self-profile/src/memory_profiling/config.rs
```diff
@@ -69,6 +69,21 @@ pub struct MemoryProfilingConfig {
...
+    /// a warning on failure. By default a failure also returns
+    /// [`InstallError::MissingFramePointers`]; set this to `false` to have
+    /// `install()` proceed instead, relying on the warning alone. Only do
+    /// that if you've independently verified frame pointers are enabled, or
+    /// are prepared to accept near-empty allocation stacks.
+    ///
+    /// [`MemoryProfiler::install`]: crate::memory_profiling::MemoryProfiler::install
+    /// [`InstallError::MissingFramePointers`]: crate::memory_profiling::InstallError::MissingFramePointers
+    #[builder(default = true)]
+    fail_on_missing_frame_pointers: bool,
```
tiny-ac, line 86: Wondering if we want the failure to be default behavior or just the warning

perf-self-profile/src/unwinder.rs
tiny-ac, line 252: > cargo check --target x86_64-linux-android -p dial9-perf-self-profile --features memory-profiling

Image

rcoh (approved):

PR #818, opened 2026-08-24
fix: safely bootstrap frame-pointer capture

rcoh:
## Summary

- validate the current frame pointer against the stack pointer before reading its frame record
- route bootstrap return-address and saved-frame-pointer reads through the existing fault-tolerant `safe_load` trampoline
- degrade missing or unusable frame pointers to empty or one-frame captures instead of crashing
- share the existing return-address and frame-size plausibility limits across bootstrap and chain walking

## Tests

- add protected-page coverage for an unreadable return-address slot
- add protected-page coverage proving a readable frame-zero PC survives an unreadable saved frame pointer
- `cargo nextest run` (1,402 passed)
- `cargo nextest run --stress-duration 20s` (1,402 passed)

## Performance

The fault-free path adds two direct `safe_load` calls and fixed validation per synchronous `Unwinder::capture`. The existing unwind benchmark indicates this is approximately one frame's fixed overhead; signal-based CPU unwinding and per-frame chain walking are unchanged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -313,6 +313,17 @@ jobs:
           sudo prlimit --pid $$ --memlock=unlimited:unlimited
           cargo test -p dial9-perf-self-profile --lib --all-features \
             -Z build-std --target x86_64-unknown-linux-gnu
+      - name: Run memory profiler under ASAN without frame pointers
+        env:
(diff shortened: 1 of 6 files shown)


jv1i (approved):
Great improvement! 
> degrade missing or unusable frame pointers to empty or one-frame captures instead of crashing
nice catch on that also

PR #844, opened 2026-08-28
Add in NVML based GPU profiling.

### LINKED ISSUE #836: Grab GPU (nvidia) telemetry
Working on some tokio services which utilise GPU acceleration via rust projects that bind to C++ code using CUDA. And generally the way I get CPU metrics it would be so handy to have a view of GPU memory and compute utilisation. I feel here making it vendor specific is probably necessary due to the amount of nvidia specific profiling etc but it's obviously a big space with a lot of potential uses wider than my own narrow focus so might need some design discussion.

xd009642:
Unfortunately this can't isolate the GPU usage to what is used by the process, that might involve more involved nvidia library bindings. I've seen some crates have created those bindings and vendored them into their crates but there's not a published version of them so might be a bit more of a headache.

I aim to start trying this out internally with some of our GPU based workloads and adapting it as needed but this seems like a reasonable enough start.

Fixes #836


rcoh (commented):
nice! approach seems good to me

perf-self-profile/src/lib.rs
```diff
@@ -153,8 +157,12 @@ pub use cpu_source::{
     CpuProfiler, CpuProfilingConfig, CpuSampleSource, SchedEventConfig, SchedProfiler,
 };
 
+#[cfg(feature = "cuda")]
+pub use cuda::{CudaGpuConfig, CudaGpuEvent, CudaGpuSource, CudaGpuStartError};
+
```
rcoh: lets leave these inside the cuda module to keep the root namespace from getting too busy

perf-self-profile/src/cuda.rs
```diff
@@ -0,0 +1,264 @@
+//! NVIDIA CUDA GPU metrics sampled through NVML.
```
rcoh, line 1: an example that uses this source would be nice

perf-self-profile/src/cuda.rs
```diff
@@ -0,0 +1,264 @@
...
+//!
+//! NVML is loaded dynamically, so enabling the Cargo feature does not require
```
rcoh: spell out nvml once, you can use the acronym after, a link like https://developer.nvidia.com/management-library-nvml would be nice too assuming that's the right link

perf-self-profile/src/recorder_ext.rs
```diff
@@ -37,6 +41,17 @@ pub trait RecorderPerfExt: Sized {
 }
 
 impl<T: RecorderSourceExt> RecorderPerfExt for T {
+    #[cfg(feature = "cuda")]
+    fn with_cuda_gpu_profiling(self, config: crate::CudaGpuConfig) -> Self {
```
rcoh: is there a way to use this API so that the caller could action on the error (e.g. abort during startup)? It doesn't have to be via `RecorderExt` (although try_with_cuda_gpu_profiling` that returned an error would be a pretty reasonable API to add)

perf-self-profile/Cargo.toml
```diff
@@ -52,6 +55,7 @@ dial9-trace-format = { workspace = true }
 scc = { version = "2", optional = true }
 tokio = { version = "1", features = ["rt"], optional = true, default-features = false }
 tracing = "0.1.44"
+nvml-wrapper = { version = "0.12.1", optional = true }
```
rcoh: 0.13 of this crate is available

rcoh (commented):
just small stuff. Thanks for your contribution!

perf-self-profile/src/recorder_ext.rs
xd009642: I've added aa try_with_cuda_gpu_profiling

perf-self-profile/src/cuda.rs
xd009642: By example do you mean like the example project? I just added that into the project tree (before I linked a separate repo version I made but it's now in the latest commit)

perf-self-profile/src/cuda.rs
```diff
@@ -0,0 +1,283 @@
...
+#[derive(Debug, Clone, bon::Builder)]
+pub struct CudaGpuConfig {
+    /// Minimum time between samples.
+    #[builder(default = DEFAULT_SAMPLE_INTERVAL)]
+    sample_interval: Duration,
+    /// Whether to sample aggregate PCIe transfer throughput.
+    #[builder(default = true)]
+    sample_pcie_throughput: bool,
+    /// NVML device index to sample. By default, all visible devices are sampled.
+    device_index: Option<u32>,
```
rcoh: I'll leave this to your judgement; is all visible devices what we want?

rcoh: I guess the events show up sensibly. The problem is only that the graphs look bad.

rcoh (approved):

perf-self-profile/src/cuda.rs
xd009642: Yeah I don't think all visible devices on the same graph makes sense tbh. I'm happy to default it to index 0 tbh. I'll just make that change and push it

PR #847, opened 2026-08-28
perf(viewer): flatten aggregate flamegraph trees

rcoh:
## Summary

- Add the negotiated `flat-v1` flamegraph encoding on top of the interned frame table.
- Serialize preorder rows as `[parent_node, frame, count, self_count]` instead of recursively nested objects.
- Reconstruct the existing nested UI tree at the API boundary.
- Preserve `interned-v1` and legacy response handling.

## Why

Frame interning removes duplicated symbol text, but a deeply nested object still carries repeated JSON property names and recursive structure for every trie node. Parent-indexed rows reduce per-node wire overhead and avoid serializing a recursive object graph.

## Dependency

Depends on #846.

## Stack

1. #846 Frame interning
2. **#847 Flat parent-indexed tree rows (this PR)**
3. #848 Deterministic node budget
4. #849 Coverage-only progress events
5. #850 Transactional in-place part merges

## Validation

Validated on the complete stack:

- `cargo fmt --check`
- `cargo nextest run` (1,444 passed)
- `cargo nextest run --stress-duration 20s` (1 full iteration, 1,444 passed)
- UI typecheck, import-boundary check, Vitest (2,181 passed), and Vite production build


dial9-viewer/src/server/flamegraph.rs
```diff
@@ -568,6 +684,7 @@ struct StreamCtx {
 enum WireFormat {
     Legacy,
     InternedV1,
+    FlatV1,
```
jv1i, line 661: Is there anything that will use InternedV1 later? can't tell from looking at the PRs

dial9-viewer/src/server/flamegraph.rs
```diff
@@ -182,96 +197,211 @@ pub struct ScopeEcho {
...
+    names: Vec<String>,
+    ids: HashMap<FrameId, u32>,
+}
+
+impl WireFrames {
+    fn new(nodes: &[TrieNode], stacks_dict: &StackDictionary) -> Self {
+        let mut frames: Vec<_> = nodes.iter().map(|node| node.frame).collect();
         frames.sort_unstable_by(|a, b| stacks_dict.resolve(*a).cmp(stacks_dict.resolve(*b)));
-        let wire_ids: HashMap<_, _> = frames
+        frames.dedup();
```
jv1i, line 378: Should we dedupe the IDs before sorting here? Otherwise we end up comparing all the nodes rather than just the distintct ones right?

jv1i (approved):
LGTM, leaving some comments

PR #848, opened 2026-08-28
perf(viewer): bound flamegraph projections

rcoh:
## Summary

- Cap `flat-v1` flamegraph projections at 50,000 rows.
- Select retained nodes deterministically and aggregate omitted subtrees into synthetic `[other]` rows.
- Preserve total and self counts across the bounded projection.
- Retain one deterministic path to the requested inspect frame when it exists.
- Report exact and omitted node counts to the UI.

## Why

Compact encodings reduce bytes per node but do not bound pathological cardinality. The production trace built a 1.65 million-node trie; this projection budget places a hard ceiling on the browser-facing response while retaining the highest-value paths and count conservation.

## Dependency

Depends on #847.

## Stack

1. #846 Frame interning
2. #847 Flat parent-indexed tree rows
3. **#848 Deterministic node budget (this PR)**
4. #849 Coverage-only progress events
5. #850 Transactional in-place part merges

## Validation

Validated on the complete stack:

- `cargo fmt --check`
- `cargo nextest run` (1,444 passed)
- `cargo nextest run --stress-duration 20s` (1 full iteration, 1,444 passed)
- UI typecheck, import-boundary check, Vitest (2,181 passed), and Vite production build

### DIFF
--- a/dial9-viewer/src/server/flamegraph.rs
+++ b/dial9-viewer/src/server/flamegraph.rs
@@ -3,13 +3,14 @@
 //!
 //! One request holds the connection open: it [resolves](refine::resolve) the
 //! scope, streams already-folded parts in bounded cumulative snapshots, then
-//! folds missing capped-prefix files and pushes a fresh full-tree snapshot as
-//! each file lands, closing when the bounded work-list drains. Each SSE `data:`
+//! folds missing capped-prefix files and pushes a fresh bounded-tree snapshot
+//! as each file lands, closing when the bounded work-list drains. Each SSE `data:`
 //! frame is one
 //! [`FlamegraphResponse`] JSON object — the same shape the UI rendered per poll
 //! before — so the client just re-renders on every event.
 
-use std::collections::HashMap;
+use std::cmp::Reverse;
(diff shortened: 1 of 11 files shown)


dial9-viewer/src/server/flamegraph.rs
```diff
@@ -146,6 +152,13 @@ pub struct FlatFlamegraphTree {
     /// row zero and names itself as its parent; every other parent precedes its
     /// children.
     pub nodes: Vec<FlatFlamegraphNode>,
+    /// Exact trie node count before projection.
+    pub total_nodes: usize,
+    /// Exact trie nodes represented only through synthetic `[other]` rows.
+    pub omitted_nodes: usize,
+    /// True when a requested inspect frame was found and its path retained.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub inspect_retained: Option<bool>,
```
jv1i, line 161: is `insepcted_retained` meant to be used by the UI later? can't find uses for it from a quick look
I wonder if that is also covered by `restoreAfterTreeChange()`

jv1i (approved):

PR #851, opened 2026-08-28
test(dial9-core): add shuttle coverage for SharedState concurrency

tiny-ac:
#764 

## Summary
- Adds `shuttle` coverage for `SharedState` concurrency paths
  - writer/drainer race on the intrusive-drain path (shuttle counterpart of the existing real-thread proptest)
  - `disable`/`enable`/`mark_stopped` raced against concurrent recording
  - writer's thread-exit flush racing `drain_all_tl_buffers`

## Notes
- New scenarios were mutation-tested.
- Scoped to `SharedState`
- `drain_epoch`/`flush_epoch`/`state` are `Ordering::Relaxed`; shuttle only models `SeqCst`. A `Relaxed`-specific bug would slip through (that's why the proptest is kept)
- `shuttle_disable_races_record` doesn't assert "no events after `disable()`". The gate is load-then-record, not atomic, so that's expected behavior.
- `shuttle_writer_exit_races_drain` covers a distinct interleaving shape, but mutation testing didn't turn up a bug only it catches

### DIFF
--- a/dial9-core/src/shared_state.rs
+++ b/dial9-core/src/shared_state.rs
@@ -313,23 +313,25 @@ impl EventBuffer<'_> {
 }
 
 #[cfg(test)]
+fn sample_event() -> crate::format::ClockSyncEvent {
+    crate::format::ClockSyncEvent {
+        timestamp_ns: 1000,
+        realtime_ns: 2000,
+    }
+}
+
+/// Helper: create a SharedState with recording enabled.
+#[cfg(test)]
+fn enabled_shared_state() -> SharedState {
+    let ss = SharedState::new(0);
+    ss.enable();
+    ss
+}
+
+#[cfg(test)]
 mod tests {
     use super::*;
     use crate::primitives::sync::atomic::AtomicBool;
-
-    fn sample_event() -> crate::format::ClockSyncEvent {
-        crate::format::ClockSyncEvent {
-            timestamp_ns: 1000,
-            realtime_ns: 2000,
-        }
-    }
-
-    /// Helper: create a SharedState with recording enabled.
-    fn enabled_shared_state() -> SharedState {
-        let ss = SharedState::new(0);
-        ss.enable();
-        ss
-    }
 
     #[test]
     fn record_event_registers_tl_buffer_handle() {
@@ -535,6 +537,13 @@ mod tests {
     }
 }
 
+// Every scenario below that touches `drain_epoch`/`flush_epoch`/`state`
+// shares one caveat: those fields are `Ordering::Relaxed` throughout, but
+// shuttle 0.9.1 only correctly models `SeqCst` (it treats every other
+// ordering as `SeqCst`). A `Relaxed`-specific bug is invisible here
+// regardless of how exhaustively a scenario searches, even though it's
+// reachable on real hardware -- only the surrounding CAS/transition logic
+// is actually under test.
 #[cfg(all(test, shuttle))]
 mod shuttle_tests {
     use super::*;
@@ -548,19 +557,6 @@ mod shuttle_tests {
     const EVENTS_PER_WRITER: u64 = 3;
     const DRAIN_TICKS: usize = 3;
 
-    fn sample_event() -> crate::format::ClockSyncEvent {
-        crate::format::ClockSyncEvent {
-            timestamp_ns: 1000,
-            realtime_ns: 2000,
-        }
-    }
-
-    fn enabled_shared_state() -> SharedState {
-        let ss = SharedState::new(0);
-        ss.enable();
-        ss
-    }
-
     crate::shuttle_test! {
         default;
         // Shuttle counterpart of `concurrent_record_and_drain_preserves_event_count`:
@@ -568,6 +564,9 @@ mod shuttle_tests {
         // the drain epoch and intrusively drains, exploring interleavings
         // exhaustively instead of sampling them. Every recorded event must reach
         // the collector exactly once, regardless of schedule.
+        //
+        // Keeping real-thread proptest above because of the `Ordering::Relaxed` caveat
+        // this scenario shares with the others below (see module comments).
         fn shuttle_concurrent_record_and_drain() {
             let ss = Arc::new(enabled_shared_state());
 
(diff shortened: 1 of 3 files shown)


dial9-core/src/shared_state.rs
```diff
@@ -534,3 +536,222 @@ mod tests {
         }
     }
 }
+
```
tiny-ac, line 536: Hand-injected synthetic faults to see which would be caught by the new tests

**M1 — `enable()` loses its CAS guard**

  `dial9-core/src/shared_state.rs`, `SharedState::enable` ([currently lines 115-122](https://github.com/dial9-rs/dial9/blob/64b8e9ff6af3fde0e13cbcd31b6e7c6b5891913f/dial9-core/src/shared_state.rs#L115-L122)):
  ```diff
   fn enable(&self) {
  -    let _ = self.state.compare_exchange(
  -        State::Disabled as u8,
  -        State::Enabled as u8,
  -        Ordering::Relaxed,
  -        Ordering::Relaxed,
  -    );
  +    self.state.store(State::Enabled as u8, Ordering::Relaxed);
   }
  ```

  **M2 — `disable()` loses its CAS guard** (mirror of M1)

  `dial9-core/src/shared_state.rs`, `SharedState::disable` ([currently lines 126-133](https://github.com/dial9-rs/dial9/blob/64b8e9ff6af3fde0e13cbcd31b6e7c6b5891913f/dial9-core/src/shared_state.rs#L126-L133)):
  ```diff
   pub(crate) fn disable(&self) {
  -    let _ = self.state.compare_exchange(
  -        State::Enabled as u8,
  -        State::Disabled as u8,
  -        Ordering::Relaxed,
  -        Ordering::Relaxed,
  -    );
  +    self.state.store(State::Disabled as u8, Ordering::Relaxed);
   }
  ```

  **M3 — TOCTOU reintroduced between the enabled-gate and the write**

  `dial9-core/src/shared_state.rs`, `EventBuffer::record_encodable_event`
  ([currently lines 300-306](https://github.com/dial9-rs/dial9/blob/64b8e9ff6af3fde0e13cbcd31b6e7c6b5891913f/dial9-core/src/shared_state.rs#L300-L306)):
  ```diff
   pub(crate) fn record_encodable_event(&self, event: &dyn encoder::Encodable) {
  +    // Redundant re-check reintroduces a TOCTOU between if_enabled's outer
  +    // gate and this call.
  +    if !self.0.is_enabled() {
  +        return;
  +    }
       if let Some(handle) =
           encoder::record_encodable_event(event, &self.0.collector, &self.0.drain_epoch)
       {
           self.0.tl_buffers.lock().unwrap().push(handle);
       }
   }
  ```

  **M4 — dead-handle pruning silently disabled**

  `dial9-core/src/shared_state.rs`, `SharedState::drain_all_tl_buffers`
  ([currently line 255](https://github.com/dial9-rs/dial9/blob/64b8e9ff6af3fde0e13cbcd31b6e7c6b5891913f/dial9-core/src/shared_state.rs#L255)):
  ```diff
   let mut guard = self.tl_buffers.lock().unwrap();
   let before = guard.len();
  -guard.retain(|h| h.buffer.strong_count() > 0);
  +guard.retain(|_h| true);
   stats.dead_pruned = (before - guard.len()) as u64;
  ```

  **M5 — off-by-one on thread-exit flush (drops the last pending event)**

  `dial9-core/src/encoder.rs`, `impl Drop for ThreadLocalBuffer`
  ([currently lines 285-287](https://github.com/dial9-rs/dial9/blob/64b8e9ff6af3fde0e13cbcd31b6e7c6b5891913f/dial9-core/src/encoder.rs#L287)):
  ```diff
   impl Drop for ThreadLocalBuffer {
       fn drop(&mut self) {
  -        if self.event_count > 0 {
  +        if self.event_count > 1 {
               if let Some(collector) = self.collector.take() {
                   collector.accept_flush(self.flush());
  ```

  **M6 — empty batch reaches the collector** (guard dropped in the drain loop)

  `dial9-core/src/shared_state.rs`, `SharedState::drain_all_tl_buffers`
  ([currently lines 241-246](https://github.com/dial9-rs/dial9/blob/64b8e9ff6af3fde0e13cbcd31b6e7c6b5891913f/dial9-core/src/shared_state.rs#L241-L246)):
  ```diff
   stats.buffers_locked += 1;
  -if buf.has_pending_events() {
  -    let batch = buf.flush();
  -    stats.events_flushed += batch.event_count();
  -    stats.buffers_flushed += 1;
  -    self.collector.accept_flush(batch);
  -}
  +let batch = buf.flush();
  +stats.events_flushed += batch.event_count();
  +stats.buffers_flushed += 1;
  +self.collector.accept_flush(batch);
  ```

dial9-core/src/shared_state.rs
```diff
@@ -534,3 +536,222 @@ mod tests {
...
+    crate::shuttle_test! {
+        default;
+        // `disable`/`enable`/`mark_stopped` raced against the recording
+        // gate (`if_enabled`). The shutdown-adjacent transition other
+        // tests skip. Every `Some` must land exactly one event, every
+        // `None` zero, regardless of interleaving. Does NOT assert "no
+        // events after disable()": a writer that passed `is_enabled()`
+        // completes its write after `disable()` returns, by design. Also
+        // subject to the `Ordering::Relaxed` caveat above.
+        fn shuttle_disable_races_record() {
```
tiny-ac, line 607: Catches [M1, M2, M3](https://github.com/dial9-rs/dial9/pull/851/changes#r3895696605)

M2 was initially missed (lifecycle thread's disable() calls both land before mark_stopped()) — fixed by mirroring the existing post-stop enable() assertion

dial9-core/src/shared_state.rs
```diff
@@ -534,3 +536,222 @@ mod tests {
...
+    }
+
+    crate::shuttle_test! {
+        default;
+        // Writer threads exit *while* the drainer is mid-drain, so the TLB
+        // `Drop` flush overlaps `drain_all_tl_buffers`'s `Weak::upgrade`+prune
+        // logic. Not every interleaving exercises the `Drop` flush itself,
+        // since that depends on whether a writer still holds a pending event
+        // at exit. Also subject to the `Ordering::Relaxed` caveat above.
+        fn shuttle_writer_exit_races_drain() {
```
tiny-ac, line 717: Catches [M4](https://github.com/dial9-rs/dial9/pull/851/changes#r3895696605)

Also caught by pre-existing plain test [drain_prunes_dead_handles](https://github.com/dial9-rs/dial9/blob/64b8e9ff6af3fde0e13cbcd31b6e7c6b5891913f/dial9-core/src/shared_state.rs#L401).  Mostly a regression insurance for an untested interleaving shape. Couldn't  find a bug shape this test would catch uniquely

tiny-ac: While searching for bugs shapes this test would caught, also tried these (targeted at "no grace period, one record per writer" design)

**1 - intrusive-drain lock split into two acquisitions (reopens a
  check-then-act window between "has pending events" and the flush)**

  `dial9-core/src/shared_state.rs`, `SharedState::drain_all_tl_buffers`
  (currently lines 226-249):
  ```diff
   if let Some(arc) = handle.buffer.upgrade() {
  -    let mut buf = match arc.lock() {
  -        Ok(guard) => guard,
  -        Err(_) => { /* ...rate-limited error, continue... */ }
  -    };
  -    stats.buffers_locked += 1;
  -    if buf.has_pending_events() {
  -        let batch = buf.flush();
  +    let has_pending = match arc.lock() {
  +        Ok(guard) => guard.has_pending_events(),
  +        Err(_) => { /* ...rate-limited error, continue... */ }
  +    };
  +    stats.buffers_locked += 1;
  +    if has_pending {
  +        let mut buf = arc.lock().unwrap();
  +        let batch = buf.flush();
           stats.events_flushed += batch.event_count();
           stats.buffers_flushed += 1;
           self.collector.accept_flush(batch);
       }
       handle.flush_epoch.store(epoch);
   }
```

**No prod impact, code is robust.** Both the intrusive
  drain and `Drop` serialize through the same `Mutex` either way; worst case
  is a harmless extra empty flush, never a double-count or lost event.

---

**2 — intrusive drain made a complete no-op**

  `dial9-core/src/shared_state.rs`, `SharedState::drain_all_tl_buffers`
  (currently line 220, top of the per-handle loop):
  ```diff
   for handle in &handles {
  +    continue;
       // Skip buffers that self-flushed during the current epoch.
       if handle.flush_epoch.load() >= epoch {
  ```
**The suite already covers it (not with a shuttle test)**
  ```
  thread 'shared_state::tests::drain_all_tl_buffers_flushes_idle_buffer' panicked at
  dial9-core/src/shared_state.rs:366:41:
  expected a batch after drain

  thread 'shared_state::tests::drain_flushes_live_worker_buffer' panicked at
  dial9-core/src/shared_state.rs:455:14:
  intrusive drain should have flushed the live worker's event
  ```

---

  **3 — reordering the epoch stamp relative to the flush**

  `dial9-core/src/shared_state.rs`, `SharedState::drain_all_tl_buffers`
  (currently lines 240-248):
  ```diff
   stats.buffers_locked += 1;
  +handle.flush_epoch.store(epoch);
   if buf.has_pending_events() {
       let batch = buf.flush();
       stats.events_flushed += batch.event_count();
       stats.buffers_flushed += 1;
       self.collector.accept_flush(batch);
   }
  -handle.flush_epoch.store(epoch);
  ```
Reordering only matters if something reads `flush_epoch` mid-drain; nothing does right now, so this is currently benign

---

  **4 — removed the stale-epoch self-flush trigger**

  `dial9-core/src/encoder.rs`, `with_encoder` (currently line 358):
  ```diff
   let current_epoch = drain_epoch.load(Ordering::Relaxed);
  -if buf.should_flush() || buf.flush_epoch.load() < current_epoch {
  +if buf.should_flush() {
       collector.accept_flush(buf.flush());
       buf.flush_epoch.store(current_epoch);
   }
  ```

It just causes a delay, not data loss. Removing it just defers the event to the intrusive drain or `Drop`, both of which still run. Not sure a test for a latency regression is worth it.

File: dial9-core/src/shared_state.rs [hunk 4]
```diff
@@ -534,3 +536,222 @@ mod tests {
...
+    const WRITERS: usize = 3;
+    const EVENTS_PER_WRITER: u64 = 3;
+    const DRAIN_TICKS: usize = 3;
+
+    crate::shuttle_test! {
+        default;
+        // Shuttle counterpart of `concurrent_record_and_drain_preserves_event_count`:
+        // exhaustively explores interleavings instead of sampling. Keeping
+        // the real-thread proptest too, for the `Ordering::Relaxed` caveat above.
+        fn shuttle_concurrent_record_and_drain() {
```
tiny-ac: catches [M5, M6](https://github.com/dial9-rs/dial9/pull/851/changes#r3895696605)

dial9-core/src/shared_state.rs
```diff
@@ -312,25 +312,27 @@ impl EventBuffer<'_> {
...
+    }
+}
+
+/// Helper: create a SharedState with recording enabled.
+#[cfg(test)]
+fn enabled_shared_state() -> SharedState {
+    let ss = SharedState::new(0);
+    ss.enable();
+    ss
+}
```
rcoh: why are these not in the test module?

dial9-core/src/shared_state.rs
```diff
@@ -534,3 +536,222 @@ mod tests {
...
+                    for _ in 0..DRAIN_TICKS {
+                        ss.bump_drain_epoch();
+                        // Grace period for an in-flight writer to self-flush
+                        // before the intrusive drain locks its buffer, mirroring
+                        // the real flush loop's tick-then-drain gap.
+                        shuttle::thread::yield_now();
+                        ss.drain_all_tl_buffers();
+                    }
+                })
+            };
```
rcoh: this is a duplicate of the actual drain logic. Is there a reason we need to duplicate this here? This logic is duplicated in several places

rcoh (commented):
I'd like to propose a thought experiment. How can we catch these sorts of mutations but in an "integration" style test that only looks at what events are flushed into what batches?

What would we need to change about the code today to make this possible? What code paths are currently only testable today via the manually powered flush loop you have? How can we make it possible to test these behaviors in a way that we are testing the exact behavior of the production code?

rcoh (approved):
Nice!!

PR #856, opened 2026-08-30
fix(viewer): expand inlined frames innermost-first

### LINKED ISSUE #855: possible bug in aggregated flamegraph rendering
Image

this stack trace is impossible; my suspicion is inlined functions are being rendered out of order. It shows "make poll start" calling "record poll start" but the opposite is the actual structure of the code.

rcoh:
Fixes #855 

## The bug

The flamegraph rendered `make_poll_start` **calling** `record_poll_start::{{closure}}`. That is impossible — `record_poll_start` calls `make_poll_start` ([`runtime_context.rs:429`](https://github.com/dial9-rs/dial9/blob/main/dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs#L429)). Every inline edge in the server-side flamegraph was upside down.

## Root cause

Stacks are **leaf→root** everywhere: perf callchains, the flat frame vectors in the stacks dictionary, and `build_flamegraph_tree`, which reverses the whole vector to walk root→leaf.

An address's *inline group* is stored the other way round, by design — blazesym reports depth 0 as the function that owns the machine code and 1..N as successively deeper inlined callees.

Both flat-chain builders expanded that group in depth order:

```
[ record_poll_start::{{closure}}, make_poll_start, <caller>, … ]
   ^ depth 0                      ^ depth 1
```

The address-level order then reversed correctly; every inline group reversed *wrongly*.

The local (non-`api`) JS flamegraph was never affected: `buildFlamegraphTree` walks the callchain end→0 and iterates each group 0→N, so it never reverses a group.

## Verification

I manually verified this fix.

### DIFF
--- a/dial9-viewer/src/ingest/decode.rs
+++ b/dial9-viewer/src/ingest/decode.rs
@@ -30,6 +30,7 @@ pub(crate) mod spans;
 mod types;
 
 use events::*;
+use lasso::{Rodeo, Spur};
 use rustc_hash::FxHashMap;
 #[cfg(test)]
 use spans::span_builder;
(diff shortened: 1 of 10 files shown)


dial9-viewer/src/ingest/decode.rs
```diff
@@ -30,6 +30,7 @@ pub(crate) mod spans;
 mod types;
 
 use events::*;
+use lasso::{Rodeo, Spur};
```
jv1i, line 33: Nice TIL about lasso

jv1i (approved):

PR #857, opened 2026-08-31
test(dial-9-core) add shuttle worker time coverage

tiny-ac:
#764 

## Summary
- Adds `primitives::time`/`shuttle_select!` (shuttle-drivable `sleep`/`sleep_until`, biased `select!`) and ports `WorkerLoop::run_triggered` onto them
 - Adds `shuttle` coverage for `WorkerLoop`'s segment-handoff and dump-resolution paths
   - `shuttle_worker_handoff` — writer/worker segment handoff via the real `run()`/`run_continuous()` path
   - `shuttle_dump_resolves_exactly_once` — an on-demand dump racing writer threads
   - `shuttle_dump_time_range_resolves_via_deadline` — only scenario giving a dump a real deadline, forcing a wait on `sleep_until`
- Splits run_background_task into a thin runtime wrapper + run_background_task_inner, now directly shuttle-drivable
    
    ## Notes
- shuttle_dump_resolves_exactly_once::determinism flaked: fixed by giving test segments a parseable header instead of one falling back to SystemTime::now()
- shuttle_dump_time_range_resolves_via_deadline::determinism was flaking ~1-in-5 from two causes, both fixed
- Dropped a should_panic scenario for a documented run_triggered deadlock: real but unreachable via the only production caller, not worth the CI cost

### DIFF
--- a/dial9-core/src/primitives.rs
+++ b/dial9-core/src/primitives.rs
@@ -186,21 +186,19 @@ pub mod time {
     }
 }
 
-/// `tokio::select!` normally; `shuttle_tokio::select!` under `--cfg shuttle`.
+/// `tokio::select!` normally; `shuttle_tokio_impl_inner::select!` under
+/// `--cfg shuttle`, which patches `select!`'s branch tie-break to draw from
+/// `shuttle::rand::thread_rng()` instead of real OS entropy, so a replayed
+/// schedule picks the same branch every time.
 ///
-/// `select!`'s branch tie-break otherwise calls tokio's own `thread_rng_n`,
-/// seeded from real OS entropy per thread -- not shuttle-controlled, so a
-/// replayed schedule can pick a different branch. `shuttle-tokio` patches
-/// that one function to draw from `shuttle::rand::thread_rng()` instead.
-///
-/// Doesn't cover `sleep`/`sleep_until`: `shuttle-tokio`'s `Sleep` resolves
-/// immediately on first poll for any realistic deadline rather than
-/// suspending, so `primitives::time` keeps its own `Yield` for those.
+/// Doesn't cover `sleep`/`sleep_until`: its `Sleep` resolves immediately on
+/// first poll instead of suspending, so `primitives::time` keeps its own
+/// `Yield` for those.
 #[cfg(all(shuttle, feature = "pipeline"))]
 #[macro_export]
 macro_rules! shuttle_select {
     ($($arms:tt)*) => {
-        shuttle_tokio::select! { $($arms)* }
+        shuttle_tokio_impl_inner::select! { $($arms)* }
     };
 }
 #[cfg(all(not(shuttle), feature = "pipeline"))]
--- a/dial9-core/src/primitives.rs
+++ b/dial9-core/src/primitives.rs
@@ -45,6 +45,10 @@ pub use crate::define_thread_local as thread_local;
 #[cfg(all(not(shuttle), feature = "pipeline"))]
 pub mod time {
     pub use tokio::time::{Instant, sleep, sleep_until};
+
+    pub fn now() -> Instant {
+        Instant::now()
+    }
 }
 
 // ── shuttle path (deterministic testing) ────────────────────────────────────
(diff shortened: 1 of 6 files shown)


dial9-core/src/primitives.rs
```diff
@@ -127,6 +132,87 @@ macro_rules! define_thread_local {
 #[cfg(shuttle)]
 pub use crate::define_thread_local as thread_local;
 
+#[cfg(all(shuttle, feature = "pipeline"))]
```
rcoh, line 143: there is `shuttle-tokio` which implements tokio primitives properly with shuttle. should we use that intead?

rcoh: if you look at the source of `shuttle-tokio` it is https://docs.rs/shuttle-tokio/latest/src/shuttle_tokio/lib.rs.html#1-29 so for this pattern, we probably want to depend on shuttle_tokio_impl_inner instead.

dial9-core/src/primitives.rs
```diff
@@ -127,6 +132,85 @@ macro_rules! define_thread_local {
...
+    use std::future::Future;
+    use std::pin::Pin;
+    use std::sync::atomic::{AtomicUsize, Ordering};
+    use std::task::{Context, Poll};
+
+    pub use tokio::time::Instant;
+
+    /// Shuttle has no virtual clock. Like `primitives::thread::sleep`, this
+    /// is a single scheduling point, not a real delay.
+    pub fn sleep(_duration: std::time::Duration) -> Yield {
```
rcoh, line 179: shuttle tokio defines time wrappers: https://docs.rs/shuttle-tokio-impl-inner/latest/src/shuttle_tokio_impl_inner/time.rs.html#230-235

why did we re-implement something manually here?

dial9-core/src/worker/shuttle_tests.rs
```diff
@@ -0,0 +1,366 @@
...
+            "dump must not double-dispatch a segment to the processor"
+        );
+    }
+}
+
+/// Must be nonzero so `ActiveDump::register` sets a real `deadline`. The
+/// value itself doesn't matter, shuttle's `sleep_until` ignores it.
+const LOOKFORWARD: Duration = Duration::from_millis(1);
+
+// `determinism` SIGBUSes here because shuttle's
```
rcoh: this isn't true. The nondeterminism is from the use of `Instant::now`

dial9-core/src/primitives.rs
```diff
@@ -127,6 +132,85 @@ macro_rules! define_thread_local {
...
 pub use crate::define_thread_local as thread_local;
 
+#[cfg(all(shuttle, feature = "pipeline"))]
+pub mod time {
+    use std::future::Future;
+    use std::pin::Pin;
+    use std::sync::atomic::{AtomicUsize, Ordering};
+    use std::task::{Context, Poll};
+
+    pub use tokio::time::Instant;
```
rcoh, line 151: this should be re-exporting shuttle_tokio_impl_inner; the use of `tokio::time::Instant::now` creates non determinism.

rcoh (changes requested):

dial9-core/src/primitives.rs
tiny-ac: I tried replacing it, it just [re-exports the same tokio::time::Instant](https://github.com/awslabs/shuttle/blob/main/wrappers/tokio/impls/tokio/inner/src/time.rs#L11), so same issue arises.
image



image

dial9-core/src/primitives.rs
tiny-ac: [Shuttle's implementation](https://github.com/awslabs/shuttle/blob/main/wrappers/tokio/impls/tokio/inner/src/time.rs#L224-L254) doesn't put the future through the pending state, it resolves immediately on first check (unless the deadline is a year or more away).

tiny-ac: [The loop that's supposed to wait ](https://github.com/dial9-rs/dial9/blob/main/dial9-core/src/worker/mod.rs#L496-L497)on it doesn't wait, so the loop goes around and around, reaching a "exceeded max_steps" error 
image

dial9-core/src/primitives.rs
rcoh: I see...perhaps we can make Instant::now increment a counter when it is called? (Basically make it deterministic)

rcoh: I see, you've done this below, great!

dial9-core/src/primitives.rs
rcoh: yikes OK.

rcoh (approved):

Issue #843, opened 2026-08-28
Add task-scoped mixed CPU/async flamegraph


noxware:
Implement the task-scoped mixed flamegraph described in `docs/design/task-dump-capture-sampling.md`.

Includes grouping sibling callchains and representative-stack selection.

PR #859, opened 2026-08-31
test(viewer): add self-describing telemetry integration coverage

### LINKED ISSUE #841: Self-describing telemetry integration test app
Implement the test app described in `docs/design/telemetry-integration-test-app.md`.

noxware:
## Changes

- Add a self-describing telemetry example.
- Add JS E2E validation.

Test runs as part of the Linux CI when `cargo test --all-targets --all-features` happens.

## Deferred until #843

- Add task dumps to aggregate Parquet output (probably means re-adding https://github.com/noxware/dial9/commit/8844549676b4ec39d5db59b8ca9ca95f2f42ecd9 and bumping cache version)
- Add Rust/Parquet E2E validation (through a mixed API introduced for the flamegraph)

Closes #841


dial9-viewer/tests/telemetry_test_app/aggregate.rs
```diff
@@ -0,0 +1,318 @@
...
+use arrow::array::{
+    Array as _, FixedSizeBinaryArray, ListArray, MapArray, StringArray, StructArray, UInt8Array,
+};
+use dial9_viewer::ingest::TelemetryTestParquetParts;
+use parquet::arrow::arrow_reader::ParquetRecordBatchReader;
+use std::collections::{BTreeMap, BTreeSet, HashMap};
+
+type StackId = [u8; 16];
+
+pub(crate) fn observe_aggregate_trace(segments: &[Vec<u8>]) -> Result<Observations> {
```
rcoh: what if we ran the viewer then hit API endpoints on the viewer? That would avoid duplicating this code and would also ensure a real E2E test

noxware: I agree. I will adopt that idea. Rust/parquet was the last thing done, so I'm still correcting it.

noxware: I'm reconsidering.

We may want to cut `8844549`, `2a0cd89` and `9da2a9f` out of this PR and add Rust/Parquet checks on #843, so we don't add API changes here just for the test if the API will be naturally introduced with the mixed flamegraph. 

I would just leave here the test app, JS check and CI step, no duplication or wrong e2e.

dial9-viewer/tests/telemetry_test_app/expectations.rs
```diff
@@ -0,0 +1,638 @@
...
+use dial9_trace_format::decoder::Decoder;
+use serde::{Deserialize, Deserializer, de};
+use std::{collections::BTreeSet, num::NonZeroU64};
+
+const EXPECTATION_EVENT: &str = "TelemetryFixtureExpectationEvent";
+const MARKER_EVENT: &str = "TelemetryFixtureMarkerEvent";
+const FIXTURE_PREFIX: &str = "dial9_fixture_";
+const WEIGHT_SEPARATOR: &str = "_weight_";
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
```
rcoh, line 11: is there a reason this doesn't derive(Deserialize)?

dial9-viewer/tests/telemetry_test_app/expectations.rs
```diff
@@ -0,0 +1,638 @@
+use anyhow::{Context as _, Result, bail, ensure};
+use dial9_trace_format::decoder::Decoder;
+use serde::{Deserialize, Deserializer, de};
+use std::{collections::BTreeSet, num::NonZeroU64};
+
+const EXPECTATION_EVENT: &str = "TelemetryFixtureExpectationEvent";
+const MARKER_EVENT: &str = "TelemetryFixtureMarkerEvent";
+const FIXTURE_PREFIX: &str = "dial9_fixture_";
+const WEIGHT_SEPARATOR: &str = "_weight_";
```
rcoh, line 9: these deserialize impls can all be derived

dial9-viewer/tests/telemetry_test_app/expectations.rs
```diff
@@ -0,0 +1,638 @@
+use anyhow::{Context as _, Result, bail, ensure};
+use dial9_trace_format::decoder::Decoder;
+use serde::{Deserialize, Deserializer, de};
+use std::{collections::BTreeSet, num::NonZeroU64};
+
+const EXPECTATION_EVENT: &str = "TelemetryFixtureExpectationEvent";
```
rcoh, line 6: this file would really be helped by a doc explaining what the structure is

examples/telemetry-test-app/src/main.rs
```diff
@@ -0,0 +1,242 @@
...
+
+const DEFAULT_CYCLES: u64 = 40;
+const WARMUP_CYCLES: u64 = 4;
+const CPU_QUANTUM: Duration = Duration::from_millis(10);
+const WAIT_QUANTUM: Duration = Duration::from_millis(10);
+const MAX_TRACE_SIZE: u64 = 100_000_000;
+
+const MIXED_CYCLE: &str = "dial9_fixture_mixed_cycle";
+const MIXED_INNER: &str = "dial9_fixture_mixed_inner";
+const CPU_OUTER: &str = "dial9_fixture_cpu_outer_weight_1";
```
rcoh, line 36: are these weight names load bearing?

examples/telemetry-test-app/src/main.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+async fn run_cycles(cycles: u64) {
+    for cycle in 0..cycles {
+        dial9_fixture_mixed_cycle(cycle).await;
+    }
+}
+
+#[inline(never)]
+async fn dial9_fixture_mixed_cycle(cycle: u64) {
```
rcoh, line 182: it would be good to mix inline with non-inline

rcoh (approved):
seems like a good start

examples/telemetry-test-app/src/main.rs
noxware: Sorry, I forgot to test https://github.com/dial9-rs/dial9/pull/856. This cover that, right?

examples/telemetry-test-app/src/main.rs
noxware: Weights are parsed, and there are relative comparisons like "the cpu function with more weight has more samples than the cpu function with less weight".

Checking the exact ratios is not currently implemented. I simplified this way due to the test app doc:

```markdown
The initial integration test needs only presence, hierarchy, and coarse
relative-weight assertions. Task-dump mixed-flamegraph tests can later consume
the same names to check the `4:3` whole-cycle and `3:2` inner-subtree mixes.
```

```markdown
### Initial assertions

...

- the long CPU function has more samples than the short CPU function after a
  minimum total sample count;
  
...
  
- at least one task dump contains each wait symbol; and
```

(i don't know if i'm answering your question)

dial9-viewer/tests/telemetry_test_app/expectations.rs
noxware: I annotated `FunctionSymbol` and `SpanName` with `Deserialize`.

But `FixtureFeature` and `MarkerPhase` cannot use a direct derive with the current field deserializer.

```rust
Err(DeserError("invalid type: string \"cpu\", expected enum FixtureFeature"))
```

`CpuSampleSource` in `analysis_events.rs` follows the same manual deserialize pattern.

PR #873, opened 2026-09-03
test: add shuttle coverage for concurrent Tokio runtime attach

tiny-ac:
#764
## Description
- Adds shuttle coverage for RuntimeContextRegistry's concurrent-attach path.
  - Several threads each call attach_tokio_runtime at once, racing against the flush thread
  - Asserts that every concurrent attach is observed exactly once and under its own name
- Adds runtime_handle_spawn_with_targets_correct_runtime_worker_ids (tests/spawn_with.rs).
  - Verifies two attached runtimes' polls resolve to disjoint worker ids

## Notes
- Doesn't check bind_runtime's effect directly. Covered instead by the new spawn_with.rs test


dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -1060,3 +1068,118 @@ mod tests {
...
+        fn shuttle_concurrent_attach() {
+            const ATTACHERS: usize = 3;
+            let shared = Arc::new(SharedState::new(0));
+            shared.enable();
+
+            let attachers: Vec<_> = (0..ATTACHERS)
+                .map(|i| {
+                    let shared = shared.clone();
+                    crate::primitives::thread::spawn(move || {
+                        let handle = dial9_core::test_util::connected_handle(shared);
```
rcoh: why can't we use the regular public API here?

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -1060,3 +1068,118 @@ mod tests {
...
+                        let options = TokioAttachOptions::builder()
+                            .runtime_name(format!("runtime-{i}"))
+                            .build();
+                        let runtime = handle.attach_tokio_runtime(builder, options).unwrap();
+                        let id = runtime.handle().id();
+                        drop(runtime);
+
+                        let state = recorder_tokio::tokio_attach_state(&handle)
+                            .expect("attach_tokio_runtime installed the source above");
+                        let registry = state.registry.lock().unwrap();
```
rcoh: can this test observe some behavior on the generated trace file instead of reaching into the internal API? By observing the emitted trace file (e.g. that known task ids are showing polls on the correctly mapped runtimes, that the reserved worker ids in the trace files are correct etc. etc.) we can have a test that is not brittle to the implementation details and instead tests the observable behavior.

That makes these tests more useful when internals are changed.

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -1060,3 +1060,123 @@ mod tests {
...
+            // Fixed time, like `test_core_pipeline`: otherwise the flush
+            // thread's rotation checks read real elapsed time, which can
+            // differ between a shuttle replay's record and re-run passes.
+            let _ts_guard = metrique_timesource::set_time_source(metrique_timesource::TimeSource::custom(
+                metrique_timesource::fakes::StaticTimeSource::at_time(std::time::UNIX_EPOCH),
+            ));
+
+            const ATTACHERS: usize = 3;
+            let writer = dial9_core::buffer::MemoryBuffer::new(1 << 16).unwrap();
+            let sealed = dial9_core::test_util::writer_sealed_segments(&writer);
```
rcoh, line 1095: instead of exposing a test util hook, could we add a pipeline stage that captures the segments at the end?

In general, I prefer that these tests test as close to observed behavior as possible.

dial9-core/src/test_util.rs
```diff
@@ -86,6 +87,47 @@ mod pipeline_helpers {
         DumpId::new()
     }
 
+    /// Opaque handle to `writer`'s in-memory sealed-segment store. Capture
+    /// this before `writer` moves into a `Recorder`/`RecorderBuilder`, so a
+    /// sibling-crate test can later inspect the real trace via
```
rcoh, line 92: this isn't actually the real trace since it didn't go through the pipeline / worker though?

```rust
let (capture, captured) = CapturingProcessor::new();
let recorder = dial9_core::recorder::recorder(writer)
    .pipe(capture)
    .build();

// test workload...

recorder.graceful_shutdown(Duration::from_secs(1));
let segments = captured.lock().unwrap();
```

rcoh (commented):
remove the extra pub function, go through the capture stage

dial9-core/src/primitives.rs
```diff
@@ -219,6 +220,42 @@ pub mod time {
...
+    #[derive(Debug)]
+    pub struct Elapsed(());
+
+    pub struct Timeout<F> {
+        future: F,
+    }
+
+    // Per-pending-poll odds of simulating the deadline. Low enough that a
+    // handful of polls (a typical drain) is very unlikely to get cut off.
+    const FIRE_PROBABILITY_PER_PENDING_POLL: f64 = 0.02;
```
rcoh, line 241: how was this chosen?

rcoh (approved):

dial9-core/src/primitives.rs
tiny-ac: This ended up being a placeholder since the branch that uses this value isn't being exercised right now. Will adjust as part of the cleanup follow-up PR

PR #876, opened 2026-09-03
feat(viewer): add off-CPU points of interest

### LINKED ISSUE #861: feat(viewer): add "off CPU polls / worker active periods" as a point-of-interest for the viewer

Fluzko:
closes #861
## Summary

Adds two new filters in the issues rail:

- Descheduled Worker Periods so it stretches where a worker was awake but kernel kept it off the CPU, ranked by how much running time was lost. Shows as ticks on the minimap.
- Off CPU-Polls that ran long without ever being on the CPU, meaning the task blocked synchronously instead of yielding.

The first signal already exists in the agent toolkit as the `cpu-contention` red flag, so an agent could tell you a worker was being descheduled while someone reading the timeline had no way to find those windows. The second is the case the `dial9-diagnose-long-poll` skill currently walks through by hand.


dial9-viewer/ui/src/pages/viewer/poi.ts
```diff
@@ -77,6 +78,8 @@ export function filterLabel(type: PointOfInterestType): string {
       // No threshold in the label, unlike its fixed-threshold siblings: the
       // rail renders the live value in its own input.
       return "Spawn->First Poll Delays";
+    case "off-cpu-active":
+      return "Descheduled Worker Periods (CPU <50%)";
```
rcoh: lets just show the top N here instead

Fluzko: Removed the <50% filter, and the >1ms one next to it since it was hiding short periods that were fully de scheduled. Ranked by off-CPU time now.

Did the same for the other five detectors, and added a worst 10/50/200/all selector.

Two things:
- I skip anything with zero severity, without a cut-off every park qualified, and most have no wait at all. Thoughts?
- Counts no longer mean much, since with no cut-off, a detector's count is just how many things it looked at. The toolbar chip shows the worst value instead.

rcoh (changes requested):

rcoh:
this came out awesome. thanks for iterating

rcoh (approved):

PR #878, opened 2026-09-04
test(dial9-utils): add shuttle coverage for the tracing-layer schema cache

tiny-ac:
#764 
## Description
- Adds shuttle coverage for Dial9TracingLayer::get_schemas's per-callsite schema cache (dial9-utils/src/tracing_layer.rs).
  - Several threads enter the same span callsite at once
  - Asserts the cache builds that callsite's schema exactly once
- Adds get_schemas_is_keyed_per_callsite, a small focused unit test covering the one property the shuttle scenario can't
- Fixes get_schemas_is_keyed_per_callsite panicking under
- Wires dial9-utils's shuttle tests into scripts/test-shuttle.sh.

## Notes
- SCHEMA_BUILD_COUNT/SCHEMA_BUILD_COUNT_LOCK are plain (non-shuttle) statics. The counter must survive shuttle's own per-iteration state reset
- cargo fmt --check/cargo clippy clean on dial9-utils, both with and without --cfg shuttle.


dial9-utils/src/tracing_layer.rs
```diff
@@ -98,8 +98,33 @@ struct CallsiteSchemas {
...
+fn take_schema_build_count() -> usize {
+    SCHEMA_BUILD_COUNT.swap(0, std::sync::atomic::Ordering::SeqCst)
+}
+
+/// Serializes access to `SCHEMA_BUILD_COUNT` across `pct`, `determinism`,
+/// and `get_schemas_is_keyed_per_callsite` (all bump it via `get_schemas`),
+/// which `cargo test` runs concurrently. Plain `std::sync::Mutex`: only
+/// keeps those test threads apart, not shuttle's own interleavings.
+#[cfg(all(test, shuttle))]
+static SCHEMA_BUILD_COUNT_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
```
rcoh, line 121: in general, lets not add this sort of code into the production code. `callsite_schemas` / `CallsiteSchemas` is a private function / private type so we have some options here:

1. Return `(CallSiteSchemas, bool)` indicating if it came from cache
2. (a good change anyway) change the hashmap to store Arc<CallsiteSchemas> which would make the clone cheaper (by having it be a single arc bump) and also expose the strong_count (so you could see if it was actually cloned)

rcoh: I did a quick benchmark of the Arc change, it seems like an improvement

tiny-ac: Ended up racing `get_schemas` directly instead of calling `info_span!`. Internals that use plain `std::thread_local!` (dispatcher default, per-layer filters, sharded-slab's shard allocator) make the test fail under shuttle. They assume there's one OS thread per logical thread, which isn't the case under shuttle.

rcoh (approved):

PR #891, opened 2026-09-08
fix(core): stop rotation stranding events between files

### LINKED ISSUE #866: bug(analysis): partial files confound analysis
There is a class of bug where the same wall-clock period crosses multiple files. This really confounds analysis tooling because threads can be totally invisible; when a thread is invisible, it looks like the events didn't happen.

There are a few mitigations we should think about:
- [ ] adaptive file rotation period (with lower / upper bounds)
- [ ] add segment metadata to files when they are flushed that indicates whether they were flushed "cleanly" or "dirty" (note that a file is only clean if the file _before_ it was also flushed cleanly)
- [ ] consider filtering analysis algorithms to either run on clean files only or to look at files in groups of three

jv1i:
Fixes #866.

Whenever a file is missing events analysis would just miss data without knowing that a segment was incomplete, the causes are:
- a file that crosses `max_file_size` is sealed mid-write without draining thread-local buffers, so it loses whatever was still buffered and the next file picks those up.
- the drain interval was pinned to the configured rotation period, so once files started rotating on size instead of the clock, buffers were going stale relative to the file they landed in.

## Changes

`rotation_period` is a ceiling now, each rotation estimates the byte rate from the file it just closed and shortens the period so the next one lands inside `max_file_size`, floored at 5s. The drain interval follows the shortened period, so buffers get emptied at the rate files are actually filling.

The estimate aims a bit under `max_file_size` so a period that runs long still rotates on the clock, but files can still overrun it and seal without a drain, which is what the flags below are for. `test_adaptive_period_converges_off_the_size_path` covers this: the early files overrun, then the shortened period keeps the rest on the clock.

Each file is now stamped at seal time with `segment.sealed_clean` and `segment.prior_sealed_clean`. Two flags because a bad seal leaves one file short and the next one holding its strays, so a file has to answer both "did I lose events" and "did I inherit someone else's" (the second one is technically duplicated info, but in purpose so a file can be judged without fetching its predecessor).

`TraceReader::is_complete()` reads both, `None` for traces written before this. It also merges segment metadata now, since a file carries one event when it opens and another when it is sealed.

## Convergence

Adaptation can take a few rotations to settle, an overflow seal doesn't drain, so the next drain dumps the backlog and overflows again until the estimate catches up.

## No viewer changes

As I understand it we don't want to drop segments from the viewer, especially not on "dirty" ones where probably the heaviest/richest load of an app is happening. Also the viewer concatenates the files before parsing so the boundary polls are there anyway.

## Not covered

A poll starting in one file and ending in the next is still dropped from both. I think only reading neighbouring files fixes that. We have `DecodeStats::polls_end_unproven` which already counts polls whose end nothing proved, but nothing reads it, so we should follow up with something that makes use of it. Might be mostly rotation cuts, though a second `PollStart` with no end counts too.


dial9-core/src/metrics.rs
```diff
@@ -46,6 +46,17 @@ pub(crate) struct FlushStats {
     pub cpu_flush_duration: Duration,
 }
 
+/// Running totals of how segments were sealed, since the writer started.
+#[metrics(subfield, rename_all = "PascalCase")]
```
rcoh: Don't need rename on subfield

rcoh (approved):

PR #897, opened 2026-09-10
fix(dial9-core): contain panicking Source implementations

### LINKED ISSUE #783: `SharedState::flush_sources()` has no panic containment around `Source::flush`
Found while working on https://github.com/dial9-rs/dial9/pull/781

### Summary

`SharedState::flush_sources()` ([dial9-core/src/shared_state.rs:273](https://github.com/dial9-rs/dial9/blob/fa365f80674e4974670be12fbdfd0189a601ba16/dial9-core/src/shared_state.rs#L273-L279)) calls every registered `Source`'s `flush` method with no panic guard.
```rust
pub fn flush_sources(&self) {
    let ctx = self.flush_context();
    let mut sources = self.sources.lock().unwrap();
    for source in sources.iter_mut() {
        source.flush(&ctx);   // <-- no catch_unwind
    }
}
```

## Impact
A panic inside any Source::flush unwinds straight through the flush thread and kills it permanently (no more draining, TL buffer, collector for the rest of the process's life, and teardown() never runs). Recorder::stop_flush_thread() discards the join result and the caller never learns recording died. The failure is `silent`.

Compare with `worker::process_segments` ([dial9-core/src/worker/mod.rs](https://github.com/dial9-rs/dial9/blob/fa365f80674e4974670be12fbdfd0189a601ba16/dial9-core/src/worker/mod.rs#L765-L770)), which does implement a `catch_unwind` guard

tiny-ac:
#764 
closes #783
## Description
- Fixes the flush thread's teardown being skipped when a `Source::flush` panics
- Extends the same fix to `Source::segment_metadata`
- Adds a `# Panic safety` contract to the `Source` trait, matching `SegmentProcessor`'s existing one
- Un-skips pre-existing shuttle tests in pipeline_shuttle_tests.rs
- Extracts shared test-only decode helpers into a new `test_support` module

## Notes
- `cargo fmt --check`/clippy clean on dial9-core, both with and without `--cfg shuttle`
- `cargo nextest run -p dial9-core`: 225/225. Shuttle suite: 31/31 (`pct`+`determinism`)


dial9-core/src/flush_loop.rs
```diff
@@ -142,7 +142,18 @@ pub(crate) fn run_flush_loop<M: BufferMode>(
         {
             let mut sources = shared.sources.lock().unwrap();
             for source in sources.iter_mut() {
-                source.segment_metadata(&mut source_entries);
+                // Truncate back to the pre-call length on panic so a partial
```
rcoh, line 157: Should we leave a metadata entry that says the source panicked? Could be helpful for debugging when entries are missing

dial9-core/src/pipeline_shuttle_tests.rs
```diff
@@ -146,22 +147,15 @@ crate::shuttle_test! {
...
         let source_pending: Arc<Mutex<Vec<ValidationEvent>>> = Arc::new(Mutex::new(Vec::new()));
 
-        let shared = Arc::new(SharedState::new(clock_monotonic_ns()));
-        shared.push_source(Box::new(MockSource::new(source_pending.clone())));
-        let mut recorder = Recorder::start(shared, writer, None, || || {});
-        recorder.handle().enable();
+        // Small segments force frequent rotation: the 100 MiB budget (set by
+        // `start_shuttle_memory_recorder`) is far above the test's data, so
+        // the ring never evicts before we drain it.
+        let (mut recorder, fs) = crate::test_support::start_shuttle_memory_recorder(vec![Box::new(
```
rcoh, line 156: Nice refactoring!

dial9-core/src/recording.rs
```diff
@@ -289,3 +289,280 @@ impl Drop for Recorder {
...
+                    warned_sources: subscriber_source.clone(),
+                });
+                move || drop(guard)
+            })
+            .build();
+        recorder.handle().enable();
+
+        // Give the flush thread time to run at least one cycle: both
+        // sources panic on their very first flush.
+        std::thread::sleep(Duration::from_millis(200));
```
rcoh: Unfortunately you can't do things like sleep in tests. It is inherently flaky. You will need to find a way either:
- use shuttle to test many orderings and assert an invariant (possibly with a check to ensure you cover the ordering you are interested in)
- make the test deterministic by finding a way to deterministically wait for the condition you expect

dial9-core/src/test_support.rs
```diff
@@ -0,0 +1,103 @@
+//! Shared helpers for this crate's own `#[cfg(test)]` unit tests.
+//!
+//! Distinct from `test_util`, which is `pub` under the `test-util` feature
+//! for sibling-crate tests; this module is crate-internal and feature-free,
+//! so it's always available wherever `#[cfg(test)]` is.
```
rcoh, line 5: You might want a root module cfg gate then using ! to express that it applies to the parent. Otherwise this is an empty module in prod

dial9-core/src/test_support.rs
```diff
@@ -3,13 +3,21 @@
...
 
+use crate::primitives::sync::Mutex;
 use dial9_trace_format::decoder::{DecodedFrameRef, Decoder};
 use dial9_trace_format::types::FieldValueRef;
 
+/// Serializes tests that build a real `Recorder` via the public builder:
+/// `SoleRecorderGuard` (recorder.rs) allows only one per process, so
+/// concurrent test threads would otherwise race for it and silently get a
+/// disabled one.
+pub(crate) static SOLE_RECORDER_LOCK: Mutex<()> = Mutex::new(());
```
rcoh: I would prefer we find a different way to handle this (probably allowing multiple recorders per process in tests?)

jv1i: Regarding this, @tiny-ac I saw the last commit but I think we can avoid adding that new test-specific API, I might be missing something here but I ran into something similar before

Under test-util, multi-recorders per process are allowed (SoleRecorderGuard::claim() always succeeds, see #810). As it is now that wouldn't work on the test-util-off job (`cargo test -p dial9-core --lib`).

Could we just scope that job to the test that needs the guard? e.g. `cargo test -p dial9-core --lib single_recorder_tests`, then we might not need `build_for_test`/`build_inner` and the tests just go through `build()`

dial9-core/src/shared_state.rs
```diff
@@ -273,7 +273,11 @@ impl SharedState {
             let ctx = self.flush_context();
             let mut sources = self.sources.lock().unwrap();
             for source in sources.iter_mut() {
-                source.flush(&ctx);
+                crate::source::catch_source_panic(
+                    source.as_mut(),
+                    crate::source::SourceCall::Flush,
+                    |source| source.flush(&ctx),
+                );
```
jv1i, line 284: should we record the `dial9.source.<name>.panicked` marker here too? to warn of missing events

jv1i (commented):
Previous comments aside, overall LGTM.
I think there's a few other places that could use panic protection (fine to do as follow-ups):   
- `with_encoder`: a panic poisons the flush thread's buffer (see `with_encoder` in `encoder.rs`) and every source stops recording after it
- same for `on_thread_start` / `on_thread_stop` and the `with_source` closures (panic poisons the `sources` lock)

PR #899, opened 2026-09-10
fix(taskdump): remove unreleased capture changes

noxware:
## Changes

- Restore `just_captured` logic for capture skips.
- Test against capture loops.
- Recover `1.52.0` as the minimum Tokio version (as `1.53.0` is not strictly needed anymore, and that Cargo.toml version would introduce https://github.com/tokio-rs/tokio/issues/8434 for people currently locked at 1.52 that would be unaffected otherwise).
- Remove these changes from the `Unreleased` section in the changelog.

### DIFF
--- a/dial9-tokio-telemetry/Cargo.toml
+++ b/dial9-tokio-telemetry/Cargo.toml
@@ -24,7 +24,7 @@ unexpected_cfgs = { level = "warn", check-cfg = [
 missing_debug_implementations = "warn"
 
 [dependencies]
-tokio = { version = "1.53.0", features = [
+tokio = { version = "1.52.0", features = [
     "rt",
     "macros",
     "time",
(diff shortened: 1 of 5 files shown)


dial9-tokio-telemetry/Cargo.toml
```diff
@@ -24,7 +24,7 @@ unexpected_cfgs = { level = "warn", check-cfg = [
 missing_debug_implementations = "warn"
 
 [dependencies]
-tokio = { version = "1.53.0", features = [
+tokio = { version = "1.52.0", features = [
```
noxware, line 27: Once the bug disappears from Tokio in a stable version, we may want to bump this to a minimum that has the fix included to prevent people from using the problematic version by accident. 

We could also backport that min version to 0.3.x.

rcoh (approved):

PR #901, opened 2026-09-11
feat(cpu): add system info as segement metadata

### LINKED ISSUE #879: emit more default information in perf-self-profile
The CPU profiler should probably include segment metdata like kernel version, CPU, CPU architecture, total memory etc. etc.

jv1i:
Closes #879

Adds some linux and android system information to segment metadata in cpu profiling. Metadata that cannot be read or parsed is logged and omitted from the trace.

New fields:

| Field | Source | Example from my Docker setup | Example from Android |
| --- | --- | --- | --- |
| cpu.profile.cpu_model | /proc/cpuinfo (`model name`, `Processor`, or `Hardware`) | Skipped, arm64 kernels don't expose a CPU model name in /proc/cpuinfo  | <- same |
| cpu.profile.kernel_release | uname.release | `6.10.14-linuxkit` | `5.15.123-android13-3-28579584` |
| cpu.profile.kernel_version | uname.version | `#1 SMP Sat May 17 08:28:57 UTC 2025` | `#1 SMP PREEMPT Fri Jul 5 17:55:35 KST 2024` |
| cpu.profile.machine_hostname | uname.nodename | `6830cf53fb83` | `localhost` |
| cpu.profile.machine_architecture | uname.machine | `aarch64` | `aarch64` |
| cpu.profile.memory_total_bytes | /proc/meminfo (`MemTotal`) | `13584277504` | `5691387904` |

Also updates the android example build script (dial9_tokio_telemetry -> dial9) and fixes the trace reading in `cpu_profile_workload_example`

### DIFF
--- a/perf-self-profile/src/sys/linux/system_metadata.rs
+++ b/perf-self-profile/src/sys/linux/system_metadata.rs
@@ -1,0 +1,174 @@
+use std::ffi::CStr;
+use std::io::{self, ErrorKind};
+use std::mem::MaybeUninit;
+use std::sync::OnceLock;
+
+pub(crate) fn system_metadata() -> Vec<(String, String)> {
+    static SYSTEM_METADATA: OnceLock<Vec<(String, String)>> = OnceLock::new();
+
+    SYSTEM_METADATA.get_or_init(collect_system_metadata).clone()
+}
+
+fn collect_system_metadata() -> Vec<(String, String)> {
+    let mut out = Vec::new();
+
+    match uname_metadata() {
+        Ok(entries) => out.extend(entries),
+        Err(e) => tracing::warn!("failed to read uname metadata: {e}"),
+    }
+
+    match std::fs::read_to_string("/proc/cpuinfo").and_then(|value| {
+        parse_cpu_model(&value).ok_or_else(|| {
+            io::Error::new(
+                ErrorKind::InvalidData,
+                "no supported CPU model field in /proc/cpuinfo",
+            )
+        })
+    }) {
+        Ok(model) => out.push(("cpu.profile.cpu_model".to_string(), model)),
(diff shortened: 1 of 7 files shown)


perf-self-profile/src/sys/linux/system_metadata.rs
```diff
@@ -0,0 +1,174 @@
...
+        (
+            "cpu.profile.kernel_version".to_string(),
+            uname_field(&utsname.version)?,
+        ),
+    ])
+}
+
+fn uname_field(field: &[libc::c_char]) -> io::Result<String> {
+    // SAFETY: fields populated by uname are null-terminated C strings.
+    let value = unsafe { CStr::from_ptr(field.as_ptr()) };
```
rcoh, line 84: `CStr::from_ptr` requires a null terminated string; we should validate that here defensively

perf-self-profile/src/sys/linux/system_metadata.rs
```diff
@@ -0,0 +1,174 @@
...
+    if rc != 0 {
+        return Err(io::Error::last_os_error());
+    }
+    // SAFETY: uname returned success and initialized `utsname`.
+    let utsname = unsafe { utsname.assume_init() };
+
+    Ok([
+        (
+            "cpu.profile.machine_hostname".to_string(),
+            uname_field(&utsname.nodename)?,
```
rcoh: if one of these fails, we'll bail out of the entire function; wdty about setting these to `error` (or ideally a decent error message) if they ever fail?

rcoh (approved):

perf-self-profile/src/sys/linux/system_metadata.rs
jv1i: definetly! I had this doing that properly but it regressed when moving it from cpu_source to this new file

PR #905, opened 2026-09-15
perf(memory-profiling): skip bucket reservation on dealloc miss

seroze:
## What

`on_dealloc` resolved every deallocation through `liveset.entry(addr)`. On a miss that is not free: `entry()` takes the per-bucket write lock and reserves a vacant entry before the caller can observe that the address was never in the map. The liveset holds only *sampled* allocations — the hook's own docs put the filter rate at ~99.9% of deallocs — so nearly every call pays for a reservation it immediately discards.

This takes a lock-free `peek_with` first and returns early on a miss.

## Correctness

Hits still resolve through `entry()`, so the invariant documented on the original line is preserved: the bucket lock is held across the read and the remove, so a concurrent `on_alloc` overwriting the same address cannot interleave between them and put stale metadata on the emitted `RawFree`.

The added peek does not widen any race:

- **Miss at peek** — the address was never sampled, so there is nothing to emit. A concurrent `on_alloc` cannot be inserting this same address, since the allocator does not hand out a live address twice.
- **Hit at peek, vacant at `entry()`** — a concurrent dealloc of the same address or the shutdown drain got there first; `entry()` returns `Vacant` and we emit nothing, exactly as before.

The `SAMPLE_STATE` re-entrancy guard still wraps both calls, and `check_shutdown()` still runs before any liveset op, so the `sdd` TLS init ordering is unchanged.

This mirrors the structure already used in `on_alloc`, which keeps its hot path on the lock-free `insert` and pays the bucket-lock cost only in the rare stale-entry case.

## Testing

- `cargo test -p dial9-perf-self-profile --features memory-profiling --lib` → **119 passed, 0 failed**, including all 32 `memory_profiling` tests.
- Added `on_dealloc_ignores_address_that_was_never_sampled`, pinning the new miss branch: no `RawFree` pushed, no entry left behind.
- Negative control: forcing the fast path to always bail makes `on_alloc_overwrites_stale_liveset_entry` fail, confirming the hit path is actually covered rather than passing vacuously.
- `cargo clippy --features memory-profiling --all-targets` clean; `cargo fmt --check` clean.

Two tests in `tests/sched_triggers.rs` (`captures_sleep_stack`, `captures_lock_acquisition_stack`) fail in my environment with `perf_event_open blocked (Permission denied)` — WSL2 with `perf_event_paranoid=2`. They fail identically on unmodified `main`, so they are unrelated to this change.

## Performance

I did not have a representative allocator workload to measure end to end, so treat this as directional. A standalone microbenchmark of the two `scc` call shapes (2.4.0, `FxBuildHasher`, 20K live entries, all-miss probes) gives:

```
entry() only        : 92.17 ns/op
peek_with() only    : 39.59 ns/op
peek_with + entry() : 39.74 ns/op
```

Absolute numbers swing a lot on my machine (WSL2), but the ratio held at **2.3–2.5x** across runs. The structural argument is the durable part: a miss goes from bucket-lock-plus-reservation to a single lock-free hash lookup.


perf-self-profile/src/memory_profiling/hook.rs
```diff
@@ -300,10 +300,22 @@ pub(crate) fn on_dealloc(inner: &MemoryProfilerInner, ptr: *mut u8, _size: usize
...
+        if liveset.peek_with(&addr, |_, _| ()).is_none() {
+            return;
+        }
+        // Hit: re-resolve through the entry() API for atomic peek + remove.
+        // The bucket lock is held for the duration of the Entry, so a
+        // concurrent on_alloc overwriting the same address cannot interleave
+        // between our read and remove — preventing stale metadata on the
+        // emitted RawFree. The entry can have gone vacant since the peek (a
+        // concurrent dealloc of the same address, or the shutdown drain), in
+        // which case there is nothing to emit.
```
rcoh: this justification is not correct. A concurrent on_alloc can't occur (because this address is still considered alive, it hasn't returned to the allocator yet) due to how the hooks work.

There is a pathological race for realloc which existed before this change due to how the hook interacts with `realloc` (this is something we should fix separately.)

jv1i:
looks sound to me after #911 lands

rcoh (approved):

PR #911, opened 2026-09-16
fix(memory-profiling): pop the liveset entry before the inner realloc

rcoh:
## What

`Dial9Allocator::realloc` forwards to the inner allocator *before* it can run any hook:

```rust
let new_ptr = unsafe { self.0.realloc(ptr, old_layout, new_size) };
if !new_ptr.is_null() && let Some(inner) = ACTIVE.get() {
    on_realloc(inner, ptr, old_layout.size(), new_ptr, new_size);  // -> on_dealloc(ptr)
}
```

If the block moved, `self.0.realloc` already released the old pointer internally. Another thread can therefore have been handed that address, sampled it, and run `on_alloc`'s stale-entry path (`insert` → `entry().update(v)`) before our hook looks at the liveset. `on_dealloc` then pops the **new** allocation's entry:

- the emitted `RawFree` carries the wrong `(size, alloc_ts_ns)`, and
- the live allocation loses its liveset entry, so its real free is never reported.

`on_dealloc` itself is fine, because `Dial9Allocator::dealloc` runs the hook *before* forwarding to `self.0.dealloc(ptr)`, so the calling thread still exclusively owns the address. `alloc`/`alloc_zeroed` are fine because `ptr` is not published until the outer call returns. **`realloc` is the only path that resolves an address it no longer owns.**

## How

Give realloc the same ordering the dealloc path already has: take the liveset entry out before forwarding, emit after.

- `on_realloc_before` pops the entry while we still own `old_ptr`, and captures `ts_ns`.
- `on_realloc_success` emits the `RawFree` from that captured state, then the alloc-of-new.
- `on_realloc_failed` puts the entry back — a failed realloc never releases the old pointer, so the block is still live and must not be recorded as freed.

The shared `entry()` sequence moves into `take_liveset_entry`, whose doc states the exclusive-ownership invariant both callers depend on.

Two follow-on fixes fall out of the reordering:

- **The shutdown-drain variant becomes sound.** Today a realloc during TLS teardown pushes its flagged `RawFree` with a timestamp sampled *after* the inner realloc, so a new allocation of that address can be stamped before it and slip past `MemoryProfileSource::handle_free`'s `alloc_ts_ns >= f.ts_ns` guard. Sampling `ts_ns` up front closes that.
- **Free-of-old now strictly precedes alloc-of-new.** Design §3 says timestamps disambiguate the free-then-alloc of an in-place realloc, but today both are sampled after the realloc and can tie. The consolidator's merge-by-timestamp drain now has a real ordering to work with.

## Cost

Hot-path op count is unchanged — one liveset lookup, moved earlier. The extra `insert` only happens on realloc failure. `ACTIVE.get()` is read once instead of once, and reused across the call.


perf-self-profile/src/memory_profiling/hook.rs
```diff
@@ -321,27 +314,210 @@ pub(crate) fn on_dealloc(inner: &MemoryProfilerInner, ptr: *mut u8, _size: usize
...
+#[inline]
+pub(crate) fn on_realloc_before(inner: &MemoryProfilerInner, old_ptr: *mut u8) -> ReallocState {
+    let ts_ns = clock_monotonic_ns();
+    let Some(liveset) = &inner.liveset else {
+        return ReallocState {
+            ts_ns,
+            taken: None,
+            shutdown: false,
+        };
+    };
```
jv1i: should we / is it worth moving the clock read to happen only if liveset tracking is on? (maybe make ts_ns an Option)

jv1i (approved):
Looks good. 

re: full soundness, there may be a separate (pre-existing) issue with a race in the shutdown drain: `handle_free` does `peek_with` and `remove` separately, so maybe an address could be reused between them? (and cause the newer entry to be removed)

perf-self-profile/src/memory_profiling/hook.rs
rcoh: good catch; will move that

PR #924, opened 2026-09-16
fix(viewer): locate the toolkit skill by content, not by name

### LINKED ISSUE #915: Symposium & Dial9: Tools break when the toolkit's directory is renamed
Part of symposium-dev/symposium#294.

Depends on: nothing. Edits the same files as #913 and #914.

## What happens

Three of the five analysis tools find their shared trace parser by looking for a neighbouring directory with an exact, hardcoded name: `dial9-toolkit`. That name is not dial9's to choose. Symposium decides the installed directory's name on the user's machine, and appends a hash to it whenever the plain name is already taken, for instance by a directory the user created themselves. When that happens the three tools die on an unrecoverable module error.

Two things make it worse than a missing file. Every lookup gives up by returning its last guess **without checking whether it exists**, so the message names the file's location in the source tree, which cannot exist under an install root: it points the reader somewhere that is real in a checkout and impossible where they are. And the name is hardcoded in more places than the three tools: also in a fourth tool that happens to live inside that directory, in the shared library itself, in the build script, and in a test.

One caution for whoever fixes this: the source-tree fallback is not dead code. It is the branch that makes the tools work when run from a checkout, which is how the reproduction in #914 runs them. Make it robust rather than removing it.

## Before you verify this

**Do not check your work through a Symposium install that copies from a source tree.** Such an install is currently broken for an unrelated reason (symposium-dev/symposium#288 drops symlinked files, so the shared library never arrives at all), and it will fail whether or not your change is correct. Use the binary's own unpack, which produces real files:

```bash
cargo build -p dial9-viewer
D=$(mktemp -d) && ./target/debug/dial9-viewer agents skills "$D"
```

That is the right surface for this issue and it does not wait on another organisation.

## How to see it

```bash
# needs: a dial9 checkout, cargo, node

cargo build -p dial9-viewer
D=$(mktemp -d)
./target/debug/dial9-viewer agents skills "$D"
node "$D/dial9-zoom-window/scripts/zoom.js" dial9-viewer/ui/public/demo-trace.bin 100
# -> runs to completion, exit 0

# now rename the TOOLKIT's directory, which is what Symposium does on a collision.
# renaming the install root instead changes nothing: every lookup is relative to the
# script's own location.
mv "$D/dial9-toolkit" "$D/anything-else"
node "$D/dial9-zoom-window/scripts/zoom.js" dial9-viewer/ui/public/demo-trace.bin 100
# -> Error: Cannot find module '<parent of $D>/ui/trace_parser.js'
#    that is the source-tree layout, one level above the install root.
#    node prints four lines before the Error, so pipe through `head -8`.
#
# use zoom rather than the red-flag scanner here: the scanner crashes on this trace
# for an unrelated reason (#914), which would muddy the result.

# every place the name is hardcoded
grep -rn "dial9-toolkit" dial9-viewer/skills/*/scripts/*.js \
     dial9-viewer/ui/trace_parser.js dial9-viewer/build.rs \
     dial9-viewer/tests/server_test.rs
```

Do **not** build the standalone case with `cp -R`: it preserves the symlinks, which then dangle, and you get the same message for an entirely different reason. `cp -RL` dereferences.

## Done when

- Every analysis tool works when the **toolkit skill's own directory** has an arbitrary name.
  Rename `dial9-toolkit` to something unrelated, not to a hash-suffixed variant, or a fix that merely matches `dial9-toolkit*` will pass while the problem survives.
- Every tool works when its neighbouring directories are absent, and when a dependency genuinely cannot be found it says so, naming the locations it searched.
- Running from a checkout still works.
- A test renames the toolkit's directory and then **runs** each of the five analysis tools against a fixture, asserting exit zero and non-empty output. The existing `unpacked_scripts_resolve_dependencies` does not qualify: it checks that five paths exist, never runs anything, and hardcodes the very name this issue removes, so it passes today with the problem fully present.

## Resolution notes

The fix in progress (branch `toolkit-dir-independent-lookup`) locates the toolkit by content, never by name: the first directory holding all of `trace_parser.js`, `trace_analysis.js` and `decode.js`, trying the script's own directory, then every `<skills root>/<entry>/scripts` in sorted order, then the checkout's `ui/`. When none qualifies the error lists every location searched and names the skill to install. Two consequences to know about. If a user keeps a complete copy of their own at the plain `dial9-toolkit/` next to an installer-suffixed `dial9-toolkit-<hash>/`, the plain one wins by sort order even if it is older. And the new test runs the five tools with `node`, so `node` must be on `PATH` for `cargo test -p dial9-viewer`; CI has it.

One finding from running Claude against a renamed install: Claude Code registers a project skill under its directory name, so the suffixed toolkit is exposed as `dial9-toolkit-1a2b3c4d`. The tools work regardless (that is this issue), but prose in other skills that refers to "the `dial9-toolkit` skill" no longer matches a registered name. Tracked in symposium-dev/symposium#292.

Fluzko:
Three of the five analysis tools (`red_flag_scan.js`, `diagnose_long_poll.js`, `zoom.js`) find the shared trace parser by looking for a neighbouring directory called exactly `dial9-toolkit`. That name is not ours to choose: whoever installs the skills picks it, and Symposium appends a hash to it when the plain name is already taken. When that happens the three tools die with a module error pointing at a path that only exists in a source checkout.

## What changes

- Every tool locates the toolkit by what it contains, not by the directory's name, so the toolkit can be installed under any name.
- When the toolkit is genuinely missing, the tool says so and lists every location it searched.
- Running from a checkout keeps working.
- The unpack test now renames the toolkit directory and runs all five tools against a trace, instead of checking that some files exist.

## How it was verified

- cargo tests, clippy, fmt
- By hand: unpacked the skills, renamed the toolkit to an unrelated name, ran the five tools; removed it and checked the error.
- With Claude Code headless against an install where the toolkit carried a hash suffix: the agent ran the scan and the zoom without reaching outside the install.

## Notes for review
- If a user keeps a complete copy of their own at the plain `dial9-toolkit/` next to an installer-suffixed one, the plain copy wins.
- Claude Code registers a project skill under its directory name, so a suffixed toolkit shows up to the agent as `dial9-toolkit-<hash>`. The tools work regardless; the naming question is tracked in symposium-dev/symposium#292.

Closes #915


rcoh:
does symposium provide some general primitive to provide code alongside the skills btw?

Fluzko:
> symposium provide some general primitive to provide code alongside

not quite sure what you mean, but:
- Symposium _imports_ the whole directory, including the `skill.md` + `scripts/`. This means the code's actually _there_ and can be seen/used
- If you mean code shared between skills, that's https://github.com/symposium-dev/symposium/issues/292 territory. Open to suggestions/opinions there

rcoh (approved):

Fluzko:
Superseded by #952, which merged this commit together with the ambiguous-lookup warning.

PR #925, opened 2026-09-16
feat(taskdump): sample captures with per-worker budgets

### LINKED ISSUE #842: Optimize task-dump sampling and expose sampling metadata
Implement the task-dump sampling and trace-contract changes described in `docs/design/task-dump-capture-sampling.md`.

noxware:
## Changes

- Sample task dumps before capture using per-worker budgets.
- Add `captures_per_second_per_worker` and `DIAL9_TASK_DUMP_PER_WORKER_HZ`, deprecating idle thresholds.
- Record capture probabilities and sampling metadata for future mixed flamegraphs.
- Refine the design doc to account for worker handoffs and configuration compatibility.

## For reviewers

Just in case, check the changes in the design first, then prioritize:

- **Public API and configuration:** `dial9-tokio-telemetry/src/telemetry/task_dump_config.rs`, `dial9/src/env_config.rs`.
- **Core sampling logic:** `dial9-tokio-telemetry/src/task_dump_sampler.rs`, integrated into task polling in `dial9-tokio-telemetry/src/task_dumped.rs`.
- **Worker state and metadata:** `dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs`.

There is also a format change due to the new probability field.

Closes #842


dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -381,8 +370,29 @@ impl Source for TokioRuntimesSource {
...
+            };
+            out.push(("task_dump.sampler".into(), "per_worker_bernoulli_v1".into()));
+            // Always use worker rates: metadata merges are additive, so a
+            // scalar could become stale if a different-rate runtime attaches.
+            for (worker, sampler) in ctx.task_dump_workers.lock().unwrap().iter() {
+                out.push((
+                    format!("task_dump.worker.{worker}.captures_per_second"),
+                    config.captures_per_second_per_worker().to_string(),
+                ));
+                if let Some(timestamp) = sampler.sampling_active_ns() {
```
rcoh, line 388: lets rename this field, it took me a long time to figure out what it meant. Lets call it:

```
task_dump.worker.{worker}.sampling_started_at_ns
```

dial9-tokio-telemetry/src/telemetry/task_dump_config.rs
```diff
@@ -47,14 +36,213 @@ impl Default for TaskDumpConfig {
...
+        self.capture_interval
     }
 
     /// Optional fixed RNG seed for deterministic sampling.
     pub fn rng_seed(&self) -> Option<u64> {
         self.rng_seed
     }
 }
+
+impl<S: task_dump_config_builder::State> TaskDumpConfigBuilder<S> {
```
rcoh, line 102: I wonder if we should make TaskDumpConfigV2 no strong preference, but its possible some code would be simpler if we had two independent code paths.

dial9-tokio-telemetry/src/task_dump_sampler.rs
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    #[cold]
+    fn advance_epoch(&mut self, now: u64, elapsed: u64) {
+        let epochs = elapsed / EPOCH_NS;
+        // If whole epochs passed without a transition, the immediately
+        // preceding epoch was empty. A quiet worker can sample every transition.
+        let previous_count = if epochs == 1 { self.eligible } else { 0 };
+        self.probability = if previous_count == 0 {
+            1.0
```
rcoh, line 109: this is definitely not going to work for production. Setting probability to 1 would be a large bottleneck for a production service. I think you are already looking into some options, but we need to have a cap.

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
```diff
@@ -44,6 +48,14 @@ pub(crate) struct RuntimeContext {
     /// Global worker IDs within this runtime.
     /// Populated lazily the first time each worker thread resolves its identity.
     pub worker_ids: Mutex<BTreeSet<u64>>,
+    #[cfg(feature = "taskdump")]
+    pub(super) task_dump_config: Option<crate::telemetry::TaskDumpConfig>,
+    /// Shared by every thread that drives a logical worker. The source reads
+    /// activation metadata without locking a worker's sampling decision.
+    #[cfg(feature = "taskdump")]
+    task_dump_workers: Mutex<BTreeMap<u64, Arc<crate::task_dump_sampler::WorkerTaskDumpSampler>>>,
```
rcoh, line 56: having this be a btree map is probably going to be a performance problem. Did you benchmark this? This should probably be a Vec instead.

Secondarily, I think having this mutex be on every Pending might end up being too hot. Would be good to have some benchmarks with this on vs. off.

One option could be a TLS that stores the current worker id and has some sort of fast path for when there _wasn't_ a worker handoff?

rcoh (commented):
mostly looking at high level stuff. didnt' get super deep in the code itself

dial9-tokio-telemetry/src/telemetry/recorder/runtime_context.rs
noxware: The `Mutex<BTreeMap<...>` is not locked/looked-up on every `Pending`. I cache in TLS the `WorkerTaskDumpSampler`.

I'm in the middle of moving things around but look for `static SAMPLER: RefCell<Option<Arc<WorkerTaskDumpSampler>>>`, which is set from the existing `register_worker_if_needed`.

`TaskDumped<F>` always obtains its sampler from TLS. It also makes an `Arc::clone` of it (only if pointers are really different) to keep it stable during the `poll`.

The only `Mutex::lock` being called in `Pending`s is the internal one, inside `WorkerTaskDumpSampler`, which should be mostly un-contended (except in the `block_in_place` case which my move the worker's core to a different thread to continue operating, while the original thread also uses that Mutex).

noxware: Note: BTreeMap is used because `register_worker_if_needed` currentely receives dial9's global/displaced worker index, which has no fixed maximum (as runtimes can be dynamically created).

But, there is not reason to actually use that global id for this, I think? I can use Tokio's raw worker index without displacement, and put data into pre-allocated `Box<[Option<...>]>` using `num_workers()` as the size.

(I'm trying this change now)

noxware: @rcoh Done at [e509bd6](https://github.com/dial9-rs/dial9/pull/925/commits/e509bd69879883e2a4b968ddf7216791cfaccc7e).

However, I spotted a weird issue with the existing `worker_metrics(...)` in [9f570fa](https://github.com/dial9-rs/dial9/pull/925/commits/9f570fac22a82e652f57ddb1f8f1f26c39c0e206), and I noticed it was affecting other feature as well, so I fixed it properly here [9b9d9f7](https://github.com/dial9-rs/dial9/pull/925/commits/9b9d9f7462e0ee57695048d07095f8880e4614b2). Regression tests validate the problem.

PR #931, opened 2026-09-17
feat(telemetry): record optional worker_id on TaskSpawnEvent

### LINKED ISSUE #929: record `worker_id` on TaskSpawnEvent
(needs to be optional, tasks can be spawned off workers)

maskjelly:
Closes #929.

A `TaskSpawnEvent` said which task and where it was spawned, but not which worker ran the spawn — so a trace could not attribute spawns to workers the way poll events already are.

**The wire change:** an appended optional `worker_id` (`Option<WorkerId>`), `None` for spawns made off a worker. The format is self-describing, so existing decoders read old and new traces unchanged (absent field → `None`), which is the additive rule AGENTS.md states. `Option<WorkerId>` needed a `TraceField` impl that the orphan rule keeps out of `dial9-tokio-telemetry`, so the concrete per-type invocations in `dial9-trace-format` collapse into one blanket `impl<T: TraceField> TraceField for Option<T>` — verified byte-identical encoding for all 13 prior optional types.

**Where the id comes from, and the two traps checked:** the task-spawn hook runs on the *spawning* thread, so `RuntimeContext::spawning_worker_id` first compares `Handle::try_current()`'s id against the context's own bound runtime id — a spawn made by a worker of a *different* attached runtime would otherwise derive a global id that belongs to neither trace and pollute the other runtime's registry. It then re-derives the worker index live rather than reading the cached TLS id, so a thread that was once a worker (say, in the blocking pool after `block_in_place`) is not stamped with a stale one. Off-worker spawns record `None`, never sentinel 255.

**Consumers:** `analysis_events` decodes the field (with a `Some` round-trip assertion in the builtin test); the viewer's JavaScript reads fields by name and ignores the addition, verified; the one viewer file that *must* change is `trace_shape.rs`, which rejects built-in schemas whose field signature is not in its table — the new signature is accepted alongside the four historical ones and its completeness test covers it.

**Test:** `task_spawn_records_worker_id_only_when_spawned_on_a_worker` drives both paths deterministically — spawns from the `block_on` thread record `None` (the multi-thread `block_on` thread has no scheduler context), and an outer task running on a worker that spawns once records exactly that worker (captured by `current_worker_id()` inside the task). Pinning the field to `None` makes it fail (`left: [] right: [0]`), so it can see the difference it pins.

**Gates:** nightly-`fmt --check` (CI's rustfmt), macOS clippy across the three touched crates with the CI-equivalent feature sets, `cargo test -p dial9-tokio-telemetry` (135/135 via nextest; e2e included), `dial9-trace-format` (163/163), `dial9-viewer` lib incl. trace_shape (127 passed under the filter), and the no-`tokio_unstable` smoke job (68/68).

**Known limit and one ask:**
- A spawn inside `block_in_place` resolves `Some(old_worker)` because Tokio keeps the scheduler context on the thread; the same ambiguity CONTEXT.md documents for tid→worker mapping, and not detectable through the public API. Noted rather than faked.
- AGENTS.md asks for the demo trace to be regenerated on format changes. That capture needs a host with `perf_event_paranoid <= 1`; this macOS host cannot produce the canonical trace and a regenerated one would lose scheduling data. Nothing in CI depends on it (the new signature is accepted and old traces still read), but a maintainer on a perf-capable host should run `./scripts/regenerate_demo_trace.sh` before release.


dial9-tokio-telemetry/tests/end_to_end.rs
```diff
@@ -173,6 +173,81 @@ fn task_spawn_events_from_main_thread_are_captured() {
     );
 }
 
+/// Worker attribution: spawns that run on a worker carry that worker's id;
+/// spawns from the `block_on` thread (which is not a worker) carry `None`.
+#[cfg(tokio_unstable)]
+#[test]
+fn task_spawn_records_worker_id_only_when_spawned_on_a_worker() {
```
rcoh, line 180: given that this has some interaction w/ multi runtime, a test covering multi runtime behavior would be good

rcoh (commented):
looks good, one request for another test

maskjelly:
Added the multi-runtime test and force-pushed.

`task_spawn_from_a_foreign_runtime_worker_records_no_worker`: a second, unattached multi-thread runtime is built, a spawn into the recorded runtime is made from one of *its* workers, and the test asserts that spawn records `None` — while a control spawn made on the recorded runtime's own worker still records that worker. Both directions in one test, no timing dependence.

**It catches the bug it is for:** deleting the runtime-id check in `spawning_worker_id` makes it fail with `left: [0, 0], right: [0]` — the foreign worker's index mapped through this runtime's registry as worker `0`, a worker that never ran the spawn. Restored, 7/7 in the file.

(The same two-sided shape also covers the block_on-thread spawn, so the file now pins: worker → id, block_on thread → `None`, foreign worker → `None`.)

Gates re-run: nightly fmt, the `__nonlinux_all_features` clippy set, the end_to_end file. One note for whoever else builds locally: I had to `cargo clean -p dial9-tokio-telemetry` after another worktree shared this `CARGO_TARGET_DIR` — stale rmeta aside, nothing about the change itself.

rcoh (approved):

PR #933, opened 2026-09-18
feat(viewer): build_app_with_backend for custom stores

Sh4d1:
`storage::StorageBackend` is public, dyn-compatible and unsealed, and `AppState` already holds an `Arc<dyn StorageBackend>`. But `build_app` picks between `S3Backend` and `LocalBackend` internally, so serving traces from another object store isn't really easy. 

This adds `build_app_with_backend(config, Arc<dyn StorageBackend>)`.


rcoh:
nice! just curious, what backend are you using? feel free to upstream it if you want

Sh4d1:
@rcoh we're using google cloud storage! 

I was under the impression that custom backend was the way to go instead of adding new ones upstream.
One good point for custom backend is that for instance, in our case, we can use our own GCS client (with our baked in metrics and auth for instance). 

That being said, if it's ok for you, I can upstream a GCS implementation of course :smile:

rcoh:
Yes I'm sure someone would make use of GCS. (Especially if the pipeline support was also added)

rcoh (approved):
Lgtm!

